PR #125, opened 2022-01-24
Improve validation

### LINKED ISSUE #116: Validation of union types and UUIDSets
Currently, union types in the models are not validated. This should be changed.

ClFeSc:
Closes #116

~~@Dassderdie This could be the only option to validate union types. It's not very pretty. Do you think it would be worth the effort doing this for all union types that need validation?~~

~~All validation of union types and other literals has been removed from this PR.~~

It now adds support for validating `UUIDSet`s, `{ [key: UUID]: T }` objects and `ReachableTransferPoints`.

Also, string literal unions will now be validated, albeit a little bit more manual than I'd hoped for.


ClFeSc:
@Dassderdie Although it's now a bit different than the original aim, this should now be ready for review!

Dassderdie (changes requested):
I haven't tested it out.
The comments are mostly renaming variables.

I believe this could be further improved by deduplicating some code. E.g., a general `DictionaryValidator` utility function accepts a function to validate the value and the keys. This could then be used to create the current validators with less code. Because the current implementation is very nicely encapsulated, it is ok like it is now.
It would probably be best to wait a bit more, look at the new validators, and then figure out the best API for a refactoring.

shared/src/utils/validators/is-reachable-transfer-points-object.ts
```diff
@@ -0,0 +1,43 @@
...
+function isReachableTransferPoints(value: unknown): boolean {
+    return (
+        typeof value === 'object' &&
+        value !== null &&
+        Object.entries(value).every(
+            ([key, thisValue]: [string, unknown]) =>
+                isUUID(key, 4) &&
+                typeof thisValue === 'object' &&
+                thisValue !== null &&
+                isNumber((thisValue as { duration: number }).duration) &&
```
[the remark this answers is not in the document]
ClFeSc: Then `thisValue.duration === undefined`, right? When looking at [`isNumber`](https://github.com/typestack/class-validator/blob/develop/src/decorator/typechecker/IsNumber.ts#L19) you see `if (typeof value !== 'number') {return false;}` at the top. So this should be fine, I think.

shared/src/utils/validators/is-string-literal-union.ts
```diff
@@ -0,0 +1,18 @@
+import type { ValidationOptions, ValidationArguments } from 'class-validator';
+import { isString } from 'class-validator';
+import { makeValidator } from './make-validator';
+
+// TODO: This should actually verify the type
```
Dassderdie, line 5: I was already getting my hopes up XD

Why do you want to change it from `isString` to this?

ClFeSc: This allows me to find them more easily when I actually implement proper validation for them. And until then they at least convey the intent to the reader (unfortunately not to the runtime, though).

ClFeSc:
@ClFeSc Note to myself: Look into `@Transform` from `class-transformer` for propagating the errors.

shared/src/utils/validators/get-map-validator.ts
```diff
@@ -0,0 +1,49 @@
+export interface ValidationFunctions<T> {
+    keyValidator: (key: string) => boolean;
+    // TODO: I'm not sure I'm happy with this system already.
+    /**
+     * This transformer may return a non-`T` object.
+     */
+    valueTransformer: (value: unknown) => T;
```
Dassderdie, line 7: Remove the `valueTransformer` from the `ValidationFunctions`.
The `valueValidator` and the `consistencyValidator` both get the untransformed value.

A separate helper function can do the transformation and validation in the `valueValidator` if required.

```ts
function isValidObject(type: Constructor, value): boolean {
    return validateSync(plainToInstance(type, value)).length === 0
} 
```

ClFeSc: I've simplified the whole thing a bit.

ClFeSc:
> @ClFeSc Note to myself: Look into `@Transform` from `class-transformer` for propagating the errors.

```ts
@Transform(
        (params) => {
            const plainChildren = params.value as {
                [key: UUID]: Viewport;
            };
            return Object.values(plainChildren).map((plainViewport) =>
                plainToInstance(Viewport, plainViewport)
            );
        },
        { toClassOnly: true }
)
@ValidateNested({ each: true })
```

is a good alternative for validating the values of the id map while keeping the error locations. It needs additional work to (a) use just one decorator for it and (b) validate the keys and the consistency.

ClFeSc:
@Dassderdie What do you think of https://github.com/hpi-sam/digital-fuesim-manv/pull/125/commits/62618b86e8a99db03de9b4731e555c8385f56c57?

shared/src/utils/validators/apply-decorators.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Function that returns a new decorator that applies all decorators provided by param
+ *
+ * Useful to build new decorators (or a decorator factory) encapsulating multiple decorators related with the same feature
+ *
+ * @param decorators one or more decorators (e.g., `ApplyGuard(...)`)
+ *
+ * @publicApi
+ */
+export function applyDecorators(
```
Dassderdie, line 11:
```suggestion
/**
 * @param decorators one or more decorators (e.g., `ApplyGuard(...)`)
 * @returns a new decorator that applies all the provided decorators
 * 
 * Useful to build new decorators (or a decorator factory) encapsulating multiple decorators
 */
export function combineDecorators(
```

or `mergeDecorators()`

and renaming the file appropriately.

shared/src/utils/validators/get-map-validator.ts
```diff
@@ -0,0 +1,43 @@
...
+ * 3. For every key-value pair of the object, the following are verified:
+ *
+ *      a. It's checked whether the key fulfills {@link keyValidator}
+ *
+ *      b. The value gets converted to a (suspected, see {@link valueTransformer} and {@link valueValidator}) object of the target type using {@link valueTransformer}
+ *
+ *      c. It's checked whether the transformed value is valid using {@link valueValidator}
+ *
+ *      d. When {@link consistencyValidator} is provided, it's applied to the key and value
+ *
```
Dassderdie, line 25:
```suggestion
/**
 * The provided flow is as follows:
 *
 * 1. It's checked whether `valueToBeValidated` is of type `object`
 * 2. It's checked whether `valueToBeValidated` is not `null`
 * 3. For every key-value pair of the object, the following are verified:
 *    1. It's checked whether the key fulfills {@link keyValidator}
 *    2. The value gets converted to a (suspected, see {@link valueTransformer} and {@link valueValidator}) object of the target type using {@link valueTransformer}
 *    3. It's checked whether the transformed value is valid using {@link valueValidator}
 *    4. When {@link consistencyValidator} is provided, it's applied to the key and value
 *
```

See https://github.com/markdown-it/markdown-it/issues/215

shared/src/utils/validators/get-map-validator.ts
```diff
@@ -0,0 +1,43 @@
+export interface ValidationFunctions {
```
Dassderdie, line 1:
```suggestion
interface MapValidationFunctions {
```

Dassderdie: Removing the export just because of YAGNI - would also be fine with leaving it.

shared/src/utils/validators/index.ts
```diff
@@ -0,0 +1,6 @@
+// TODO: Propagate the errors from subobjects
```
Dassderdie, line 1: This is still a problem?

shared/src/utils/validators/get-map-validator.ts
```diff
@@ -0,0 +1,43 @@
...
+    keyValidator: (key: string) => boolean;
+    /**
+     * This validator must return `false` when {@link value} is not a valid object.
+     */
+    valueValidator: (value: unknown) => boolean;
```
Dassderdie, line 6:
```suggestion
    // TODO:  `(value: unknown) => value is T` would be better 
    // but makes the rest of the code harder 
    // because type guards always seem to be converted by typescript to boolean when nesting functions.
    valueValidator: (value: unknown) => boolean;
```

shared/src/utils/validators/get-map-validator.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ *      b. The value gets converted to a (suspected, see {@link valueTransformer} and {@link valueValidator}) object of the target type using {@link valueTransformer}
+ *
+ *      c. It's checked whether the transformed value is valid using {@link valueValidator}
+ *
+ *      d. When {@link consistencyValidator} is provided, it's applied to the key and value
+ *
+ * This process is guaranteed to be short-circuiting, so later steps can rely on a successful earlier step.
+ */
+export function getMapValidator<T>({
```
Dassderdie, line 28:
```suggestion
export function createMapValidator<T>({
```

Dassderdie: or 
```suggestion
export function makeMapValidator<T>({
```

in the style of `makeValidator()`

shared/src/utils/validators/is-uuid-set.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export function IsUUIDSet(validationOptions?: ValidationOptions) {
+    return makeValidator(
+        'isUUIDSet',
+        (value: unknown, args?: ValidationArguments) => isUUIDSet(value),
+        validationOptions
+    );
+}
```
Dassderdie, line 20: Wouldn't it be more performant to do it like this:

```suggestion
const isUUIDSet = getMapValidator<true>({
    keyValidator: (key) => isUUID(key, 4),
    valueValidator: (value) => value === true,
});

// eslint-disable-next-line @typescript-eslint/naming-convention
export function IsUUIDSet(validationOptions?: ValidationOptions) {
    return makeValidator(
        'isUUIDSet',
        (value: unknown, args?: ValidationArguments) => isUUIDSet(value),
        validationOptions
    );
}
```

Dassderdie: the same for the other validators.

shared/src/utils/validators/is-valid-object.ts
```diff
@@ -0,0 +1,10 @@
+import { plainToInstance } from 'class-transformer';
+import { validateSync } from 'class-validator';
+import type { Constructor } from '../constructor';
+
+export function isValidObject<T extends object>(
+    type: Constructor<T>,
+    value: unknown
+): boolean {
```
Dassderdie, line 8:
```suggestion
): value is T {
```

I don't think this will be of much use, but at least it clarifies the intention of this function even more.

shared/src/utils/validators/get-map-validator.ts
Dassderdie: Though this doesn't return a validator in the sense of a validator-decorator. But I can't think of a better name for it :/

Dassderdie (changes requested):
I haven't tested it yet.

shared/src/utils/validators/index.ts
ClFeSc: Yes, for example the `isReachableTransferPoints` validator would only return `false` without specifying what went wrong.

shared/src/utils/validators/get-map-validator.ts
ClFeSc: I chose to actually use the `is` for these functions and use type assertions in the usages. I'd thought that the disadvantages of an additional type cast do not outweigh the advantage of better readability of the functions.

ClFeSc:
@Dassderdie You can already review the conceptual changes, I'll adopt it to the current `dev` when we're both happy with the concept.

shared/src/utils/validators/is-id-map.ts
```diff
@@ -0,0 +1,61 @@
...
+        // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+        keyValidator: <(key: unknown) => key is UUID>((key) => isUUID(key, 4)),
+        // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+        valueValidator: <(value: unknown) => value is T>(
+            ((value) => isValidObject(type, value))
+        ),
+        consistencyValidator: (key, value) =>
+            getId(plainToInstance(type, value)) === key,
+    })(valueToBeValidated);
+}
```
Dassderdie, line 25: Can this be removed? I believe it is nowhere used.

shared/src/utils/validators/is-string-literal-union.ts
```diff
@@ -0,0 +1,26 @@
...
+import { StrictObject } from '../strict-object';
+import { makeValidator } from './make-validator';
+
+export function isStringLiteralUnion<T extends string>(
+    allowedValues: { [key in T]: true },
+    valueToBeValidated: unknown
+): boolean {
+    return (
+        typeof valueToBeValidated === 'string' &&
+        StrictObject.keys(allowedValues).includes(valueToBeValidated as T)
```
Dassderdie, line 11:
```suggestion
        allowedValues[valueToBeValidated as T]
```

We already have a Set ^^

shared/src/utils/validators/is-string-literal-union.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export function IsStringLiteralUnion<T extends string>(
+    allowedValues: { [key in T]: true },
+    validationOptions?: ValidationOptions
+) {
+    return makeValidator(
+        'isStringLiteralUnion',
+        (value: unknown, args?: ValidationArguments) =>
+            isStringLiteralUnion<T>(allowedValues, value),
```
Dassderdie, line 23: Do we need the second function?

```suggestion
        (value: unknown, args?: ValidationArguments) =>
            allowedValues[value as T],
```

shared/src/utils/validators/is-string-literal-union.ts
```diff
@@ -0,0 +1,26 @@
+import type { ValidationOptions, ValidationArguments } from 'class-validator';
+import { StrictObject } from '../strict-object';
+import { makeValidator } from './make-validator';
+
+export function isStringLiteralUnion<T extends string>(
+    allowedValues: { [key in T]: true },
+    valueToBeValidated: unknown
+): boolean {
+    return (
+        typeof valueToBeValidated === 'string' &&
```
Dassderdie, line 10: I think this check is unnecessary.

shared/src/store/action-reducers/patient.ts
```diff
@@ -56,7 +57,14 @@ export class SetVisibleStatusAction implements Action {
...
 
-    @IsString()
+    @IsStringLiteralUnion<PatientStatus>({
+        black: true,
+        blue: true,
+        green: true,
+        red: true,
+        white: true,
+        yellow: true,
+    })
```
Dassderdie, line 67: Maybe define the allowed values object where you define the type (as a convention)? This would keep the code short, make refactoring easier, and remove duplication.

so in 

````ts
export type PatientStatus = keyof typeof patientStatusAllowedLiterals;

// TODO: In the best case, the name should be self-explanatory, and we wouldn't need a comment - Can you come up with a better name?

// Mostly used for validation
export const allowedPatientStatusLiterals = {
    black: true,
    blue: true,
    green: true,
    red: true,
    white: true,
    yellow: true,
};
````

**Review by Dassderdie — CHANGES_REQUESTED** [review]
I like the general concept for the unions. It's probably the best one could do with the current TS features...

File: shared/src/utils/validators/is-string-literal-union.ts (hunk 15 above)
Comment (Dassderdie) [reply shared/src/utils/validators/is-string-literal-union.ts:10]: Wait - if the value is an object this would probably throw? So disregard this.

File: shared/src/utils/validators/is-string-literal-union.ts (hunk 14 above)
Comment (ClFeSc) [reply shared/src/utils/validators/is-string-literal-union.ts:23]: This is a convention for `class-validator` validators: They all have an `@Is...` decorator, to be used for, well, decorating properties, and a `is...` function, used when a manual validation is required (we use this e.g. with `isUUID` in this PR.

File: shared/src/utils/validators/is-string-literal-union.ts (hunk 15 above)
Comment (ClFeSc) [reply shared/src/utils/validators/is-string-literal-union.ts:10]: No, this works:

```ts
const x = { a: 1 } as unknown as Record<any, any>

console.log(x['a'])
console.log(x[1])
console.log(x[{ w: 'i' } as unknown as string])
```

prints

```console
[LOG]: 1 
[LOG]: undefined 
[LOG]: undefined 
```

Tested with <https://www.typescriptlang.org/play> on TS 4.8.4.

shared/src/store/action-reducers/patient.ts
ClFeSc: Yes, this would be the plan. I'll refactor this.

shared/src/utils/validators/is-id-map.ts
ClFeSc: Ah, good point. However, as I said [here](https://github.com/hpi-sam/digital-fuesim-manv/pull/125#discussion_r988392800), it's a convention to have such a method. I don't think I could rewrite the decorator stuff to be used without a decorator. So I'll have another think about this, but I tend to just remove this, maybe add a comment to look here if you need such an implementation in the future, as we're currently not using it.

ClFeSc:
> I like the general concept for the unions. It's probably the best one could do with the current TS features...

Yes, we would need enums to do it better, I think, as they provide a combined type/value representation. However, this would be a major refactoring and I think you knew that they could create problems in the frontend, right?

shared/src/utils/validators/is-string-literal-union.ts
ClFeSc: Good point. Will change that.

shared/src/utils/validators/is-string-literal-union.ts
ClFeSc: I'll still keep the check, though. Just to be safe, as it's not that calming to just trust a weird type wouldn't crash there.

Dassderdie:
> Yes, we would need enums to do it better, I think, as they provide a combined type/value representation. However, this would be a major refactoring and I think you knew that they could create problems in the frontend, right?

Yeah. In addition, it would make debugging harder (having `personnelType: 1` instead of `NA`). And when we use enums with string literals, then the code would be as long as it is now.

shared/src/utils/validators/is-string-literal-union.ts
```diff
@@ -0,0 +1,27 @@
+import type { ValidationOptions, ValidationArguments } from 'class-validator';
+import { makeValidator } from './make-validator';
+
+export type AllowedValues<T extends string> = { [key in T]: true };
```
Dassderdie, line 4: This is an excellent idea. One problem I saw, was that the additional `...allowedValues` variable on its own didn't make any sense. But because they now have an explicit type, we can add a comment to this type describing what their purpose is.

So something like this, maybe?

```suggestion
/**
 * A type for validating unions of string literals.
 *
 * @example
 * ````ts
 * type StringLiteralUnion = 'a' | 'b';
 *
 * const stringLiteralUnionAllowedValues: AllowedValues<StringLiteralUnion> = {
 *     a: true,
 *     b: true,
 * };
 *
 * class MyClassToValidate {
 *     @IsStringLiteralUnion<StringLiteralUnion>(stringLiteralUnionAllowedValues)
 *     public readonly myString: StringLiteralUnion;
 * }
 *
 * ````
 */
export type AllowedValues<T extends string> = { [key in T]: true };
```

ClFeSc:
@Dassderdie Do you have a clue why the backend fails with the latest WIP commit? Removing the call to `validateExerciseAction` from `isExerciseAction` fixes the crash.

Dassderdie:
> @Dassderdie Do you have a clue why the backend fails with the latest WIP commit? Removing the call to `validateExerciseAction` from `isExerciseAction` fixes the crash.

> Removing the call to `validateExerciseAction` from `isExerciseAction` fixes the crash.

This didn't work for me. The WIP commit produced a runtime error like `Cannot read ImageProperties before initialization` (I believe).

My2Cents:

* This is not a trivial problem
* I believe the concrete Model in the error message is only by chance, the first that popped up. So I would debug this by trying to go step by step through how the actions, validators, and functions get initialized. I would guess you would find some sort of circular runtime dependency instead of a nice linear tree-like structure (if you get what I mean).

Dassderdie:
> * I believe the concrete Model in the error message is only by chance, the first that popped up. So I would debug this by trying to go step by step through how the actions, validators, and functions get initialized. I would guess you would find some sort of circular runtime dependency instead of a nice linear tree-like structure (if you get what I mean).

Update: I got a similar error message in the frontend during a refactoring. It was a circular import.
You should be able to debug this by only looking at the `import` statements in the files you changed.

ClFeSc:
I've found the problem: The validation of actions requires all actions to be loaded, which in turn uses validators, which in turn requires the validation of actions, and we have our cycle. I've solved it by moving the decorator for action validation to the other action validation function. I haven't used it yet, though.

Dassderdie:
Note: [the new class validator version has a fix for `@NestedValidation`](https://github.com/typestack/class-validator/blob/develop/CHANGELOG.md#breaking-changes) though I doubt that this is relevant for us.

ClFeSc:
@Dassderdie I think this is now ready for a new review. Errors from `Actions` are not properly propagated, but this could be a bit challenging, I think. So I think we should first focus on getting this merged and then improving such small problems. Otherwise, this will take forever again.

Dassderdie (changes requested):
I haven't looked at the specifics but already have something to say ^^:

1.  You force-pushed this branch. Therefore, I have no idea what I had already looked at and what not. Which commits did you add since my last review?

2. I have a proposal for improved type-safety:
Consider the following error:

````ts
    @IsIdMap(Viewport)
    public readonly viewports: { readonly [key: UUID]: number } = {};
````
should actually be:
````ts
    @IsIdMap(Viewport)
    public readonly viewports: { readonly [key: UUID]: Viewport } = {};
````

You can achieve a type check for this by specifying the return type of the decorator factory like this:

```ts
/**
 * T is the type of the property that should be decorated
 */
type PropertyDecorator<T> = <
    Target extends { readonly [key in Key]: T },
    Key extends string
>(
    target: Target,
    propertyKey: Key
) => void;
````

```ts
export function IsIdMap<T extends object>(
    type: Constructor<T>,
    getId: (value: T) => UUID = (value) => (value as { id: UUID }).id,
    validationOptions?: ValidationOptions
// Only the return type has been added:
): PropertyDecorator<{ [key: UUID]: T }> { 
```

See https://stackoverflow.com/a/47425850/12698757

ClFeSc:
> I haven't looked at the specifics but already have something to say ^^:
> 
>     1. You force-pushed this branch. Therefore, I have no idea what I had already looked at and what not. Which commits did you add since my last review?

Only the last two commits should contain any new content.

>     2. I have a proposal for improved type-safety:
>        Consider the following error:
> 
> 
> ```ts
>     @IsIdMap(Viewport)
>     public readonly viewports: { readonly [key: UUID]: number } = {};
> ```
> 
> should actually be:
> 
> ```ts
>     @IsIdMap(Viewport)
>     public readonly viewports: { readonly [key: UUID]: Viewport } = {};
> ```
> 
> You can achieve a type check for this by specifying the return type of the decorator factory like this:
> 
> ```ts
> /**
>  * T is the type of the property that should be decorated
>  */
> type PropertyDecorator<T> = <
>     Target extends { readonly [key in Key]: T },
>     Key extends string
> >(
>     target: Target,
>     propertyKey: Key
> ) => void;
> ```
> 
> ```ts
> export function IsIdMap<T extends object>(
>     type: Constructor<T>,
>     getId: (value: T) => UUID = (value) => (value as { id: UUID }).id,
>     validationOptions?: ValidationOptions
> // Only the return type has been added:
> ): PropertyDecorator<{ [key: UUID]: T }> { 
> ```
> 
> See https://stackoverflow.com/a/47425850/12698757

I'll look into it.

Dassderdie:
> I'll look into it. Edit: Added it.

The `PropertyDecorator` should probably live in a separate file and you can add this to all of our decorators (directly or indirectly via `makeDecorator`).
This should in theory work for most decorators in `class-validator` (`@isString`, ...). But their typings are for some reason not so strong.

ClFeSc:
> > I'll look into it. Edit: Added it.
> 
> The `PropertyDecorator` should probably live in a separate file and you can add this to all of our decorators (directly or indirectly via `makeDecorator`). This should in theory also work for most decorators in `class-validator` (`@isString`, ...). But their typings are for some reason not so strong.

I had the same idea. But we have to find a better name then, as importing `PropertyDecorator` will shadow the global `PropertyDecorator` constant which makes ESLint unhappy n(and rightfully so).

> Edit: This solution only works for `public` properties...

Interesting. But it's better than nothing and most of the decorated properties are `public`.

Dassderdie:
> I had the same idea. But we have to find a better name then, as importing `PropertyDecorator` will shadow the global `PropertyDecorator` constant which makes ESLint unhappy n(and rightfully so).

"TypedPropertyDecorator" or  "GenericPropertyDecorator"? ^^


> > Edit: This solution only works for `public` properties...
> 
> Interesting. But it's better than nothing and most of the decorated properties are `public`.

A good compromise would be to have 
`Target extends { [key in Key]?: T }`.
In this case `public` properties should get the full typesafety while `private` and `protected` ones would still be possible.

Maybe this would be worth an issue in `class-validator`?

ClFeSc:
> > I had the same idea. But we have to find a better name then, as importing `PropertyDecorator` will shadow the global `PropertyDecorator` constant which makes ESLint unhappy n(and rightfully so).
> 
> "TypedPropertyDecorator" or "GenericPropertyDecorator"? ^^

Hm, ok.

> > > Edit: This solution only works for `public` properties...
> > 
> > 
> > Interesting. But it's better than nothing and most of the decorated properties are `public`.
> 
> A good compromise would be to have `Target extends { [key in Key]?: T }`. In this case `public` properties should get the full typesafety while `private` and `protected` ones would still be possible.
> 
> Maybe this would be worth an issue in `class-validator`?

Oh, I thought just the type errors wouldn't work for private properties. Unfortunately, your suggestion doesn't work for me with the error

```ts
Argument of type 'ExerciseState' is not assignable to parameter of type '{ alarmGroups?: { [key: string]: AlarmGroup; } | undefined; }'.
  Property 'alarmGroups' is private in type 'ExerciseState' but not in type '{ alarmGroups?: { [key: string]: AlarmGroup; } | undefined; }'.ts(2345)
```

when making `alarmGroups` in `ExerciseState` `private`.

ClFeSc:
I've noticed a problem: The type-safe approach needs to be able to work both with simple properties (like `string`) and also with arrays (like `string[]`) as it's possible to use `{ each: true }` as an option to validate arrays (see <https://github.com/typestack/class-validator#validating-arrays>), and we're using this for `StateHistoryCompound.actionHistory`.

Dassderdie:
> I've noticed a problem: The type-safe approach needs to be able to work both with simple properties (like `string`) and also with arrays (like `string[]`) as it's possible to use `{ each: true }` as an option to validate arrays (see <https://github.com/typestack/class-validator#validating-arrays>), and we're using this for `StateHistoryCompound.actionHistory` (and also in `VehicleTemplate` which now fails the build).

Would it be a good solution to change the type based on the `ValidationOptions` in `makeValidator`? If this is only a special case for `{ each: true }` and with `T[]` (or `Record<any, T>` dictionary) this should be feasible.

ClFeSc:
> > I've noticed a problem: The type-safe approach needs to be able to work both with simple properties (like `string`) and also with arrays (like `string[]`) as it's possible to use `{ each: true }` as an option to validate arrays (see https://github.com/typestack/class-validator#validating-arrays), and we're using this for `StateHistoryCompound.actionHistory` (and also in `VehicleTemplate` which now fails the build).
> 
> Would it be a good solution to change the type based on the `ValidationOptions` in `makeValidator`? If this is only a special case for `{ each: true }` and with `T[]` (or `Record<any, T>` dictionary) this should be feasible.

This should now be included.

shared/src/store/action-reducers/alarm-group.ts
```diff
@@ -10,12 +10,13 @@ import { AlarmGroup } from '../../models/alarm-group';
 import { AlarmGroupVehicle } from '../../models/utils/alarm-group-vehicle';
 import type { Mutable } from '../../utils';
 import { cloneDeepMutable, UUID, uuidValidationOptions } from '../../utils';
+import { IsStringLiteralUnion } from '../../utils/validators';
 import type { Action, ActionReducer } from '../action-reducer';
 import { ReducerError } from '../reducer-error';
 import { getElement } from './utils/get-element';
 
 export class AddAlarmGroupAction implements Action {
-    @IsString()
```
Dassderdie, line 18: There is a ts error if the two strings are not identical. 

In the future, the boilerplate could maybe be reduced via a `@IsConstant` decorator that creates n instance and looks at what the properties value is by default.

Dassderdie (changes requested):
Only smaller stuff.

Please also look at the commit I just pushed.

shared/src/utils/validators/make-validator.ts
```diff
@@ -0,0 +1,33 @@
...
+export function makeValidator<T, Each extends boolean = false>(
+    name: `is${string}`,
+    validationFunction: (
+        value: unknown,
+        validationArguments?: ValidationArguments
+    ) => boolean,
+    validationOptions?: ValidationOptions & { each?: Each }
+    // Disabled as this is the suggested way for [class-validator](https://github.com/typestack/class-validator#custom-validation-decorators)
+    // eslint-disable-next-line @typescript-eslint/ban-types
+): GenericPropertyDecorator<T, Each> {
```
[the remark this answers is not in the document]
ClFeSc: I've removed the eslint-comment.

shared/src/store/action-reducers/alarm-group.ts
ClFeSc: I've added a `IsValue` for now.

shared/src/utils/validators/is-value.ts
```diff
@@ -0,0 +1,25 @@
...
+import { makeValidator } from './make-validator';
+
+export function isValue<T>(
+    expectedValue: T,
+    valueToBeValidated: unknown
+): boolean {
+    return isIn(valueToBeValidated, [expectedValue]);
+}
+
+// With TS 5.0 use `const T` (https://github.com/microsoft/TypeScript/pull/51865)
```
Dassderdie, line 13:
```suggestion
// TODO [typescript@>=5.0]: Use `const T` (https://github.com/microsoft/TypeScript/pull/51865)
```

See https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md

shared/src/utils/validators/is-value.ts
```diff
@@ -0,0 +1,25 @@
...
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export function IsValue<T, Each extends boolean = false>(
```
Dassderdie, line 15: Does this work with arrays and object literals like `{ a: 0 }`?

Please add a class comment for this here. (compares shallow/deep by value/reference wether the decorated property has the specified value)

Dassderdie (changes requested):

shared/src/utils/validators/is-value.ts
Dassderdie: FYI: Maybe we should add this to the documentation. I'm considering refactoring it a bit and adding a `DEVELOPMENT.md` to the root...

shared/src/utils/validators/is-value.ts
ClFeSc: I'll add a comment stating that it's using `class-validator`'s `isIn` which in turn currently uses `===`.

shared/src/utils/validators/is-literal-union.ts
```diff
@@ -21,30 +22,29 @@ import { makeValidator } from './make-validator';
...
  * ````
  */
-export type AllowedValues<T extends string> = { [key in T]: true };
+export type AllowedValues<T extends number | string | symbol> = {
+    [key in T]: true;
+};
 
-export function isStringLiteralUnion<T extends string>(
+export function isLiteralUnion<T extends number | string | symbol>(
     allowedValues: AllowedValues<T>,
```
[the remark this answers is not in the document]
ClFeSc: Hm, I see. I'd stay with the new name but with a doc comment saying that only `number | string | symbol` may be used.

shared/src/utils/validators/is-value.ts
```diff
@@ -0,0 +1,41 @@
...
+ * A validator for a constant value. Uses {@link isIn} for validation, all limitations of this validator apply.
+ *
+ * This means that, as of `class-validator` version `0.14.0`, `===` is used for equality checks.
+ * @param expectedValue The single value that is allowed for this property.
+ * @param validationOptions {@link ValidationOptions} passed on to `class-validator`.
+ * @returns A `class-validator` validator that verifies that the value is exactly {@link expectedValue}.
+ */
+// TODO [typescript@>=5.0]: Use `const T` (https://github.com/microsoft/TypeScript/pull/51865)
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export function IsValue<T, Each extends boolean = false>(
```
Dassderdie, line 31:
```suggestion
export function IsValue<T extends string | number | boolean | null | undefined | symbol, Each extends boolean = false>(
```

Using a non-primitive value that is compared by reference would, in most cases, be a mistake.

Dassderdie (approved):
Good job.

I added one last thing. If you add the change, I'm happy with this PR.

shared/src/utils/validators/is-value.ts
ClFeSc: Any reason to not include `bigint` here?

ClFeSc:
I think I'll squash-merge here as there are a lot of commits that somewhat represent the line of thought throughout this year but add no meaningful information about the contents of this PR.

shared/src/utils/validators/is-value.ts
Dassderdie: No, I just forgot about bigint.

ClFeSc:
@Dassderdie I've added two commits with small changes I noticed while writing the commit message. When you're happy with the changes, resolve the previous conversation, and I'll enable auto-merge (squash).

PR #266, opened 2022-04-09
Increase test coverage

### LINKED ISSUE #265: Improve code coverage

ClFeSc:
Closes #265


backend/src/config.spec.ts
```diff
@@ -2,40 +2,40 @@ import type { ReporterOptions } from 'envalid';
 import { cleanEnv } from 'envalid';
 import { Config } from './config';
 
+type Report = ReporterOptions<{ port: unknown }> | undefined;
+
+function getReport(port: any): Report {
+    // This is a workaround to call a private method
+    const validator = (Config as any).createTCPPortValidator();
```
Dassderdie, line 9: In general, one shouldn't call private methods even in tests. The public methods should always provide an interface that is enough to test everything. If a private method has so much internal logic that it needs a unit test, it should be refactored to a separate function/class.

This is a general remark; I will try to follow myself in future architecture decisions. :/

Dassderdie (approved):
LGTM

backend/src/config.spec.ts
ClFeSc: Yes, generally, I agree.

PR #326, opened 2022-04-26
Test for secureOn

### LINKED ISSUE #215: Backend can be crashed with WebSocket requests
When sending WebSocket requests to the backend that have an existing message name (e.g. `getState`) but don't provide correct types for the parameters, the backend may crash. This is easiest to test with sending no function as the last argument, resulting in a crash when `callback` gets called.

ClFeSc:
Closes #215 

~~Waiting for #325; still very much WIP~~

### DIFF
--- a/backend/src/exercise/websocket-handler/secure-on.spec.ts
+++ b/backend/src/exercise/websocket-handler/secure-on.spec.ts
@@ -4,40 +4,19 @@ describe('secure on', () => {
     const environment = createTestEnvironment();
     it('does not crash in production environment when fed with invalid data', async () => {
         const originalEnv = process.env.NODE_ENV;
+        // Errors are only catched in production mode
         process.env.NODE_ENV = 'production';
-        await environment.withWebsocket(async (socket) => {
-            const promise = new Promise<void>((resolve) => {
+        try {
+            await environment.withWebsocket(async (socket) => {
                 socket.insecureEmit('getState', 'any');
                 // wait a bit so that the error can occur
-                sleep(1000).then(async () => {
-                    // Check whether the server is still available
-                    await environment
-                        .httpRequest('get', '/api/health')
-                        .expect(200);
-                    resolve();
-                });
+                await sleep(1000);
+                // Check whether the server is still available
+                await environment.httpRequest('get', '/api/health').expect(200);
             });
-            await promise;
-        });
-        process.env.NODE_ENV = originalEnv;
+        } finally {
+            process.env.NODE_ENV = originalEnv;
+        }
     });
-    // it('does crash in development environment when fed with invalid data', async () => {
-    //     const originalEnv = process.env.NODE_ENV;
-    //     process.env.NODE_ENV = 'development';
-    //     await environment.withWebsocket(async (socket) => {
-    //         const promise = new Promise<void>((resolve) => {
-    //             socket.insecureEmit('getState', 'any');
-    //             // wait a bit so that the error can occur
-    //             sleep(1000).then(async () => {
-    //                 // Check whether the server is still available
-    //                 await environment
-    //                     .httpRequest('get', '/api/health')
-    //                     .expect(200);
-    //                 resolve();
-    //             });
-    //         });
-    //         await promise;
-    //     });
-    //     process.env.NODE_ENV = originalEnv;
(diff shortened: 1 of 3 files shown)


ClFeSc:
I don't get the second test to work properly, so it's excluded for now.

ClFeSc:
@Dassderdie *friendly reminder*

backend/src/exercise/websocket-handler/secure-on.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    //                 await environment
+    //                     .httpRequest('get', '/api/health')
+    //                     .expect(200);
+    //                 resolve();
+    //             });
+    //         });
+    //         await promise;
+    //     });
+    //     process.env.NODE_ENV = originalEnv;
+    // });
```
Dassderdie, line 42:
```suggestion
    // TODO: test that it does crash in development environment when fed with invalid data
```

Always write a comment explaining why you commented code out.

backend/src/exercise/websocket-handler/secure-on.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+                    // Check whether the server is still available
+                    await environment
+                        .httpRequest('get', '/api/health')
+                        .expect(200);
+                    resolve();
+                });
+            });
+            await promise;
+        });
+        process.env.NODE_ENV = originalEnv;
```
Dassderdie, line 22: Can `withWebsocket` throw? If yes, this line should be in the `finally` block of the promise. Else this is fine.

Dassderdie (changes requested):

backend/src/exercise/websocket-handler/secure-on.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+                sleep(1000).then(async () => {
+                    // Check whether the server is still available
+                    await environment
+                        .httpRequest('get', '/api/health')
+                        .expect(200);
+                    resolve();
+                });
+            });
+            await promise;
+        });
```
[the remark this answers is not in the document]
Dassderdie: If it crashes before `// Check whether the server is still available`, the whole test crashes (as in the current implementation).

backend/src/exercise/websocket-handler/secure-on.spec.ts
ClFeSc: Not sure, but probably yes. It's working with networks and stuff like that.

backend/src/exercise/websocket-handler/secure-on.spec.ts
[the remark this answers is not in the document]
ClFeSc: I think this is expected. The whole point is to test that nothing crashes, so a crash should also fail the test, which is happening when a crash occurs.

backend/src/exercise/websocket-handler/secure-on.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+            await environment.withWebsocket(async (socket) => {
+                socket.insecureEmit('getState', 'any');
+                // wait a bit so that the error can occur
+                await sleep(1000);
+                // Check whether the server is still available
+                await environment.httpRequest('get', '/api/health').expect(200);
+            });
+        } finally {
+            process.env.NODE_ENV = originalEnv;
+        }
```
Dassderdie, line 19:
```suggestion
        await environment
            .withWebsocket(async (socket) => {
                socket.insecureEmit('getState', 'any');
                // wait a bit so that the error can occur
                await sleep(1000);
                // Check whether the server is still available
                await environment.httpRequest('get', '/api/health').expect(200);
            })
            .finally(() => {
                process.env.NODE_ENV = originalEnv;
            });
```

I would have thought this is shorter but it's actually the same number of lines.

You can decide what's better.

Either way, approved.

Dassderdie (approved):

backend/src/exercise/websocket-handler/secure-on.spec.ts
ClFeSc: I think I like your version better. Visually it better ties the finally to the Promise instead of the await imo.

PR #360, opened 2022-05-03
WIP: Exercise Walkthrough test

ClFeSc:
This aims at adding a test running all actions and checking the state along the way. As this is a lot of work, that basically just defines the reducers a second time, I'll not spend too much time on this.

Based on #266


ClFeSc:
I don't think I'll do anything for this in the foreseeable future, so closing it for now.

Issue #244, opened 2022-04-05
Add error pages


Dassderdie:
* Nginx sets to site to e.g. `/403.html`

anonym-HPI:
See reasonig in https://github.com/hpi-sam/digital-fuesim-manv/pull/362

PR #362, opened 2022-05-04
add error pages

anonym-HPI:
fixes https://github.com/hpi-sam/digital-fuesim-manv/issues/244

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -25,6 +25,9 @@ RUN chown -R www-data:www-data /usr/local/app/frontend/dist/digital-fuesim-manv
 #nginx config file
 COPY ./docker/nginx/default-conf /etc/nginx/sites-available/default
 
+#nginx error.html file
+COPY ./docker/nginx/error-page /var/www/
+
 #copy docker-entrypoint.sh file into the container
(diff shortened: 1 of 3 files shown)


Dassderdie:
@anonym-HPI What is the current status on this?

anonym-HPI:
As we found out we actually only need the 502 page (as all other errors should be caught by angular, except in a multi docker environment where nginx can't find the container), should we remove most of it and only have a 502 error page?

ClFeSc:
> As we found out we actually only need the 502 page (as all other errors should be caught by angular, except in a multi docker environment where nginx can't find the container), should we remove most of it and only have a 502 error page?

Could you please elaborate on why we need a 502 error page?

anonym-HPI:
> 

Error 502 happens in docker when nginx already started, but nodejs hasn't, in this case nginx can't proxy a request and giving a 502 error page (Bad Gateway).
Without a dedicated page, there will be just the standard nginx 502 error page. Also such a page could enable auto refresh (so the user would have not to reload again). This 502 error could also accur, when the server or crashes or is not responsive (but nginx is still running).

anonym-HPI:
Closing this, nginx already serves the frontend and the only 502 errors would happen with /api or /socket.io
The UI can give an error notification in the top right already.

PR #368, opened 2022-05-06
Restore from database

### LINKED ISSUE #264: Restore from database
When starting the backend, all relevant data should be loaded from the database.

Depends on #263.

ClFeSc:
Closes #264 

Depends on #350 

TODOs:
- [x] A test fails


shared/src/store/reducer-error.ts
```diff
@@ -1,10 +1,16 @@
 /**
  * This error gets thrown if an error in a reducer function occurs because the action isn't compatible to the state.
  */
+// TODO: This comment is now outdated.
```
Dassderdie, line 4: Just remove the `and we don't need a stack trace,` below

Dassderdie (commented):
Not tested out yet. Only small stuff

backend/src/utils/user-readable-id-generator.ts
```diff
@@ -44,4 +44,12 @@ export class UserReadableIdGenerator {
     public static freeAll() {
         this.generatedIds.clear();
     }
+
+    /**
+     * Notes all provided {@link ids} as used.
+     * @param ids The ids to lock
```
anonym-HPI, line 50:
```suggestion
     * @param ids to lock
```

anonym-HPI (changes requested):
Tested it (killing dfm container) and starting it again, seems to be working.

Dassderdie (approved):

anonym-HPI (approved):

PR #369, opened 2022-05-08
change HTTP and websocket port via ENV

anonym-HPI:
please check it, possibly by building docker locally, the ENV is HTTP_PORT

HTTPS comes with another PR


frontend/src/environments/environment.deployment.ts
```diff
@@ -1,7 +1,8 @@
+// https://pumpingco.de/blog/environment-variables-angular-docker/ also see frontend/src/index.html
 import type { Environment } from './environment-type';
 
 export const environment: Environment = {
     production: true,
-    httpPort: 80,
-    websocketPort: 80,
+    httpPort: ((window as any).env.httpPort as number) || 80,
+    websocketPort: ((window as any).env.websocketPort as number) || 80,
```
ClFeSc, line 7: It could be better to use a JSON file instead of a function and load the values from this file (thanks to @Dassderdie for the suggestion).

ClFeSc (changes requested):

anonym-HPI:
Code is active on https://dev.fuesim-manv.de, seems to work

Only case that is not covered is:
location.protocol is not a number and it is not http with port 80 or https with port 443.

frontend/src/environments/environment.deployment.ts
```diff
@@ -2,6 +2,12 @@ import type { Environment } from './environment-type';
 
 export const environment: Environment = {
     production: true,
-    httpPort: 80,
-    websocketPort: 80,
+    httpPort:
+        Number(window.location.port) || window.location.protocol === 'https:'
```
ClFeSc, line 6: Why not `Number.parseInt`?

ClFeSc: @Dassderdie What do you think here?

ClFeSc: I'm happy now.

ClFeSc (approved):

Issue #333, opened 2022-04-27
Add project descriptions


ClFeSc:
There should be more information on the front page, like a link to the Website (maybe also this repo), and a warning that it's currently in preview and features may change at any time without prior notice (or something like that).

* add information to README
* add information to landing page
* add contributers file

PR #376, opened 2022-05-12
Include Warning in Landing Page

anonym-HPI:
fixes https://github.com/hpi-sam/digital-fuesim-manv/issues/333

We could maybe do it without a code snippet, but we will probably remove it later again. What could also be possible is to maybe set a cookie with a unique ID (not per person, but per message), so that when a person already saw the message and closed it, it won't be seen again.

### DIFF
--- a/frontend/src/app/pages/landing-page/landing-page/landing-page.component.html
+++ b/frontend/src/app/pages/landing-page/landing-page/landing-page.component.html
@@ -1,7 +1,11 @@
 <div class="container">
-    <h1>Digitale FüSim MANV</h1>
-    <br />
-    <div class="p-5 mb-4 bg-light rounded-3">
+    <!--https://www.w3schools.com/howto/howto_js_alert.asp-->
+    <div class="alert">
+        <span
+            class="closebtn"
+            onclick="this.parentElement.style.display='none';"
+            >&times;</span
+        >
         <div class="display-6">Hinweise/Informationen</div>
         <br />
         <p>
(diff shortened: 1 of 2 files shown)


frontend/src/app/pages/landing-page/landing-page/landing-page.component.html
```diff
@@ -1,4 +1,31 @@
...
+            Weitere Informationen zu dieser Seite finden sie auf der
+            <a
+                class="text-white"
+                target="_blank"
+                rel="noopener noreferrer"
+                href="https://hpi.de/teaching/bachelor-projects/digitales-fuehrungssimulationstraining.html"
+                >Projektwebseite</a
+            >.
+        </p>
+    </div>
```
Dassderdie, line 28: Use Bootstrap https://getbootstrap.com/docs/5.1/components/alerts/ and angular DOM manipulation.

In this case you could also use https://ng-bootstrap.github.io/#/components/alert/examples

- You then also have to import the `NgbAlertModule` into the `landing-page-module`

frontend/src/app/pages/landing-page/landing-page/landing-page.component.scss
```diff
@@ -0,0 +1,21 @@
...
+    float: right;
+    font-size: 22px;
+    line-height: 20px;
+    cursor: pointer;
+    transition: 0.3s;
+}
+
+.closebtn:hover {
+    color: black;
+}
```
Dassderdie, line 21: Use bootstrap instead

Issue #378, opened 2022-05-12
Hospital system


Dassderdie:
* Add/edit hospitals
* Transfer to hospitals
* Statistics @Dassderdie  #370 

Edit:
previous proposal (that was in the pull request):
-   make transferpoints have a type (normal, hospital) and an event when a vehicle arrives at them
    -   advantage: hospital benefit from changes on transferpoints and features (e.g. can be seen in transfer overview)
    -   disadvantage: how to implement capacities (overall and for red, yellow, what about special treatment patients, right now we seem only having a free text with description)?
    -   events: normal having no event (right now), hospitals having event: "patient arrived and gets deleted + TODO: what to do with vehicle? Send it back to where it came with like 30 min. delay (for offloading, refilling, ...)?"
-   make transferpoints only available for vehicles (not sending personnel to hospital)
-   pause patients on the way to a hospital
-   let trainers create hospitals
-   statistics distinguish between normal transferpoints and hospitals

Dassderdie:
@anonym-HPI I think an issue is a better place to discuss this instead of pull request #381 

> -   make transferpoints have a type (normal, hospital) and an event when a vehicle arrives at them
>     -   advantage: hospital benefit from changes on transferpoints and features (e.g. can be seen in transfer overview)
>     -   disadvantage: how to implement capacities (overall and for red, yellow, what about special treatment patients, right now we seem only having a free text with description)?

I think it would be better to separate `TransferPoints` and `Hospitals` (Hospitals don't have a position and no internal name). Maybe introduce a new `TransferTarget` class so that the transfer behavior is better encapsulated (composition pattern).

I mostly start with a bit of code duplication and refactor it after I have a clear idea of which parts of the code can be encapsulated.

>     -   events: normal having no event (right now), hospitals having event: "patient arrived and gets deleted + TODO: what to do with vehicle? Send it back to where it came with like 30 min. delay (for offloading, refilling, ...)?"

This would be done in the `ExerciseTickActionReducer` when refreshing the transfers. For now, it would be enough to delete patients and vehicles when they arrive.

> -   make transferpoints only available for vehicles (not sending personnel to hospital)

Maybe even only full vehicles (else we could end up with personnel without a vehicle because it got deleted -> currently an invalid state).
It should be relatively straight forward with our current architecture to add such an validator (`TransferPointFeatureManager.onFeatureDrop` and `ChooseTransferTargetPopupComponent`)
 
> -   pause patients on the way to a hospital

[...]

PR #381, opened 2022-05-15
Feature/378 hospital system

anonym-HPI:
fixes https://github.com/hpi-sam/digital-fuesim-manv/issues/378

- [x] hospitals and hospitalPatients in shared
- [x] UI to manage hospitals
- [x] hospitals can be connected with transferPoints via their pop menu
- [x] vehicles (only vehicles) can be dropped onto transferPoints and transported to hospitals
- [x] fix UI of hospital item in transferPoint popup


anonym-HPI:
I got the backend crashed when working on statistics (after I changed something in the frontend and tried to rejoin the exercise), in the console I could read that the UUIDs of the vehicles where sent to a hospital do not exist (and something about "This is expected by a optimistic update".

shared/src/store/action-reducers/transfer-point.ts
```diff
@@ -236,7 +290,7 @@ function estimateDuration(startPosition: Position, targetPosition: Position) {
     const overheadSummand = 10 * 1000;
     // In meters per second
     // On average an RTW drives 11.5 m/s (41.3 km/h https://www.leitstelle-lausitz.de/leitstelle/haeufig-gestellte-fragen/#:~:text=Wie%20viel%20schneller%20ist%20ein,30%2C4%20km%2Fh.)
-    // Be aware that this could be significantly off for longer distances due to, e.g., the use of the Autobahn.
+    // Be aware that this could be significantly off for longer distances due to, e.g., the use of an Autobahn.
```
Dassderdie, line 293: I think there is a difference between the German and English use of this word. In English you mostly mean *the* federal controlled-access highway system in Germany. Whereas in German "Autobahn" is only the type of a street. You can check the beginnings of the German and English Wikipedia articles for reference: https://en.wikipedia.org/wiki/Autobahn https://de.wikipedia.org/wiki/Autobahn

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/transfer-point-popup/transfer-point-popup.component.html
```diff
@@ -136,7 +136,7 @@ <h5 class="popover-header">
                                     title="Verbindung löschen"
                                     (click)="
                                         disconnectTransferPoint(
-                                            reachableTransferPoint.key
+                                            reachableTransferPointConnection.id
                                         )
                                     "
                                 >
```
Dassderdie, line 133: I reverted this change in the second commit, because it was incorrect

Dassderdie (approved):
@anonym-HPI I applied my feedback directly in fc472af4f1009d66b5c23abefcc8b781c9af7a82 and 85040e3063f88220ac03df761eac54436bfd491f.
I did something wrong in the first commit and therefore reverted this change in the second one -> don't be confused when reviewing.

PR #387, opened 2022-05-20
make docker image smaller

anonym-HPI:
- extend `.dockerignore`
- new `deployment`, `prune` and `prune:deployment` command
- [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/)
  - cleaning up node_modules
  - only copy necessary for production in root, frontend, shared and backend

fixes https://github.com/hpi-sam/digital-fuesim-manv/issues/386

.dockerignore could include more, but actually it is not needed, as everything is deleted afterwards (maybe saving a bit of CI time)

reduced docker image from 1.96GB down to 287MB (via `docker image ls`)


anonym-HPI:
TODO 
`npm run setup:deployment` does not work
Error: `The command '/bin/sh -c npm run setup:deployment' returned a non-zero code: 127`

.dockerignore
```diff
@@ -26,3 +28,13 @@ shared/tsconfig.build.tsbuildinfo
...
 frontend/cypress*
+
+## VSCode
+
+.vscode
+
+## more unnecessary for production, not needed inside the docker container
+
+docker-compose.yml
+README.md
```
ClFeSc, line 39:
```suggestion
**/README.md
```

docker/Dockerfile
```diff
@@ -1,40 +1,70 @@
...
+####  prod container ####
 # Use official node image as the base image, important to be on debian based for apt-get
 FROM node:16-bullseye-slim
 
+# needed to use PRODUCTION_PATH before the first FROM
+ARG PRODUCTION_PATH
+
 # Disable interactive frontend to prevent install dialogues
 ARG DEBIAN_FRONTEND=noninteractive
+# set production patch
```
ClFeSc, line 30:
```suggestion
# set production path
```

Or what is a `production patch`?

docker/Dockerfile
```diff
@@ -1,40 +1,70 @@
...
-COPY ./ /usr/local/app/
-
-# set workdir
-WORKDIR /usr/local/app
+# install necessary packages
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install --no-install-recommends \
+    nginx \
+    ca-certificates -y && \
```
ClFeSc, line 39:
```suggestion
        nginx \
        ca-certificates -y && \
```

Does this work? It'd be nicer to read imo.

docker/Dockerfile
```diff
@@ -1,40 +1,70 @@
...
-RUN npm run setup
+# copy frontend
+COPY --chown=www-data:www-data --from=builder $PRODUCTION_PATH/frontend/dist $PRODUCTION_PATH/frontend/dist
 
-# Generate the build of the application
-RUN npm run build:deployment
+# copy shared
+COPY --chown=www-data:www-data --from=builder $PRODUCTION_PATH/shared/dist $PRODUCTION_PATH/shared/dist
+COPY --chown=www-data:www-data --from=builder $PRODUCTION_PATH/shared/node_modules $PRODUCTION_PATH/shared/node_modules
+COPY --chown=www-data:www-data --from=builder $PRODUCTION_PATH/shared/package.json $PRODUCTION_PATH/shared
```
ClFeSc, line 48: Why do you need this file?

docker/Dockerfile
```diff
@@ -1,40 +1,70 @@
...
 
-# set nginx frontend file rights
-RUN chown -R www-data:www-data /usr/local/app/frontend/dist/digital-fuesim-manv
+# copy backend
+COPY --chown=www-data:www-data --from=builder $PRODUCTION_PATH/backend/dist $PRODUCTION_PATH/backend/dist
+COPY --chown=www-data:www-data --from=builder $PRODUCTION_PATH/backend/node_modules $PRODUCTION_PATH/backend/node_modules
+COPY --chown=www-data:www-data --from=builder $PRODUCTION_PATH/backend/package.json $PRODUCTION_PATH/backend
```
ClFeSc, line 53: You don't even need this file as you're not running any `npm` scripts, I think.

package.json
```diff
@@ -10,6 +10,9 @@
         "cy:install": "cd frontend && npm run cy:install",
         "install:all": "npm i && concurrently \"cd shared && npm i\" \"cd frontend && npm i\" \"cd backend && npm i\"",
         "setup": "npm i && cd shared && npm i && npm run build && cd .. && concurrently \"cd frontend && npm i\" \"cd backend && npm i\"",
+        "prune": "npm prune && cd shared && npm prune && cd ../frontend && npm prune && cd ../backend && npm prune",
+        "prune:deployment": "npm prune --production && cd shared && npm prune --production && cd ../backend && npm prune --production",
+        "deployment": "npm run setup:ci && npm run build:deployment && npm run prune:deployment",
```
ClFeSc, line 15: Where are these scripts used?

ClFeSc (changes requested):
Just some smaller stuff. Could you please also either squash all changes together, or group the changes in some logical commits? I don't think 14 Commits for so few changes are necessary.

package.json
ClFeSc: I think you could use `prune:deployment` to make the container even smaller, right?

anonym-HPI: `prune:deployment` is used in `deployment`
having `prune:deployment` without a `prune` would be maybe not great, `prune` can be used locally, when you want to clean up all packages (is not used).

ClFeSc: Ah, I didn't read the file correctly. All good then here.

.dockerignore
ClFeSc: Or maybe even

```suggestion
**/*.md
```

You'd have to be very, very careful with this, though, as there could be valid use cases for utilizing `.md` files at run-time (although we don't have something like this right now, I think).

docker/Dockerfile
anonym-HPI: when I comment out shared/package.json comes smth. like:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '/usr/local/app/backend/node_modules/digital-fuesim-manv-shared/' imported from /usr/local/app/backend/dist/src/exercise/websocket.js
```
when I comment out backend/package.json comes smth. like:
```
 import 'dotenv/config';
digital-fuesim-manv    | ^^^^^^
digital-fuesim-manv    |
digital-fuesim-manv    | SyntaxError: Cannot use import statement outside a module
```

ClFeSc: Ah, right. Completely forgot about that.

ClFeSc (approved):

Issue #380, opened 2022-05-13
Double clicking Join button sends event twice


ClFeSc:
When entering a name, you can double-click (e.g. by double-tapping enter) the join button which sends two `joinExercise` events and adds the client twice to the state.

I'd imagine this could be fixed both in the frontend and the backend:
* The frontend should not allow multiple clicks on this button.
* The backend should not allow multiple clients from one socket. -> Fixed in #388

Dassderdie:
This is  a wontfix for the frontend (cost/benefit)

PR #388, opened 2022-05-20
Forbid a socket to connect to multiple exercises at the same time

ClFeSc:
Addresses backend part of #380

### DIFF
--- a/backend/src/exercise/websocket-handler/join-exercise-handler.ts
+++ b/backend/src/exercise/websocket-handler/join-exercise-handler.ts
@@ -10,9 +10,15 @@ export const registerJoinExerciseHandler = (
         client,
         'joinExercise',
         (exerciseId: string, clientName: string, callback): void => {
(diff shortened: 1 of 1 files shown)


backend/src/exercise/websocket-handler/join-exercise-handler.ts
```diff
@@ -10,9 +10,15 @@ export const registerJoinExerciseHandler = (
         client,
         'joinExercise',
         (exerciseId: string, clientName: string, callback): void => {
-            const clientId = clientMap
-                .get(client)
-                ?.joinExercise(exerciseId, clientName);
+            const clientWrapper = clientMap.get(client)!;
```
Dassderdie, line 13: The `client` is always already in the `clientMap` when this function is called? Maybe add a comment for this.

Dassderdie (approved):
One comment, but lgtm

PR #405, opened 2022-05-27
adding an option to disable blue patients

hpistudent72:


Dassderdie (approved):
Not for dev, but the other PR

Issue #361, opened 2022-05-03
Allow fullscreen OpenLayers


ClFeSc:
Using https://openlayers.org/en/latest/examples/full-screen-drag-rotate-and-zoom.html.

PR #409, opened 2022-05-28
Zoom to fit all viewports

Dassderdie:
* Replace ugly OpenLayers buttons
* Add fullscreen mode #361

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
@@ -31,6 +31,7 @@ import {
     selectTransferLines,
     selectViewports,
 } from 'src/app/state/exercise/exercise.selectors';
+import { getStateSnapshot } from 'src/app/state/get-state-snapshot';
 import type { TransferLinesService } from '../../core/transfer-lines.service';
 import { startingPosition } from '../../starting-position';
 import { CateringLinesFeatureManager } from '../feature-managers/catering-lines-feature-manager';
@@ -82,6 +83,8 @@ export class OlMapManager {
         VectorLayer<VectorSource>,
         FeatureManager<any>
     >();
+
+    private static readonly defaultZoom = 20;
 
     constructor(
         private readonly store: Store<AppState>,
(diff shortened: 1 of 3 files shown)


frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
```diff
@@ -506,6 +515,54 @@ export class OlMapManager {
         });
     }
 
+    /**
+     * Sets the map's view to see all viewports.
+     */
+    public tryToFitViewToViewports() {
+        if (
+            getSelectRestrictedViewport(this.apiService.ownClientId)(
+                getStateSnapshot(this.store)
```
anonym-HPI, line 524: Is this generated multiple times or only once (if multiple times, maybe save it in between)?

anonym-HPI (changes requested):
- [ ] The trashbin is not correctly reset when going into fullscreen or out (moving the map, puts it into the correct position).
- [x] And the transition animation (especially when zooming in and out are gone (old buttons had them), same is true for the new feature with getting a view of all viewports.

What is not anymore possible is that I can't rotate the map on a touchscreen (which we didn't want in the first place), before I could rotate but when I don't touch the screen anymore it bounces back to the angle before.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
Dassderdie: Caching doesn't make any sense for this (we need the current snapshot value and this operation is only a few lookups -> very cheap).

Dassderdie: On second thoughts I could make `const currentState = getStateSnapshot(this.store);`.  Because it is used again a few lines below.

I couldn't see this from your comment because you selected different lines...

Dassderdie:
> * [ ]  The trashbin is not correctly reset when going into fullscreen or out (moving the map, puts it into the correct position).

I don't see a **clean** way to fix this. The same problem occurs when e.g. the window is resized. -> I'd leave it as it currently is.

anonym-HPI (approved):
the trashbin bug existed before (e.g. resizing browser window), could be fixed in another PR

anonym-HPI:
> the trashbin bug existed before (e.g. resizing browser window), could be fixed in another PR

https://github.com/hpi-sam/digital-fuesim-manv/issues/411

ClFeSc (approved):

PR #410, opened 2022-05-28
carrying-unit as vehicle with visual change (when loading a patient or unloading)

anonym-HPI:


shared/src/store/action-reducers/vehicle.ts
```diff
@@ -294,6 +300,18 @@ export namespace VehicleActionReducers {
                         // If a personnel is in transfer, this doesn't change that
                         draftState.personnel[personnelId].position = undefined;
                     });
+                    if (
```
ClFeSc, line 303: I don't like this solution for `dev`, but for an experimental branch, it should be fine.

ClFeSc:
@anonym-HPI I'd suggest restarting this work on a new branch from the current `dev` as this branch is completely messed up.

Dassderdie:
Completely outdated.

PR #412, opened 2022-05-29
add patient categories

hpistudent72:


frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.ts
```diff
@@ -63,4 +65,8 @@ export class TrainerMapEditorComponent {
     public editMapImageTemplate(mapImageTemplateId: UUID) {
         openEditImageTemplateModal(this.ngbModalService, mapImageTemplateId);
     }
+
+    public setCurrentCategory(category: any) {
```
ClFeSc, line 69: Wouldn't `string` or something like this be a better type?

shared/src/data/default-state/patient-templates.ts
```diff
@@ -12,173 +13,1210 @@ const defaultMaleImage: ImageProperties = {
...
-const defaultFemaleImage: ImageProperties = {
-    url: '/assets/female-patient.svg',
-    height: 80,
-    aspectRatio: 1,
-};
+// const defaultFemaleImage: ImageProperties = {
+//     url: '/assets/female-patient.svg',
+//     height: 80,
+//     aspectRatio: 1,
+// };
```
ClFeSc, line 20: Is this still required?

shared/src/data/default-state/patient-templates.ts
```diff
@@ -12,173 +13,1210 @@ const defaultMaleImage: ImageProperties = {
...
+
+// returns the change needed to reach the targetHealth from startHealth in the time of phases
+function calculateHealthChange(
+    starthealth: number,
+    targetHealth: number,
+    phases: number
+) {
+    return (targetHealth - starthealth) / (phases * 12 * 60);
```
ClFeSc, line 28:
```suggestion
    return (targetHealth - starthealth) / (phases * phaseTime);
```

Or do you want something else here?

shared/src/data/default-state/patient-templates.ts
```diff
@@ -12,173 +13,1210 @@ const defaultMaleImage: ImageProperties = {
...
+        -calculateHealthChange(
+            healthPointsDefaults.yellowAverage,
+            healthPointsDefaults.redAverage,
+            1
+        ),
+        -calculateHealthChange(
+            healthPointsDefaults.yellowAverage,
+            healthPointsDefaults.redAverage,
+            1
+        )
```
ClFeSc, line 337: Why are you using negations here?

shared/src/data/default-state/patient-templates.ts
```diff
@@ -12,173 +13,1210 @@ const defaultMaleImage: ImageProperties = {
...
+        calculateHealthChange(
+            healthPointsDefaults.yellowAverage,
+            healthPointsDefaults.greenAverage,
+            1
+        ) -
+            calculateHealthChange(
+                healthPointsDefaults.yellowAverage,
+                healthPointsDefaults.redAverage,
+                1
+            ),
```
ClFeSc, line 327: What are the semantics of this subtraction?

shared/src/data/default-state/patient-templates.ts
ClFeSc: You are actually using this more below, but not above.

shared/src/data/default-state/patient-templates.ts
```diff
@@ -12,173 +13,1210 @@ const defaultMaleImage: ImageProperties = {
...
+        ),
+        0
+    ),
+    [
+        {
+            matchingHealthStateId: noChangesState.id,
+            minimumHealth: healthPointsDefaults.yellowAverage,
+        },
+        {
+            matchingHealthStateId: noChangesState.id,
```
ClFeSc, line 545: Just as a general note: This phase change has no effect as dead patients are not considered for treatment and therefore keep their health at `0`. But this approach works, too.

shared/src/data/default-state/patient-templates.ts
```diff
@@ -12,173 +13,1210 @@ const defaultMaleImage: ImageProperties = {
...
+    FunctionParameters.create(0, 0, 0, 0),
+    [
+        {
+            matchingHealthStateId: yellowStartPhase7RSDecisionState.id,
+            earliestTime: 3 * phaseTime,
+        },
+    ]
+);
+
+const redFor2PhasesState = PatientHealthState.create(
```
ClFeSc, line 619: This name doesn't match the behavior (at least the naming scheme is not consistent: above, the same scheme means that there is no change, here there is change). I think a name alongside `redUntilPhase2State` would be better.

shared/src/data/default-state/patient-templates.ts
```diff
@@ -12,173 +13,1210 @@ const defaultMaleImage: ImageProperties = {
...
+        {
+            matchingHealthStateId: redInstantTransportState.id,
+            earliestTime: 3 * phaseTime,
+        },
+    ]
+);
+
+export const defaultPatientCategories: readonly PatientCategory[] = [
+    // XAXAXA Patients
+    PatientCategory.create('XAXAXA', defaultMaleImage, [
```
ClFeSc, line 819: Just as a reminder: In the long run, rename this image and remove the other one. This name gets confusing a few lines below.

shared/src/data/default-state/patient-templates.ts
```diff
@@ -12,173 +13,1210 @@ const defaultMaleImage: ImageProperties = {
...
+        PatientTemplate.create(
+            {
+                sex: 'female',
+                externalFeatures: 'blaue Augen, rothaarig, 1,69 m',
+                age: 16,
+            },
+            { [noChangesState.id]: noChangesState },
+            defaultMaleImage,
```
ClFeSc, line 827: Isn't this a duplication with the image in the category above?

shared/src/models/patient-category.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export class PatientCategory {
+    @IsString()
+    public readonly name: PatientStatusCode;
+
+    @ValidateNested()
+    @Type(() => ImageProperties)
+    public readonly image: ImageProperties;
+
+    @IsDefined()
```
ClFeSc, line 15:
```suggestion
    @IsArray()
    @ValidateNested({ each: true })
    @Type(() => PatientTemplate)
```

shared/src/models/patient-category.ts
```diff
@@ -0,0 +1,29 @@
...
+    public readonly patientTemplates: readonly PatientTemplate[] = [];
+
+    constructor(
```
ClFeSc, line 18: Deprecation comment missing

shared/src/models/utils/patient-status-code.ts
```diff
@@ -0,0 +1,55 @@
...
+type ColorCode = 'V' | 'W' | 'X' | 'Y' | 'Z';
+type BehaviourCode = 'A' | 'B' | 'C' | 'D' | 'E';
+class PatientStatusDataField {
+    @IsString()
+    public readonly colorCode: ColorCode;
+
+    @IsString()
+    public readonly behaviourCode: BehaviourCode;
+
+    constructor(colorCode: ColorCode, behaviourCode: BehaviourCode) {
```
ClFeSc, line 14: Deprecation comment missing

shared/src/models/utils/patient-status-code.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    @ValidateNested()
+    @Type(() => PatientStatusDataField)
+    public readonly secondField!: PatientStatusDataField;
+
+    @ValidateNested()
+    @Type(() => PatientStatusDataField)
+    public readonly thirdField!: PatientStatusDataField;
+
+    constructor(code: string) {
```
ClFeSc, line 35: Deprecation comment missing

ClFeSc (changes requested):
I haven't tried it out, and I haven't checked the patient health states to match their `XAXAXA`-scheme. Otherwise looks mostly good.

shared/src/data/default-state/patient-templates.ts
hpistudent72: Yes in that case it's abot tick speed and therefore not in milliseconds

shared/src/data/default-state/patient-templates.ts
hpistudent72: Currently yes, since all images are now male defaults. But we want to use the pictures with marked places later on

shared/src/data/default-state/patient-templates.ts
ClFeSc: Ok, but this relies on the ticks coming every second, which is not guaranteed. @Dassderdie What do you think here?

shared/src/data/default-state/patient-templates.ts
hpistudent72: In these cases personnel just holds the current healthpointlevel, in the other cases they actually can increase them

shared/src/data/default-state/patient-templates.ts
hpistudent72: You negate the offset and increase healthpoints afterwards

ClFeSc:
I just noticed that there is a XAXAXD patient that generates `Not all health state's ids match their key id`. Not all patients of this group do this, though.

Dassderdie:
@hpistudent72 I changed the style of the select input:

Before: 

image

After:

image

shared/src/data/default-state/patient-templates.ts
Dassderdie: has been changed to `redUntilYellowFor2PhasesState`

shared/src/data/default-state/patient-templates.ts
Dassderdie: Add TODO comment

ClFeSc (approved):

shared/src/models/patient-category.ts
```diff
@@ -0,0 +1,34 @@
+import { Type } from 'class-transformer';
+import { IsArray, IsString, ValidateNested } from 'class-validator';
+import { PatientTemplate } from './patient-template';
+import { getCreate, ImageProperties } from './utils';
+import { PatientStatusCode } from './utils/patient-status-code';
+
+export class PatientCategory {
+    @IsString()
```
ClFeSc, line 8: For future reference: This is incorrect. `PatientStatusCode` is an object.

frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
```diff
@@ -14,6 +14,7 @@ import {
     PatientTemplate,
     MapImage,
 } from 'digital-fuesim-manv-shared';
+import type { PatientCategory } from 'digital-fuesim-manv-shared/dist/models/patient-category';
```
ClFeSc, line 17: For future reference: This import is bad style.

PR #420, opened 2022-05-29
fix 404 problems with transferPoint deletion

anonym-HPI:
elements arrive immediately at targetTransferPoint before it gets deleted + fix EOC chosen deleted transferPoint

UI problems/discuss points:
- [ ] the last known position of the transferPoint is set to the position of the trashbin (it seems), elements will arrive at the position of the trashbin
- [ ] when sending multiple elements, they will all arrive on top of each other, but that was true before (creating alarm group, all same time of arrival)


shared/src/store/action-reducers/transfer-point.ts
```diff
@@ -219,8 +220,36 @@ export namespace TransferPointActionReducers {
         {
             action: RemoveTransferPointAction,
             reducer: (draftState, { transferPointId }) => {
+                // check if transferPoint exists
                 getElement(draftState, 'transferPoints', transferPointId);
-                delete draftState.transferPoints[transferPointId];
+                // TODO: make it dynamic
```
ClFeSc, line 225: Make what dynamic?

shared/src/store/action-reducers/transfer-point.ts
```diff
@@ -219,8 +220,36 @@ export namespace TransferPointActionReducers {
...
+                // Let all vehicles and personnel arrive that are on transfer to this transferPoint before deleting it
+                for (const _vehicleId of Object.keys(draftState.vehicles)) {
```
ClFeSc, line 227: Why are you using underscores?

shared/src/store/action-reducers/transfer.ts
```diff
@@ -1,10 +1,41 @@
...
+ * Personnel/Vehicle in transfer will arrive now at new transferPoint
+ * @param elementId of an element that is in transfer
+ */
+export function letElementArrive(
+    draftState: Mutable<ExerciseState>,
+    elementType: 'personnel' | 'vehicles',
+    elementId: UUID
+) {
+    const element = getElement(draftState, elementType, elementId);
+    // check
```
ClFeSc, line 21: Check what?

shared/src/store/action-reducers/transfer.ts
```diff
@@ -41,11 +72,25 @@ export class EditTransferAction implements Action {
     /**
      * How much time in ms should be added to the transfer time.
      * If it is negative, the transfer time will be decreased.
-     * If the time set the the end of the transfer time to the past, it sets it to the current time.
+     * If the time ist set to a time in the past it will be set to the current time.
```
ClFeSc, line 75:
```suggestion
     * If the time is set to a time in the past it will be set to the current time.
```

shared/src/store/action-reducers/transfer.ts
```diff
@@ -131,6 +180,26 @@ export namespace TransferActionReducers {
         rights: 'trainer',
     };
 
+    export const deleteTransfer: ActionReducer<DeleteTransferAction> = {
```
ClFeSc, line 183: What is the purpose of this action, and how and when would I use it?

shared/src/store/action-reducers/transfer.ts
```diff
@@ -131,6 +180,26 @@ export namespace TransferActionReducers {
...
+        ) => {
+            // check if transferPoint exists
+            getElement(draftState, 'transferPoints', targetTransferPointId);
+            const element = getElement(draftState, elementType, elementId);
+            if (!element.transfer) {
+                throw getNotInTransferError(element.id);
+            }
+            // let first an element arrive at the targetted transferPoint before deleting the transfer
+            letElementArrive(draftState, elementType, elementId);
+            delete draftState.transferPoints[targetTransferPointId];
```
ClFeSc, line 197: Are you sure this is correct?

ClFeSc (changes requested):

shared/src/store/action-reducers/transfer.ts
anonym-HPI: These action is there already, this could be used later in the UI to let vehicles arrive immediately. 
Right now only the function declared also in this file is used, to let all elements arrive immediately when a transferPoint gets deleted.

ClFeSc: Ok. I'm not sure the name is that great, but I'd be fine with it for now.

ClFeSc (approved):
I think it should be fine now.

PR #421, opened 2022-05-29
fix dynamic ports deployment

anonym-HPI:


ClFeSc:
@anonym-HPI Could you elaborate on what the problem was and why you've fixed it in that way?

anonym-HPI:
> @anonym-HPI Could you elaborate on what the problem was and why you've fixed it in that way?

Before it was just checking if "if `window.location.port` is set or `https` is used, use `443`, else `80`", there was no "if `window.location.port` is set, use this one". 
The fix last time just fixed, that it can use `https` `443` or `http` `80`.
E.g. if `http` `90` would be used, it has then tried to connect to `http` `443`

anonym-HPI:
Fixed it, so that it is like before, but with brackets

ClFeSc (approved):

PR #422, opened 2022-05-29
Add basic hospital patient overview table

Dassderdie:


frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/hospital-patients-table/hospital-patients-table.component.html
```diff
@@ -0,0 +1,30 @@
...
+            <td>{{ row.vehicleType }}</td>
+            <td>{{ row.departureTime | formatDuration }}</td>
+            <td>{{ row.arrivalTime | formatDuration }}</td>
+            <td>{{ row.hospitalName }}</td>
+        </tr>
+    </tbody>
+</table>
+
+<div *ngIf="!(sortedRows$ | async)?.length" class="text-muted">
+    Es wurden noch keine Patienten an Krankenhäuser versendet.
```
anonym-HPI, line 29: @christianzoellner what do you think about this text, maybe something like `transportiert` or `abtransportiert` or `gesendet`, instead of `versendet`?

frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/hospital-patients-table/hospital-patients-table.component.ts
```diff
@@ -0,0 +1,82 @@
...
+        ]).pipe(
+            map(([hospitalPatients, hospitals]) =>
+                Object.values(hospitalPatients).map((hospitalPatient) => {
+                    const hospital = Object.values(hospitals).find(
+                        (_hospital) =>
+                            _hospital.patientIds[hospitalPatient.patientId]
+                    );
+                    const row: HospitalPatientsRow = {
+                        hospitalName:
+                            hospital?.name ?? 'unbekanntes Krankenhaus',
```
anonym-HPI, line 36: This should probably not be possible, as right now, when a hospital is deleted, every patient in it gets also deleted.
Can keep it in, but maybe with a comment that this is probably not possible or so.
Or we change `removeHospital` to not delete patients (which makes them undeletable, except this would be a function added to this table).

anonym-HPI (approved):

frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/hospital-patients-table/hospital-patients-table.component.html
christianzoellner: The correct domain terms would be "transportiert" or "abtransportiert". I prefer "transportiert" for this specific sentence.

frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/hospital-patients-table/hospital-patients-table.component.ts
christianzoellner: In most cases, creation and deletion of hospitals will be BEFORE an exercise is running and never during an exercise, so the question of what happens to patients transported to a deleted hospital is an edge case.

However, I would prefert it if the patients remain in the list with a text "deleted hospital".

ClFeSc (approved):

PR #428, opened 2022-05-30
add patient info isWalkable with visual indication

anonym-HPI:


ClFeSc (approved):
Haven't checked the assignment of the walkable setting to the templates; otherwise looks good.

PR #438, opened 2022-06-07
Include `emitterId` in `ExerciseTimeline`

### LINKED ISSUE #437: Include client id in history response
The response to the `/exercise/:exerciseId/history` route should include the id of the client that emitted the action.

Note in this context that currently any import of an exercise loses all history regarding which client proposed an action.

ClFeSc:
Closes #437


Dassderdie (approved):
I was a bit unsure of whether it makes sense to include this information permanently on dev, even if we don't display it. But I guess it doesn't really matter.

Issue #435, opened 2022-06-03
Port over fixes/features from experimental to dev


ClFeSc:
This includes:

- [ ] Prevent crashing on http requests (#430)
- [x] `PatientCategory.name` is incorrectly validated as `@IsString()` (https://github.com/hpi-sam/digital-fuesim-manv/commit/2c091a9297da94c32d37b7bddb5269de3083d38b)
- [x] The import payload size is pretty limited (https://github.com/hpi-sam/digital-fuesim-manv/commit/f10b1543c571159ee77f8c8cb4e7495a06c3c00c)
- [x] Alarm Group Vehicle names are not applied to personnel and material (https://github.com/hpi-sam/digital-fuesim-manv/commit/c759c3bfa807e650bf398bde6a3b0e9ad6ab489b)

Dassderdie:
For the only open task, there is a separate issue -> close?

Issue #430, opened 2022-05-30
Prevent crashes on http requests to backend


ClFeSc:
Currently, if an HTTP request to the backend throws an uncaught error, the backend crashes. In production mode, it should instead return a reasonable error to the client and not crash (similar to `secureOn`).

PR #439, opened 2022-06-08
Port over fixes

ClFeSc:
Addresses the last three points of #435 

The first point (#430) will be handled in #440.


Dassderdie (approved):
Not tested, but lgtm.

PR #440, opened 2022-06-08
Add secure http handler

### LINKED ISSUE #430: Prevent crashes on http requests to backend

ClFeSc:
Closes #430

### DIFF
--- a/backend/src/exercise/http-handler/secure-http.ts
+++ b/backend/src/exercise/http-handler/secure-http.ts
@@ -1,7 +1,4 @@
-import type {
-    Request as ExpressRequest,
-    Response as ExpressResponse,
-} from 'express';
+import type { Response as ExpressResponse } from 'express';
 import type { HttpResponse } from './utils';
 
 export type HttpMethod =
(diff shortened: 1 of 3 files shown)


ClFeSc:
I don't think I can do the better typing, so this should be ready for review.

backend/src/exercise/http-handler/secure-http.ts
```diff
@@ -0,0 +1,54 @@
+// TODO: How bad is that?
+// eslint-disable-next-line @typescript-eslint/no-shadow
+import type { Express, Request } from 'express';
```
Dassderdie, line 3:
```suggestion
import type { Express, Request as ExpressRequest } from 'express';
```
And rename `Request` to `ExpressRequest` in the rest of the file.

backend/src/exercise/http-server.ts
```diff
@@ -26,35 +25,29 @@ export class ExerciseHttpServer {
...
-        app.get('/api/exercise/:exerciseId/history', async (req, res) => {
-            const response = await getExerciseHistory(req.params.exerciseId);
-            res.statusCode = response.statusCode;
-            res.send(response.body);
-        });
+        secureHttp(app, 'get', '/api/health', () => getHealth());
+
+        secureHttp(app, 'post', '/api/exercise', async (req: Request) =>
+            postExercise(databaseService, req.body)
+        );
```
Dassderdie, line 30: Currently, you replace the whole express API to proxy it. What about only wrapping the handler like this:
```ts
        app.get(
            '/api/health',
            secureHttp((req, res) => {
                const response = getHealth();
                res.statusCode = response.statusCode;
                res.send(response.body);
            })
        );
```

backend/src/exercise/http-server.ts
```diff
@@ -26,35 +23,27 @@ export class ExerciseHttpServer {
...
+
+        app.delete(
+            '/api/exercise/:exerciseId',
+            secureHttp(async (req) => deleteExercise(req.params.exerciseId))
+        );
+
+        app.get(
+            '/api/exercise/:exerciseId/history',
+            secureHttp(async (req) => getExerciseHistory(req.params.exerciseId))
+        );
```
Dassderdie, line 46:
```suggestion
        app.get('/api/health', (req) => secureHttp(getHealth));

        app.post('/api/exercise', (req) =>
            secureHttp(async () => postExercise(databaseService, req.body))
        );

        app.get('/api/exercise/:exerciseId', (req) =>
            secureHttp(() => getExercise(req.params.exerciseId))
        );

        app.delete('/api/exercise/:exerciseId', (req) =>
            secureHttp(async () => deleteExercise(req.params.exerciseId))
        );

        app.get('/api/exercise/:exerciseId/history', (req) =>
            secureHttp(async () => getExerciseHistory(req.params.exerciseId))
        );
```

backend/src/exercise/http-handler/secure-http.ts
```diff
@@ -0,0 +1,49 @@
...
+export function secureHttp<Result extends object | undefined>(
+    operation: (
+        req: ExpressRequest
+    ) => HttpResponse<Result> | Promise<HttpResponse<Result>>
+): (req: ExpressRequest, res: ExpressResponse) => Promise<void> {
+    // TODO: Better type `req` to verify the arguments
+    return async (req: ExpressRequest, res: ExpressResponse) => {
+        let response: HttpResponse<Result>;
+        try {
+            response = await operation(req);
```
Dassderdie, line 25:
```suggestion
export function secureHttp(
    operation: () =>
        | HttpResponse<object | undefined>
        | Promise<HttpResponse<object | undefined>>
): (req: ExpressRequest, res: ExpressResponse) => Promise<void> {
    return async (_req: ExpressRequest, res: ExpressResponse) => {
        let response: HttpResponse<any>;
        try {
            response = await operation();
```

Dassderdie (changes requested):
I think this is the only way to infer the correct type for `req`...

Dassderdie:
@ClFeSc the pipeline is failing - code wise I'm ok with it.

ClFeSc:
@Dassderdie It should work now...

Dassderdie (approved):
Still not tested, but code lgtm.

PR #441, opened 2022-06-08
Prettier on Windows

### LINKED ISSUE #423: Prettier not available on Windows
Since `prettier` uses `cat`, it is not available on Windows. This should be fixed. Also, the `.prettierignore-ci` should be added to the `.gitignore`.

ClFeSc:
Closes #423 

@anonym-HPI @Dassderdie @hpistudent72 Please verify that this actually works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ dist/
 /coverage
 
 .env
+.prettierignore-ci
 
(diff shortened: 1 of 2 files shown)


Dassderdie (approved):
It works fine on windows now. I updated the root readme respectively. Please have a look at it before merging.

PR #442, opened 2022-06-08
Add note about `docker compose` to root README

### LINKED ISSUE #403: Update documentation - new docker compose (without dash)
Before the guides included using `docker-compose` and how to install it and all the commands use `docker-compose`

Docker now includes in the Desktop and Linux command line now supports a new `docker compose` command, the plugin can be installed via `sudo apt-get install docker-compose-plugin` (https://docs.docker.com/compose/install/)

The documentation should hint people at this change, so that people are not confused, when the commands are not working, as they have the Binary Standalone Version installed (`docker-compose`).

New installations of Docker will probably have `docker compose` installed by default

ClFeSc:
Closes #403 

Would this be enough for now?

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -31,7 +31,7 @@
         {
             "label": "Start database",
             "type": "shell",
-            "command": "docker-compose up db"
+            "command": "docker compose up db || docker-compose up db"
         },
         {
             "label": "Start frontend",
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ This project is currently developed as a [bachelor project](https://hpi.de/en/st
 ### Option 1
 
 If you are using [vscode](https://code.visualstudio.com/), you can run the [task](https://code.visualstudio.com/docs/editor/tasks) `Start all` to start everything in one go.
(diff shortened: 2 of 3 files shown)


anonym-HPI (commented):
The question is, if we should change it to default being without the dash, for people new to docker + docker compose and make this remark, when people are using an old version and tell them how they can easily install the new one or just use it with a dash.

Dassderdie:
> The question is, if we should change it to default being without the dash, for people new to docker + docker compose and make this remark, when people are using an old version and tell them how they can easily install the new one or just use it with a dash.

I agree with you. The new command should be the default and the old one the remark.

Dassderdie:
> Good point, I've changed it. What should we do in `.vscode/tasks.json`? I think we have to decide on what to use there. Should we use the new version?

I'm in favor of using the newer version there too.

Dassderdie (approved):
Replace the `docker-compose` with `docker compose` in the `Start database` task and I am happy.

ClFeSc:
> > Replace the `docker-compose` with `docker compose` in the `Start database` task and I am happy.
> 
> Agreed, maybe you can do smth. like with `pwd`

Like in https://github.com/hpi-sam/digital-fuesim-manv/pull/442/commits/9f5dd2918f911214fb4fe24f8e3eacf88f33a1b2?

anonym-HPI (approved):

PR #443, opened 2022-06-08
Fix #390 - treat actions always as immutable

### LINKED ISSUE #390: Immutability of action in reducer could lead to problems
Currently, a reducer gets a mutable `ExerciseState` (immer draft) and an immutable `Action` that should be applied.

The following would produce a runtime error (not just semantically, but an actual error because of `Object.freeze`), but no type error:

```ts
        reducer: (draftState, { mapImage }) => {
            draftState.mapImages[mapImage.id] = mapImage;
            draftState.mapImages[mapImage.id].position.x = 0;
            return draftState;
        },
```

The underlying problem is described [here](https://github.com/microsoft/TypeScript/issues/13347).

I see two possibilities:

1. Document this somewhere 
2. Make the action mutable too (either trivially by deep cloning the action or by adding the action to the draft (like `{ draftState, draftAction }` )). This would hurt performance a bit, I guess? Though, if you only need to read from a huge action (e.g., the treatment stuff in the tick), you could still use [original()](https://immerjs.github.io/immer/original).

The same problem could occur when assigning a model that was created via our `Model.create()` method because they are also immutable (which is enforced via `Object.freeze()`).

Dassderdie:
* Fixes #390 by 
    * adding documentation about the bug, 
    * freezing state, and actions (in development) and 
    * assuming that actions are treated as immutable in the reducer.
* There are still errors in development that only occur when trying to replay the state during time travel. Maybe we could write some tests for the reducers in the future (initialState + a lot of actions -> apply all of them synchronously and check whether 1. no error occurs, 2. the resulting state is the expected one).

I originally intended to do more in this PR, but I think I will come back to some of this stuff later -> this is ready for review.

### DIFF
--- a/shared/src/store/reduce-exercise-state.ts
+++ b/shared/src/store/reduce-exercise-state.ts
@@ -21,10 +21,6 @@ export function reduceExerciseState(
     state: ExerciseState,
     action: ExerciseAction
 ): ExerciseState {
-    if (isDevelopment) {
-        // Make sure that the action isn't mutated in the reducer
-        freeze(action, true);
-    }
     // use immer to convert mutating operations to immutable ones (https://immerjs.github.io/immer/produce)
     return produce(state, (draftState) => applyAction(draftState, action));
 }
--- a/shared/src/store/reduce-exercise-state.ts
+++ b/shared/src/store/reduce-exercise-state.ts
@@ -1,10 +1,14 @@
-import { produce } from 'immer';
+import { freeze, produce, setAutoFreeze } from 'immer';
 import type { ExerciseState } from '../state';
 import type { Mutable } from '../utils';
+import { isDevelopment } from '../utils/is-development';
 import type { ExerciseAction } from './action-reducers';
 import { getExerciseActionTypeDictionary } from './action-reducers';
 
 const exerciseActionTypeDictionary = getExerciseActionTypeDictionary();
+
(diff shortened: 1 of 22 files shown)


shared/src/utils/is-development.ts
```diff
@@ -0,0 +1,10 @@
+// This environment is set at the backend and (manually) on the frontend, therefore we can use it in here
+declare global {
+    const process: {
+        env: {
+            NODE_ENV: 'development' | 'production';
```
ClFeSc, line 5: This type is incorrect. We use `migration` when running migrations. Also, there could be other `NODE_ENV`s that we are currently not using (common ones are e.g. `testing` or `staging`). So I'd rather use `string` here.

shared/src/store/reduce-exercise-state.ts
```diff
@@ -32,6 +36,10 @@ export function applyAction(
     draftState: Mutable<ExerciseState>,
     action: ExerciseAction
 ) {
+    if (isDevelopment) {
+        // Make sure that the action isn't mutated in the reducer
+        freeze(action, true);
```
ClFeSc, line 41: Maybe this would lead to worse performance, but wouldn't it be way easier to do a `cloneDeepMutable` of the whole `action` here?

Dassderdie:
Closed in favor of #448

PR #444, opened 2022-06-08
Enable no unchecked indexed access

### LINKED ISSUE #358: Enable noUncheckedIndexedAccess in tsconfigs
See https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess .
For some reason this option is not in the `strict` group.

This would improve the type safety, but requires (probably?) a lot of refactoring/non-null-assertions.

ClFeSc:
Closes #358

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
@@ -121,9 +121,10 @@ export class DragElementService {
             return;
         }
         // Get the position of the mouse on the map
-        const [x, y] = this.olMap.getCoordinateFromPixel(
+        const coordinate = this.olMap.getCoordinateFromPixel(
             this.olMap.getEventPixel(event)
         );
+        const [x, y] = [coordinate[0]!, coordinate[1]!];
         const position = { x, y };
         // create the element
         switch (this.transferringTemplate.type) {
@@ -148,7 +149,7 @@ export class DragElementService {
                                     this.transferringTemplate.template
                                         .patientTemplates.length
                             )
-                        ],
+                        ]!,
                         this.transferringTemplate.template.name
                     );
                     this.apiService.proposeAction(
--- a/frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
@@ -121,10 +121,9 @@ export class DragElementService {
             return;
         }
         // Get the position of the mouse on the map
-        const coordinate = this.olMap.getCoordinateFromPixel(
+        const [x, y] = this.olMap.getCoordinateFromPixel(
             this.olMap.getEventPixel(event)
-        );
-        const [x, y] = [coordinate[0]!, coordinate[1]!];
+        ) as [number, number];
         const position = { x, y };
         // create the element
         switch (this.transferringTemplate.type) {
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/viewport-feature-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/viewport-feature-manager.ts
@@ -25,8 +25,8 @@ export function isInViewport(
     viewport: Viewport
 ): boolean {
     return Viewport.isInViewport(viewport, {
-        x: coordinate[0],
-        y: coordinate[1],
+        x: coordinate[0]!,
+        y: coordinate[1]!,
     });
 }
 
@@ -47,10 +47,16 @@ export class ViewportFeatureManager
             olMap,
             layer,
             (targetPositions, viewport) => {
+                // We need at least on targetPosition
+                if (targetPositions.length < 1) {
+                    throw new Error(
+                        'Got unexpectedly empty targetPositions array'
+                    );
+                }
                 apiService.proposeAction({
                     type: '[Viewport] Move viewport',
                     viewportId: viewport.id,
-                    targetPosition: targetPositions[0],
+                    targetPosition: targetPositions[0]!,
                 });
             },
             createLineString
@@ -75,6 +81,12 @@ export class ViewportFeatureManager
     override createFeature(element: Viewport): Feature<LineString> {
         const feature = super.createFeature(element);
         this.modifyHelper.onModifyEnd(feature, (newPositions) => {
+            // We need at least 3 coordinates as `newPositions`
+            if (newPositions.length < 3) {
+                throw new Error(
+                    `Got unexpectedly short newPositions array: ${newPositions}`
+                );
+            }
             // Skip when not all coordinates are properly set.
             if (
                 !newPositions.every(
@@ -89,8 +101,8 @@ export class ViewportFeatureManager
             }
             const lineString = newPositions;
 
-            const topLeft = lineString[0];
-            const bottomRight = lineString[2];
+            const topLeft = lineString[0]!;
+            const bottomRight = lineString[2]!;
             this.apiService.proposeAction({
                 type: '[Viewport] Resize viewport',
                 viewportId: element.id,
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/viewport-feature-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/viewport-feature-manager.ts
@@ -75,12 +75,6 @@ export class ViewportFeatureManager
     override createFeature(element: Viewport): Feature<LineString> {
         const feature = super.createFeature(element);
         this.modifyHelper.onModifyEnd(feature, (newPositions) => {
-            // We need at least 3 coordinates as `newPositions`
-            if (newPositions.length < 3) {
-                throw new Error(
-                    `Got unexpectedly short newPositions array: ${newPositions}`
-                );
-            }
             // Skip when not all coordinates are properly set.
             if (
                 !newPositions.every(
@@ -95,6 +89,7 @@ export class ViewportFeatureManager
             }
             const lineString = newPositions;
 
+            // We expect the viewport LineString to have 4 points.
             const topLeft = lineString[0]!;
             const bottomRight = lineString[2]!;
             this.apiService.proposeAction({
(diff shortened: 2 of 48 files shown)


frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/transfer-point-popup/transfer-point-popup.component.html
```diff
@@ -1,3 +1,4 @@
+<!-- TODO: This file is broken. Check also the other html files, as npm run build doesn't find those -->
```
ClFeSc, line 1: This should be the only thing missing now.

ClFeSc:
It also needs some testing to verify no wrong assumptions were made.

ClFeSc:
I did some testing and it seems to be working just fine.

frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
```diff
@@ -121,9 +121,10 @@ export class DragElementService {
             return;
         }
         // Get the position of the mouse on the map
-        const [x, y] = this.olMap.getCoordinateFromPixel(
+        const coordinate = this.olMap.getCoordinateFromPixel(
             this.olMap.getEventPixel(event)
         );
+        const [x, y] = [coordinate[0]!, coordinate[1]!];
```
Dassderdie, line 127:
```suggestion
        const [x, y] = this.olMap.getCoordinateFromPixel(
            this.olMap.getEventPixel(event)
        ) as [number, number];
```
Is maybe better?

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/viewport-feature-manager.ts
```diff
@@ -47,10 +47,16 @@ export class ViewportFeatureManager
             olMap,
             layer,
             (targetPositions, viewport) => {
+                // We need at least on targetPosition
+                if (targetPositions.length < 1) {
+                    throw new Error(
+                        'Got unexpectedly empty targetPositions array'
+                    );
+                }
```
Dassderdie, line 55: I'm unsure about these. On the one hand, this is defensive programming (bad practice -> Prof. Baudisch either PT1 or PT2). On the other hand, the contract defined by OpenLayers is not clearly documented anywhere.

I tend a bit more to just keep it as before.

frontend/src/app/state/exercise/exercise.selectors.ts
```diff
@@ -17,8 +41,10 @@ export const selectMapImagesTemplates = (state: AppState) =>
     state.exercise.mapImageTemplates;
 export const getSelectMapImageTemplate =
     (mapImageTemplateId: UUID) => (state: AppState) =>
-        state.exercise.mapImageTemplates.find(
-            (template) => template.id === mapImageTemplateId
+        errorIfNotExists(
```
Dassderdie, line 44: I think these should rather return `T | undefined` instead.

shared/src/store/action-reducers/transfer-point.ts
```diff
@@ -256,12 +256,14 @@ export namespace TransferPointActionReducers {
                     draftState.transferPoints
                 )) {
                     const transferPoint =
-                        draftState.transferPoints[_transferPointId];
+                        draftState.transferPoints[_transferPointId]!;
```
Dassderdie, line 259:
```ts
for (const [
    _transferPointId,
    transferPoint,
] of Object.entries(draftState.transferPoints)) {
```

shared/src/store/action-reducers/transfer-point.ts
```diff
@@ -256,12 +256,14 @@ export namespace TransferPointActionReducers {
...
                     for (const connectedTransferPointId of Object.keys(
                         transferPoint.reachableTransferPoints
                     )) {
                         const connectedTransferPoint =
-                            draftState.transferPoints[connectedTransferPointId];
+                            draftState.transferPoints[
```
Dassderdie, line 264: Use `Object.entries()` instead

shared/src/store/reduce-exercise-state.spec.ts
```diff
@@ -36,7 +36,10 @@ describe('exerciseReducer', () => {
     });
 
     it('should apply simple actions', () => {
-        const viewports = [generateViewport(), generateViewport()];
+        const viewports: [Viewport, Viewport] = [
+            generateViewport(),
+            generateViewport(),
+        ];
```
Dassderdie, line 42:
```suggestion
        const viewports = [
            generateViewport(),
            generateViewport(),
        ] as const;
```

Dassderdie (changes requested):
I'm not finished yet, but already left some comments

frontend/src/app/state/exercise/exercise.selectors.ts
ClFeSc: Are you sure? This would require much work on the usage side of the selectors, I think.

Dassderdie: An inflationary use of `!` should make this manageable (works in the templates too). But I think it is better to make this assumption on a lower level.

frontend/src/app/shared/pipes/keys.pipe.ts
```diff
@@ -12,7 +12,8 @@ export class KeysPipe implements PipeTransform {
     // Accepts undefined and null too, to make it easier to use in templates with e.g. | async
     transform<Value extends { [key: string]: any } | null | undefined>(
         object: Value
-    ): Value extends null | undefined ? Value : (keyof Value)[] {
+        // The keys must be strings, we don't want to allow number indexing here (see https://github.com/microsoft/TypeScript/issues/41966#issuecomment-745434551)
```
Dassderdie, line 15:
```suggestion
        // To be consistent with `Object.keys()` it is necessary to only allow string or string literals as keys (see https://github.com/microsoft/TypeScript/issues/41966#issuecomment-745434551)
```

frontend/src/app/shared/pipes/values.pipe.ts
```diff
@@ -10,9 +10,11 @@ export class ValuesPipe implements PipeTransform {
      * @returns an array of the values in the object in their original order
      */
     // Accepts undefined and null too, to make it easier to use in templates with e.g. | async
-    transform<Value extends { [key: string]: Item } | null | undefined, Item>(
-        object: Value
-    ): Value extends null | undefined ? Value : Item[] {
+    transform<
+        Value extends { [key: string]: Item } | null | undefined,
+        // When the `:` side is used, `Item` is unused
```
Dassderdie, line 15: What is bad about that? I don't get any errors or warnings from TS or eslint about that...

frontend/src/app/state/exercise/exercise.selectors.ts
Dassderdie: On second thought, I would also be fine with using `!` here in the selectors.

Currently this is using defensive programming via the wrapper function and the ugly error checks. I'm against that.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/viewport-feature-manager.ts
ClFeSc: I remove the check.

frontend/src/app/state/exercise/exercise.selectors.ts
ClFeSc: I'll replace the `errorIfNotExists` with `!` in the selectors.

shared/src/store/action-reducers/transfer-point.ts
ClFeSc: I can even use `.values` as the id is only used to get the element.

shared/src/store/action-reducers/transfer-point.ts
ClFeSc: I don't know what you mean here.

frontend/src/app/shared/pipes/values.pipe.ts
ClFeSc: This is not bad. It's just a hint at the programmer, why the `Item` is `never` in this case. I think this would be a bit unintuitive for me.

shared/src/store/action-reducers/transfer-point.ts
Dassderdie: Never mind, `reachableTransferPoints` are only the ids...

frontend/src/app/shared/pipes/values.pipe.ts
Dassderdie: So you changed

```ts
    transform<Value extends { [key: string]: Item } | null | undefined, Item>(
        object: Value
    ): Value extends null | undefined ? Value : Item[] {
```

to

```ts
    transform<
        Value extends { [key: string]: Item } | null | undefined,
        // When the `:` side is used, `Item` is unused
        Item = Value extends { [key: string]: infer T } ? T : never
    >(object: Value): Value extends null | undefined ? Value : Item[] {
```

to make the code more understandable?

frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/area-statistics.service.ts
```diff
@@ -24,17 +23,20 @@ export class AreaStatisticsService {
...
-                        }))
-                        .filter((entry) => entry.value !== undefined)
+                this.store.select(
+                    (state) =>
+                        state.exercise.statistics
+                            .map((statisticEntry) => ({
+                                value:
+                                    areaId === null
+                                        ? statisticEntry.exercise
+                                        : // THis is filtered out in the next step, if the viewport didn't exist yet
```
Dassderdie, line 33:
```suggestion
                                        : // This is filtered out in the next step if the viewport didn't exist yet
```

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/viewport-feature-manager.ts
```diff
@@ -75,6 +75,12 @@ export class ViewportFeatureManager
     override createFeature(element: Viewport): Feature<LineString> {
         const feature = super.createFeature(element);
         this.modifyHelper.onModifyEnd(feature, (newPositions) => {
+            // We need at least 3 coordinates as `newPositions`
+            if (newPositions.length < 3) {
+                throw new Error(
+                    `Got unexpectedly short newPositions array: ${newPositions}`
+                );
+            }
```
Dassderdie, line 83:
```suggestion
```

Defensive programming?

Dassderdie (changes requested):
Just some small things

frontend/src/app/shared/pipes/values.pipe.ts
Dassderdie: The previous typings were not correct always correct

Dassderdie (approved):
I haven't tried out the code. I will try it out together with #474

PR #445, opened 2022-06-08
Add keyboard controls for the ol map

Dassderdie:
You can use the arrow keys to navigate and the plus/minus keys to change the zoom level.

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/exercise-map.component.html
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/exercise-map.component.html
@@ -1,4 +1,5 @@
-<div #openLayersContainer class="h-100 w-100 position-relative">
+<!-- We want keyboard events to go to the OpenLayers map, not the container -> tabindex="-1" -->
+<div #openLayersContainer class="h-100 w-100 position-relative" tabindex="-1">
     <div
         class="position-absolute btn-group-vertical"
         style="z-index: 10"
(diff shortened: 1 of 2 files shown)


frontend/src/app/pages/exercises/exercise/shared/exercise-map/exercise-map.component.html
```diff
@@ -1,4 +1,5 @@
-<div #openLayersContainer class="h-100 w-100 position-relative">
+<!-- We want keyboard events to go to the OpenLayers map, not the container -> tabindex="-1" -->
```
ClFeSc, line 1: This line is a bit ... weird.

ClFeSc (changes requested):
Still have to test it out, just a comment for now.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/exercise-map.component.html
Dassderdie:
```suggestion
<!-- `tabindex="-1"` because we want the keyboard events to be fired on the OpenLayers map instead of this container -->
```

Better? Else you need to be more specific than "weird"

ClFeSc: Ah, oh, sorry. I only now saw that the `->` is an arrow and not a HTML comment end that went badly. So it's fine then.

ClFeSc (approved):
Seems to be working.

PR #447, opened 2022-06-11
Feature/improve optimistic updates

Dassderdie:
This refactoring should be seen as part of my bachelor thesis.

* Refactor the optimistic action handler to work more intuitive
* Refactor the tests to be easier to read and write
* Add more tests
* Improve the performance for the following cases (by not triggering all the selectors again):
    * No optimistic actions are applied and the server sends an action
    * Exactly one optimistic action has been applied and the server responds to it
* When the server disconnects, the client no longer goes in `time-travel-mode (but goes in an obviously invalid state instead ^^)

What should be done later:

* Enable https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes
* Replace most of our uses of `undefined` in the models with `null`.  `undefined` should be reserved for stuff that gets deleted. (e.g. `UuidSet`)

### DIFF
--- a/frontend/src/app/core/api.service.ts
+++ b/frontend/src/app/core/api.service.ts
@@ -41,10 +41,17 @@ export class ApiService {
             this.isTimeTraveling
                 ? (this.presentState = exercise)
                 : this.store.dispatch(setExerciseState(exercise)),
-        () =>
-            this.isTimeTraveling
-                ? this.presentState!
-                : getStateSnapshot(this.store).exercise,
+        () => {
+            // TODO: this is called synchronously, because at this time isTimeTraveling is not defined, an error occurs.
+            // Probably the best solution would be to do refactor all of the time travel and
+            try {
+                return this.isTimeTraveling
+                    ? this.presentState!
+                    : getStateSnapshot(this.store).exercise;
+            } catch {
+                return getStateSnapshot(this.store).exercise;
+            }
+        },
         (action) => {
             if (!this.isTimeTraveling) {
                 this.store.dispatch(applyServerAction(action));
--- a/frontend/src/app/core/api.service.ts
+++ b/frontend/src/app/core/api.service.ts
@@ -103,6 +103,9 @@ export class ApiService {
         return this.timeTravelHelper?.timeConstraints;
     }
 
+    // TODO: It is suboptimal to differentiate between null and undefined.
+    // But because this distinction is only used in this class and better exposed via
+    // isTimeTraveling and getCurrentRole, this is ok for now.
     /**
      * The id of the client that is currently joined to the exercise with {@link exerciseId}
      * undefined: the client is not joined to an exercise
--- a/frontend/src/app/core/api.service.ts
+++ b/frontend/src/app/core/api.service.ts
@@ -89,8 +89,9 @@ export class ApiService {
             this.messageService.postError({
                 title: 'Zeitreise ist nicht aktiviert.',
             });
+            return;
         }
-        this.timeTravelHelper?.jumpToTime(exerciseTime);
+        this.timeTravelHelper.jumpToTime(exerciseTime);
     }
 
     public timeConstraints$: Observable<TimeConstraints | undefined> =
--- a/frontend/src/app/core/api.service.ts
+++ b/frontend/src/app/core/api.service.ts
@@ -11,7 +11,7 @@ import type {
 } from 'digital-fuesim-manv-shared';
 import { reduceExerciseState } from 'digital-fuesim-manv-shared';
 import type { Observable } from 'rxjs';
-import { lastValueFrom, map, of, switchMap } from 'rxjs';
+import { BehaviorSubject, lastValueFrom, map, of, switchMap } from 'rxjs';
 import type { AppState } from '../state/app.state';
 import {
     applyServerAction,
@@ -22,12 +22,19 @@ import { getStateSnapshot } from '../state/get-state-snapshot';
 import { httpOrigin } from './api-origins';
 import { MessageService } from './messages/message.service';
 import { PresentExerciseHelper } from './present-exercise-helper';
+import type { TimeConstraints } from './time-travel-helper';
 import { TimeTravelHelper } from './time-travel-helper';
 
 @Injectable({
     providedIn: 'root',
 })
 export class ApiService {
+    public get isTimeTraveling() {
+        return this.timeTravelHelper !== undefined;
+    }
+    public readonly isTimeTraveling$ = new BehaviorSubject(
+        this.isTimeTraveling
+    );
     constructor(
         private readonly store: Store<AppState>,
         private readonly httpClient: HttpClient,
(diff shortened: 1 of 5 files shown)


frontend/src/app/core/api.service.ts
```diff
@@ -41,10 +41,17 @@ export class ApiService {
             this.isTimeTraveling
                 ? (this.presentState = exercise)
                 : this.store.dispatch(setExerciseState(exercise)),
-        () =>
-            this.isTimeTraveling
-                ? this.presentState!
-                : getStateSnapshot(this.store).exercise,
+        () => {
+            // TODO: this is called synchronously, because at this time isTimeTraveling is not defined, an error occurs.
```
Dassderdie, line 45:
```suggestion
            // TODO: this is called synchronously. An error occurs because at this time `isTimeTraveling` is not defined yet.
```

frontend/src/app/core/optimistic-action-handler.spec.ts
```diff
@@ -12,213 +18,300 @@ class WordStateManager {
...
+            }
         }
     );
 
-    it('should correctly perform actions', async () => {
-        optimisticActionHandler.performAction({
-            type: 'addLetter',
-            letter: 'a',
-        });
+    it('performs actions', async () => {
```
ClFeSc, line 154:
```suggestion
    it('performs actions', () => {
```

frontend/src/app/core/optimistic-action-handler.spec.ts
```diff
@@ -12,213 +18,300 @@ class WordStateManager {
...
+        performAction(new AddLetterAction('a'));
```
ClFeSc, line 155: Maybe add an expect for the state before proposing the first letter?

frontend/src/app/core/optimistic-action-handler.spec.ts
```diff
@@ -12,213 +18,300 @@ class WordStateManager {
...
+        const actionProposalD1 = proposeAction(new AddLetterAction('d'), false);
+        expect(wordStateManager.state.word).toEqual('bcca');
+        const actionProposalD2 = proposeAction(new AddLetterAction('d'), true);
+        expect(wordStateManager.state.word).toEqual('bccad');
+        performAction(new AddLetterAction('e'));
+        expect(wordStateManager.state.word).toEqual('bccead');
+        const performedActionProposalA = actionProposalA.performAction();
+        expect(wordStateManager.state.word).toEqual('bccead');
+        const performedActionProposalB = actionProposalD1.performAction();
+        expect(wordStateManager.state.word).toEqual('bccead');
```
ClFeSc, line 268: Shouldn't there be two `d`s? One from the optimistic update (`actionProposalD2`, line 261) and one from the performed action (`actionProposalD1` and `performedActionProposalB`, lines 259 and 267)?

frontend/src/app/core/optimistic-action-handler.spec.ts
```diff
@@ -12,213 +18,300 @@ class WordStateManager {
...
+
+        await actionProposalA.performAction().resolveProposal();
+        expect(wordStateManager.state.word).toEqual('bac');
+        await actionProposalB.performAction().resolveProposal();
+        expect(wordStateManager.state.word).toEqual('babc');
+        await actionProposalC.performAction().resolveProposal();
+        expect(wordStateManager.state.word).toEqual('babc');
+    });
+
+    it('handles multiple optimistic proposals with a performAction and an rejected proposal', async () => {
```
ClFeSc, line 299:
```suggestion
    it('handles multiple optimistic proposals with a performAction and a rejected proposal', async () => {
```

ClFeSc (changes requested):

frontend/src/app/core/optimistic-action-handler.spec.ts
Dassderdie: The first `d` is optimistic proposed and, therefore, the one in the state.  The second one only gets proposed (non-optimistic) and gets rejected in line 269. The second one never made it to the state.

Dassderdie: The only way to see whether one action belongs to a proposal is by comparing the sent action with the received action by value. Because D1 and D2 are identical, I assume the performAction from D1 is the performAction from D2. Therefore, there is only one `d` here.

Stuff to change: 

* [ ] Rename  `performedActionProposalB` to `performedActionProposalD1`
* [ ] Resolve  `performedActionProposalD1` at the end

Dassderdie: This is the same behaviour as in line 287

ClFeSc: Ah, I see. Fine then. You can resolve this conversation when you've included the todos.

frontend/src/app/core/api.service.ts
```diff
@@ -75,36 +82,25 @@ export class ApiService {
...
+        this.timeTravelHelper?.jumpToTime(exerciseTime);
     }
-    public get isTimeTraveling() {
-        return this.timeTravelHelper.isTimeTraveling;
+
+    public timeConstraints$: Observable<TimeConstraints | undefined> =
+        this.isTimeTraveling$.pipe(
+            switchMap(
+                () => this.timeTravelHelper?.timeConstraints$ ?? of(undefined)
+            )
```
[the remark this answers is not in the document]
Dassderdie: Always when `isTimeTraveling` changes, we either emit the `timeConstraints` or `undefined` - depending on whether the timeTravelHelper is defined.

This could also be written as 
```ts
(isTimeTraveling) => isTimeTraveling? this.timeTravelHelper!.timeConstraints$ : of(undefined)
```

the `of(undefined)` is necessary, because `switchMap` wants an observable.

frontend/src/app/core/api.service.ts
```diff
@@ -145,23 +141,47 @@ export class ApiService {
             ? getSelectClient(this.ownClientId)(getStateSnapshot(this.store))
             : // clientId is expected to never be the empty string
               (this.ownClientId as null | undefined)
-    ) {
-        return ownClient ? ownClient.role : 'timeTravel';
+    ): CurrentRole {
+        return ownClient === null
+            ? 'timeTravel'
+            : ownClient === undefined
```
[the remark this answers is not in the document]
ClFeSc: Resolution: TODO comment

frontend/src/app/core/api.service.ts
```diff
@@ -75,36 +82,25 @@ export class ApiService {
...
-    public timeConstraints$ = this.timeTravelHelper.timeConstraints$;
-    public get timeConstraints() {
-        return this.timeTravelHelper.timeConstraints;
+    public jumpToTime(exerciseTime: number) {
+        if (!this.timeTravelHelper) {
+            this.messageService.postError({
+                title: 'Zeitreise ist nicht aktiviert.',
+            });
+        }
+        this.timeTravelHelper?.jumpToTime(exerciseTime);
```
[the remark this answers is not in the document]
ClFeSc: You have given a thumbs up but I don't think you've changed anything here.

ClFeSc (approved):

PR #448, opened 2022-06-11
Feature/390 always treat actions as immutable

Dassderdie:
Fixes https://github.com/hpi-sam/digital-fuesim-manv/issues/390 by
* adding documentation about the bug,
* freezing state, and actions
* Adding an eslint rule in the shared package to prevent such things and
* assuming that actions are treated as immutable in the reducer

* I believe the `isDevelopment` variable is not used anymore. -> YAGNI or leave it there to make it easier for the next project if they need it?

There are still errors in development that only occur when trying to replay the state during time travel. Maybe we could write some tests for the reducers in the future (initialState + a lot of actions -> apply all of them synchronously and check whether 
1. no error occurs, 
2. the resulting state is the expected one).


shared/src/utils/is-development.ts
```diff
@@ -0,0 +1,10 @@
+// This environment is set at the backend and (manually) on the frontend, therefore we can use it in here
+declare global {
+    const process: {
+        env: {
+            NODE_ENV: 'development' | 'production';
```
Dassderdie, line 5:
```suggestion
            NODE_ENV: 'development' | 'production' | 'testing' | 'staging' | string;
```

shared/src/utils/is-development.ts
```diff
@@ -0,0 +1,10 @@
...
+        };
+    };
+}
+
+export const isDevelopment = process?.env?.NODE_ENV === 'development';
```
Dassderdie, line 10: We probably want to do the same things in `testing`

shared/src/store/reduce-exercise-state.ts
```diff
@@ -1,11 +1,15 @@
...
 import type { Mutable } from '../utils';
+import { isDevelopment } from '../utils/is-development';
 import type { ExerciseAction } from './action-reducers';
 import { getExerciseActionTypeDictionary } from './action-reducers';
 
 const exerciseActionTypeDictionary = getExerciseActionTypeDictionary();
 
+// See https://immerjs.github.io/immer/freezing for more information.
+setAutoFreeze(isDevelopment);
+
```
Dassderdie, line 12:
```suggestion
```

This was evil...

shared/src/store/reduce-exercise-state.ts
```diff
@@ -32,6 +36,10 @@ export function applyAction(
     draftState: Mutable<ExerciseState>,
     action: ExerciseAction
 ) {
+    if (isDevelopment) {
+        // Make sure that the action isn't mutated in the reducer
+        freeze(action, true);
+    }
```
Dassderdie, line 42:
```suggestion
    // Make sure that the action isn't mutated in the reducer
    freeze(action, true);
```

ClFeSc (commented):
Code lgtm, but I think something is broken with building the project...

ClFeSc:
>     * I believe the `isDevelopment` variable is not used anymore. -> YAGNI or leave it there to make it easier for the next project if they need it?

I'd say keep the `declare` but remove the `isDevelopment` variable. With the proper documentation, everyone can make their own variables from it, but we don't have to have a variable that is unused.

Dassderdie:
> Code lgtm, but I think something is broken with building the project...

The problem is that the eslint rule requires either eslint 7 and TS 4.5 or eslint 8 and ts 4.7 (or something along these lines). We have eslint 8 and ts 4.5. 
To update TS we would need to update angular (bigger pull request).

Locally I just used `npm i --force` and thought that `npm ci` would just strictly use the `package-lock.json`.

I see the following solutions:
* Downgrade eslint (no idea how much work this is)
* Merge this branch after all big updates are made (I wanted to make these updates after most pull requests are merged...)
* Temporarily add `--force` to the `npm ci` in the ci (and each of us would have to do this too, during local install)
* Split this pull request into two - one provides the functionality - the other adds the eslint rule...

ClFeSc:
> > Code lgtm, but I think something is broken with building the project...
> 
> The problem is that the eslint rule requires either eslint 7 and TS 4.5 or eslint 8 and ts 4.7 (or something along these lines). We have eslint 8 and ts 4.5. 
> To update TS we would need to update angular (bigger pull request).
> 
> Locally I just used `npm i --force` and thought that `npm ci` would just strictly use the `package-lock.json`.
> 
> I see the following solutions:
> * Downgrade eslint (no idea how much work this is)
> * Merge this branch after all big updates are made (I wanted to make these updates after most pull requests are merged...)
> * Temporarily add `--force` to the `npm ci` in the ci (and each of us would have to do this too, during local install)
> * Split this pull request into two - one provides the functionality - the other adds the eslint rule...
> 

Interesting... Locally, `npm run setup` installed the plugin and it worked correctly. In this case I'd say remove the plugin from this PR and add it later.

Dassderdie:
Closed in favor of #492 and #493

PR #449, opened 2022-06-13
Calculate statistics from history

Dassderdie:
* Instead of calculating the statistics in the reducer and saving them in the state, they are now generated on demand in the frontend by applying the actions in order.
* To not block the main thread if this takes longer I use the `sleep` method in the frontend too -> moved it from backend to shared (this method is also very handy for simple experiments).

TODO: 
* [x] Write a migration that removes the `statistics` from the `state` @ClFeSc


ClFeSc:
>     * [x]  Write a migration that removes the `statistics` from the `state` @ClFeSc

@Dassderdie Done.

ClFeSc (dismissed):

Dassderdie:
@ClFeSc Should this state then be version 4 or incorporated into version 3 (merge conflicts)?

ClFeSc:
> @ClFeSc Should this state then be version 4 or incorporated into version 3 (merge conflicts)?

Version 4. Don't mix state versions, this will only break.

ClFeSc (approved):

ClFeSc:
@Dassderdie I've merged `dev` into this branch. Merge when you like it.

PR #450, opened 2022-06-14
Prototypical addition of an automatic patient tray (Thesis of @ClFeSc)

ClFeSc:
This PR tracks the overall progress of @ClFeSc's thesis. It is not meant to be merged at the moment.


ClFeSc:
We'll probably only pull some things from this. The main part will remain as a draft and can serve as the basis of future work.

Issue #453, opened 2022-06-15
Only use node, not npm in docker-entrypoint.sh


anonym-HPI:
This should solve the second problem (dev container)
https://github.com/hpi-sam/digital-fuesim-manv/issues/451#issue-1271252069

`NODE_ENV=migration node --experimental-specifier-resolution=node ./node_modules/typeorm/cli -d dist/src/database/migration-datasource.js migration:run` 
instead of 
`npm run migration:run:prod`

PR #454, opened 2022-06-15
use node instead of npm in docker-entrypoint.sh

anonym-HPI:
fixes https://github.com/hpi-sam/digital-fuesim-manv/issues/453


ClFeSc (approved):
The code looks good, and we have tested it.

PR #463, opened 2022-07-18
Add progress indicator

Dassderdie:
image

I decided to put the indicator on the right instead of the left because:
1. No layout shift
3. two icons beside each other looks bad


ClFeSc:
When we're at it, maybe we could also disable the import functionality while another import is running? Currently, it leads to weird behavior when you import multiple exercises at once, and this is nearly never what you want. If there is demand for this I'd suggest a bulk import button (not yet). What do you think?

ClFeSc (approved):

PR #468, opened 2022-07-20
Keep all actions

### LINKED ISSUE #455: Actions get lost
When the server is saving the actions to the database and has not yet marked the exercise as saved but has already saved all present actions, due to the nature of await, a websocket propose action may sneak in and add an action that was not yet saved. When the fuesim server then marks the exercise as saved, all actions are removed, including the new.

ClFeSc:
Closes #455

### DIFF
--- a/backend/src/exercise/exercise-wrapper.ts
+++ b/backend/src/exercise/exercise-wrapper.ts
@@ -40,15 +40,19 @@ export class ExerciseWrapper extends NormalType<
         return this._changedSinceSave;
     }
 
+    private numberOfActionsToBeSaved = 0;
+
     /**
      * Mark this exercise as being up-to-date in the database
      */
     public markAsSaved() {
-        this._changedSinceSave = false;
-        this.temporaryActionHistory.splice(
-            0,
-            this.temporaryActionHistory.length
-        );
+        this.temporaryActionHistory.splice(0, this.numberOfActionsToBeSaved);
+        this._changedSinceSave = this.temporaryActionHistory.length > 0;
+        this.numberOfActionsToBeSaved = 0;
+    }
+
+    public markAsAboutToBeSaved() {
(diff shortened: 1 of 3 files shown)


backend/src/fuesim-server.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+                        alarmGroup: {
+                            alarmGroupVehicles: {},
+                            id: uuid(),
+                            name: 'Alarm Group',
+                        },
+                    },
+                    null
+                );
+                savedEntities.push(...entities);
+                return Promise.resolve(entities);
```
Dassderdie, line 38:
```suggestion
                return entities;
```
The async function automatically converts this to a promise -> the Promise.resolve is unnecessary

ClFeSc:
@Dassderdie I've renamed the variable in the `ExerciseWrapper` and added a comment for the test (and also fixed a logic error).

backend/src/exercise/exercise-wrapper.ts
```diff
@@ -40,15 +40,19 @@ export class ExerciseWrapper extends NormalType<
...
     /**
      * Mark this exercise as being up-to-date in the database
      */
     public markAsSaved() {
         this._changedSinceSave = false;
-        this.temporaryActionHistory.splice(
-            0,
-            this.temporaryActionHistory.length
-        );
+        this.temporaryActionHistory.splice(0, this.numberOfActionsToBeSaved);
```
Dassderdie, line 50: Shouldn't this be 
```ts
this.temporaryActionHistory.splice(0, this.numberOfActionsToBeSaved);
this._changedSinceSave = this.temporaryActionHistory.length > 0;
```
?

Dassderdie (changes requested):

Dassderdie (approved):
Code lgtm.

PR #473, opened 2022-07-22
Remove `[Exercise] Set Participant Id` action

### LINKED ISSUE #467: Set `participantId` in the constructor of an `ExerciseState`
...instead of using an action.

ClFeSc:
Closes #467


Dassderdie (changes requested):
The code looks good. I haven't tried it out, though.

But removing the `Add Participant Action` is a breaking change. -> Shouldn't there either be a migration/incompatibility notice stuff?

ClFeSc:
> The code looks good. I haven't tried it out, though.
> 
> But removing the `Add Participant Action` is a breaking change. -> Shouldn't there either be a migration/incompatibility notice stuff?

For reference: This waits for #476

ClFeSc:
I'll resolve the failing tests later today.

Issue #434, opened 2022-05-31
Inconsistent saving (button vs. autosave)


christianzoellner:
Some field have a save button (e.g. when setting the name of a transfer point or vehicle). Other fields save automatically (e.g. names and times of alarm group vehicles and the overall settings page).

Having it this inconsistent makes it _more likely_ that users _forget_ to save.

Dassderdie:
This has now been fixed in most places.

Issue #433, opened 2022-05-31
When adding targets to a transfer point, differenciate between "normal" transfer points and hospitals


christianzoellner:
....e.g. by having two tabs ("verbundene tranferpunkte" and "verbundene krankenhäuser") or two buttons ("Transfertpunkt hinzufügen" and "Krankenhaus hinzufügen"). In the latter case, the list should be sorted by hospital/non-hospital and have color-coding or subheadings in that regard.

PR #474, opened 2022-07-23
Feature/various UI improvements

Dassderdie:
Fixes:
* #434 (Autosaving is now debounced -> fewer actions; and doesn't propose invalid actions)
* #433 
* Always show transfer point dropdown

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/transfer-point-feature-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/transfer-point-feature-manager.ts
@@ -85,79 +85,7 @@ export class TransferPointFeatureManager extends ElementFeatureManager<TransferP
         ) {
             return false;
         }
-        const reachableTransferPointIds = Object.keys(
-            droppedOnTransferPoint.reachableTransferPoints
-        );
-
-        const reachableHospitalIds = Object.keys(
-            droppedOnTransferPoint.reachableHospitals
-        );
-
-        const proposeTransfer = (
-            targetId: UUID,
-            targetType: 'hospital' | 'transferPoint'
-        ) => {
-            if (targetType === 'hospital') {
-                this.apiService.proposeAction(
-                    {
-                        type: '[Hospital] Transport patient to hospital',
-                        hospitalId: targetId,
-                        vehicleId: droppedElement.value.id,
-                    },
-                    true
-                );
-                return;
-            }
-            this.apiService.proposeAction(
-                {
-                    type: '[Transfer] Add to transfer',
-                    // TODO: The type should already be correct
-                    elementType:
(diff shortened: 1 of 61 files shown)


frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/transfer-point-feature-manager.ts
```diff
@@ -85,79 +85,7 @@ export class TransferPointFeatureManager extends ElementFeatureManager<TransferP
...
-            if (
-                reachableTransferPointIds.length === 0 &&
-                reachableHospitalIds.length === 1
-            ) {
-                proposeTransfer(reachableHospitalIds[0], 'hospital');
-                return true;
-            }
-        }
-
+        // We always want to show the chooseTransferPopup to
```
Dassderdie, line 88:
```suggestion
        // Always show the chooseTransferPopup
```

shared/src/store/action-reducers/transfer-point.ts
```diff
@@ -141,8 +143,13 @@ export namespace TransferPointActionReducers {
                     'transferPoints',
                     transferPointId
                 );
-                transferPoint.internalName = internalName;
-                transferPoint.externalName = externalName;
+                // Empty strings are ignored
```
ClFeSc, line 146: This is a breaking change. Could this break something? Probably not. Do we want to risk it?

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/transfer-point-popup/transfer-point-popup.component.html
```diff
@@ -185,8 +241,6 @@ <h5 class="popover-header">
                     </div>
```
ClFeSc, line 241: Making the comment here as GitHub doesn't allow comments on all lines: It's probably not new, but I just noticed that the transfer time to the hospital in the transfer point popup is slightly offset to the top. Maybe we can fix this here?

ClFeSc (commented):
Just very little things. Nice work!

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/transfer-point-popup/transfer-point-popup.component.html
Dassderdie: Current:
image
No monospace
image
Both monospace
image

The offset is due to the font being monospace. Without monospace different numbers above each other don't look good. -> I let it be like it is now.

ClFeSc (approved):

ClFeSc (approved):

Dassderdie:
@ClFeSc I made the pretriage table smaller and removed one unnecessary check in a function. 

| Before | After |
| --- | --- |
| image | image |

I don't think this requires another review.

PR #476, opened 2022-07-24
Fix state migrations

ClFeSc:
Fixed:

* Not awaiting updates to the database
* Impossibility of updating an `ExerciseWrapper`
* Impossibility of importing an outdated export

An example of a migration can be seen in #477

### DIFF
--- a/backend/src/database/state-migrations/impossible-migration.ts
+++ b/backend/src/database/state-migrations/impossible-migration.ts
@@ -1,8 +1,8 @@
 import type { UUID } from 'digital-fuesim-manv-shared';
 import { RestoreError } from '../../utils/restore-error';
-import type { MigrationSpecification } from './migrations';
+import type { Migration } from './migrations';
 
-export const impossibleMigration: MigrationSpecification = {
+export const impossibleMigration: Migration = {
     actions: (initialState, actions) => {
         throw new RestoreError(
             'The migration is not possible',
--- a/backend/src/database/state-migrations/impossible-migration.ts
+++ b/backend/src/database/state-migrations/impossible-migration.ts
@@ -1,0 +1,20 @@
+import { RestoreError } from '../../utils/restore-error';
+import type { MigrationFunctions } from './migrations';
+
+export const impossibleMigration: MigrationFunctions = {
+    database: (_entityManager, exerciseId) => {
+        throw new RestoreError('The migration is not possible', exerciseId);
+    },
+    inMemory: (exerciseWrapper) => {
+        throw new RestoreError(
+            'The migration is not possible',
+            exerciseWrapper.id ?? 'unknown id'
+        );
+    },
+    stateExport: (stateExport) => {
+        throw new RestoreError(
+            'The migration is not possible',
+            stateExport.currentState.id ?? 'unknown id'
+        );
+    },
+};
--- a/backend/src/database/state-migrations/impossible-migration.ts
+++ b/backend/src/database/state-migrations/impossible-migration.ts
@@ -1,14 +1,18 @@
+import type { UUID } from 'digital-fuesim-manv-shared';
 import { RestoreError } from '../../utils/restore-error';
-import type { Migrations } from './migrations';
+import type { MigrationSpecification } from './migrations';
 
-export const impossibleMigration: Migrations = {
-    database: (_entityManager, exerciseId) => {
-        throw new RestoreError('The migration is not possible', exerciseId);
-    },
-    stateExport: (stateExport) => {
+export const impossibleMigration: MigrationSpecification = {
+    actions: (initialState, actions) => {
         throw new RestoreError(
             'The migration is not possible',
-            stateExport.currentState.id ?? 'unknown id'
+            (initialState as { id?: UUID }).id ?? 'unknown'
+        );
+    },
+    state: (state) => {
+        throw new RestoreError(
+            'The migration is not possible',
+            (state as { id?: UUID }).id ?? 'unknown'
         );
     },
 };
--- a/backend/src/database/state-migrations/impossible-migration.ts
+++ b/backend/src/database/state-migrations/impossible-migration.ts
@@ -1,15 +1,9 @@
 import { RestoreError } from '../../utils/restore-error';
-import type { MigrationFunctions } from './migrations';
+import type { Migrations } from './migrations';
 
-export const impossibleMigration: MigrationFunctions = {
+export const impossibleMigration: Migrations = {
     database: (_entityManager, exerciseId) => {
         throw new RestoreError('The migration is not possible', exerciseId);
-    },
-    inMemory: (exerciseWrapper) => {
-        throw new RestoreError(
-            'The migration is not possible',
-            exerciseWrapper.id ?? 'unknown id'
-        );
     },
     stateExport: (stateExport) => {
         throw new RestoreError(
--- a/backend/src/database/state-migrations/migrations.ts
+++ b/backend/src/database/state-migrations/migrations.ts
@@ -1,39 +1,40 @@
-import type { StateExport, UUID } from 'digital-fuesim-manv-shared';
+import type {
+    ExerciseState,
+    StateExport,
+    UUID,
+} from 'digital-fuesim-manv-shared';
 import type { EntityManager } from 'typeorm';
+import { RestoreError } from '../../utils/restore-error';
+import { ActionWrapperEntity } from '../entities/action-wrapper.entity';
+import { ExerciseWrapperEntity } from '../entities/exercise-wrapper.entity';
 import { impossibleMigration } from './impossible-migration';
 
 /**
- * Such a function MUST update the initial state of the exercise with the provided {@link exerciseId} as well as every action associated with it from its current state version to the next version in a way that they are valid states/actions.
- * It MAY throw a {@link RestoreError} in a case where upgrading is impossible and a terminal incompatibility with older exercises is necessary.
- * It MUST update the respective updates to the exercise and its associated objects in the database.
- * All database interaction MUST use the provided {@link EntityManager}.
+ * Such a function gets the already migrated initial state of the exercise and an array of all actions (not yet migrated).
+ * It is expected that afterwards the actions in the provided array are migrated.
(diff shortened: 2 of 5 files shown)


backend/src/database/state-migrations/impossible-migration.ts
```diff
@@ -0,0 +1,20 @@
+import { RestoreError } from '../../utils/restore-error';
+import type { MigrationFunctions } from './migrations';
+
+export const impossibleMigration: MigrationFunctions = {
```
Dassderdie, line 4: I would rename all the variables, classes, interfaces, etc. from `...MigrationFunctions` to `...Migrations`

backend/src/exercise/exercise-wrapper.ts
```diff
@@ -243,12 +243,23 @@ export class ExerciseWrapper extends NormalType<
...
+        update: (
+            initialState: unknown,
+            currentState: unknown
+        ) => { initialState: unknown; currentState: unknown }
+    ): void {
+        const result = update(this.initialState, this.currentState);
+        this.initialState = result.initialState as ExerciseState;
+        this.currentState = result.currentState as ExerciseState;
+    }
+
```
Dassderdie, line 262: Where is this function used, why does it get an unknown update function, and why does it have to be in the exerciseWrapper (god class)?

Dassderdie (changes requested):

backend/src/exercise/exercise-wrapper.ts
ClFeSc: I just noticed that we don't actually need the whole "migrate in memory", so I've removed the respective code. For reference, still the answers to the questions:

1. Why does it have to be in `ExerciseWrapper`? We have to assign to `private` fields.
2. It is (was) used in an actual migration that was added in #477.
3. `unknown` because the idea is (was) to use it to migrate between state versions. Let's say we're migrating from version `7` to `8` when the current version is `9`. `9` added `someMysticalProperty` to the state. When we say these are `ExerciseState`s, TS would allow us to access `someMysticalProperty` although we don't have it as we're using an older state version at this point. I agree that this may not be the best solution. A similar case arises in #477, so let's discuss it further over there, if necessary. The only solution I'd see for this would be to include `declare` types for outdated state versions, which in itself is not great.

backend/src/database/state-migrations/migrations.ts
```diff
@@ -1,31 +1,30 @@
...
-    exerciseWrapper: ExerciseWrapper
-) => Promise<ExerciseWrapper>;
+export type StateExportMigration = (
+    stateExport: StateExport
+) => Promise<StateExport>;
 
-export interface MigrationFunctions {
-    database: DbMigrationFunction;
-    inMemory: InMemoryMigrationFunction;
+export interface Migrations {
```
Dassderdie, line 25: This gets confusing when having, e.g., this code:
```ts
const migrations: {
    [key: number]: Migrations;
};
```

Instead, I would just call this `Migration`. A `Migration` consists of a `DbMigration` and a `StateExportMigration`. If you want to, you could also call these `DbMigrationFunction` and  `StateExportMigrationFunction` to avoid the naming scheme being misunderstood to imply an inheritance relation.

backend/src/database/state-migrations/impossible-migration.ts
```diff
@@ -0,0 +1,14 @@
...
+export const impossibleMigration: Migrations = {
+    database: (_entityManager, exerciseId) => {
+        throw new RestoreError('The migration is not possible', exerciseId);
+    },
+    stateExport: (stateExport) => {
+        throw new RestoreError(
+            'The migration is not possible',
+            stateExport.currentState.id ?? 'unknown id'
+        );
+    },
```
Dassderdie, line 13: Is it really necessary to have two different migration functions?

Would it be possible to do this with only one, by exporting the stuff saved in the database and migrating it with the same function to the new state?

The problem I currently see is that nobody is going to write migrations because they are too much work. -> Premature optimization.

Dassderdie (changes requested):

backend/src/database/state-migrations/migrations.ts
Dassderdie: And yes - this is a contradiction to my comment before ^^

backend/src/database/state-migrations/impossible-migration.ts
ClFeSc: I've completely reworked it now. It now uses two methods, one for states and one for actions. With this, I'm handling the loading/writing to the database in the util method, and people who write migrations only have to write the absolute minimum. Do you like this approach better?

backend/src/database/state-migrations/migrations.ts
```diff
@@ -55,25 +45,145 @@ export async function migrateInDatabaseTo(
...
+                stateExport.history!.actionHistory
+            );
+        });
+        stateExport.history.actionHistory.filter((action) => action !== null);
     }
-    return currentExercise;
+    return stateExport;
+}
+
+function getRelevantFunctions(
```
Dassderdie, line 166:
```suggestion
function getMigrationFunctions(
```

backend/src/database/state-migrations/migrations.ts
```diff
@@ -55,25 +45,145 @@ export async function migrateInDatabaseTo(
...
+    initialVersion: number,
+    targetVersion: number
+): { actions: MigrateActionsFunction[]; state: MigrateStateFunction[] } {
+    const functions = Object.entries(migrations)
```
Dassderdie, line 170:
```suggestion
    const necessaryMigrations = Object.entries(migrations)
```

backend/src/database/state-migrations/migrations.ts
```diff
@@ -55,25 +45,145 @@ export async function migrateInDatabaseTo(
...
+        .filter(
+            ([key]) =>
+                Number.parseInt(key) > initialVersion &&
+                Number.parseInt(key) <= targetVersion
+        )
+        .map(([, migration]) => migration);
+    return {
+        actions: functions
+            .filter((thisFunctions) => thisFunctions.actions !== null)
```
Dassderdie, line 179:
```suggestion
            .filter((migration) => migration.actions !== null)
```

backend/src/database/state-migrations/migrations.ts
```diff
@@ -55,25 +45,145 @@ export async function migrateInDatabaseTo(
...
+            .map(
+                (thisFunctions) => thisFunctions.actions
```
Dassderdie, line 181:
```suggestion
                (migration) => migration.actions
```

backend/src/database/state-migrations/migrations.ts
```diff
@@ -55,25 +45,145 @@ export async function migrateInDatabaseTo(
...
-        currentExercise = await migrations[currentVersion]!.inMemory(
-            currentExercise
-        );
+    stateExport: StateExport
+): StateExport {
+    const migrationFunctions = getRelevantFunctions(
+        currentStateVersion,
+        targetStateVersion
+    );
+    migrationFunctions.state.forEach((fn) => {
```
Dassderdie, line 144:
```suggestion
    migrationFunctions.state.forEach((migrateStateFunction) => {
```

backend/src/database/state-migrations/migrations.ts
```diff
@@ -55,25 +45,145 @@ export async function migrateInDatabaseTo(
...
+        patch.initialStateString = JSON.stringify(initialState);
+        patch.currentStateString = JSON.stringify(currentState);
+    }
+    await entityManager.update(
+        ExerciseWrapperEntity,
+        { id: exerciseId },
+        patch
+    );
+    if (actions !== null) {
+        let index = 0;
```
Dassderdie, line 102: `patchedActionsIndex`?

backend/src/database/state-migrations/migrations.ts
```diff
@@ -55,25 +45,145 @@ export async function migrateInDatabaseTo(
...
+        const indicesToRemove: number[] = [];
+        const actionsToUpdate: [number, string][] = [];
```
Dassderdie, line 104:
```suggestion
        const actionsToUpdate: { index: number, actionString: string }[] = [];
```

backend/src/database/state-migrations/migrations.ts
```diff
@@ -55,25 +45,145 @@ export async function migrateInDatabaseTo(
...
+        }
+        if (actionsToUpdate.length > 0) {
+            await Promise.all(
+                actionsToUpdate.map(async (action) =>
+                    entityManager.update(
+                        ActionWrapperEntity,
+                        { index: action[0] },
+                        { actionString: action[1] }
+                    )
+                )
```
Dassderdie, line 129:
```suggestion
                actionsToUpdate.map(async ({ index, actionString }) =>
                    entityManager.update(
                        ActionWrapperEntity,
                        { index },
                        { actionString }
                    )
                )
```

backend/src/database/state-migrations/migrations.ts
```diff
@@ -1,52 +1,42 @@
...
  */
-export type InMemoryMigrationFunction = (
-    exerciseWrapper: ExerciseWrapper
-) => Promise<ExerciseWrapper>;
+type MigrateStateFunction = (state: object) => object;
 
-export interface MigrationFunctions {
-    database: DbMigrationFunction;
-    inMemory: InMemoryMigrationFunction;
+export interface MigrationSpecification {
```
Dassderdie, line 30:
```suggestion
export interface Migration {
```

I would still only call this a `Migration`. An interface is always some sort of Specification and the inheritance naming problem doesn't apply anymore because they are now called `Migrate...Function`.

You can then use:

```ts
const migrations: Migration[];
const migration: Migration;
const MigrateStateFunctions: MigrateStateFunction[];
...
```

Dassderdie (changes requested):
I like this approach much more. All review comments are about naming.

ClFeSc:
@Dassderdie Do you want to resolve the conversations?

Dassderdie:
> @Dassderdie Do you want to resolve the conversations?

I reviewed via the GitHub-App -> unresolved conversations are there not easily spottable -> I forgot ^^

PR #480, opened 2022-07-24
Notifications

### LINKED ISSUE #334: Notification when participant joins
There should be a notification to all trainers when a new participant joins so that the trainer can assign the participant to the exercise.

Dassderdie:
* The messages components have been refactored (merged the `toasts` and `alerts` components to the `display-messages` component, updated the documentation, fixed the CSS to reflect the documentation correctly)
* The API of the `handleChanges` function has been improved
* Messages can now also be seen when you are in fullscreen
* Fixes #334 (and leaves)
* Clients restricted to a viewport (and in the exercise) now get notifications when a vehicle appears in one viewport
* There are no notifications for personnel because the implementation would be quite difficult. There is no easy way to see whether the personnel only appeared because they left a vehicle.


ClFeSc (approved):
Looks good. One question came to my mind: Do we want to show notifications of vehicle arrival to participants while the exercise is paused? They can't interact with the map at this time. On the other hand, this situation will be very rare, as most arrivals will be by transfers which are stopped while paused. So approval either way.

Dassderdie:
> Looks good. One question came to my mind: Do we want to show notifications of vehicle arrival to participants while the exercise is paused? They can't interact with the map at this time. On the other hand, this situation will be very rare, as most arrivals will be by transfers which are stopped while paused. So approval either way.

When the exercise is paused you can still see the map and the vehicles - therefore I would also show the notification. If you don't want to show it, you would send the user back to the waiting room.

PR #481, opened 2022-07-24
Add test checking whether a fresh state is correct

ClFeSc:
This was bugging me for months. With this, we should at least be somewhat safe (even more if we ever do #125).

### DIFF
--- a/shared/src/state.spec.ts
+++ b/shared/src/state.spec.ts
@@ -2,7 +2,8 @@ import { ExerciseState } from './state';
 import { validateExerciseState } from './store';
 
 describe('ExerciseState', () => {
-    it('validates a fresh state', () => {
+    // If this fails, either the created state is invalid, or a validator is incorrect/missing in a model.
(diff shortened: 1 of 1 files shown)


shared/src/state.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { ExerciseState } from './state';
+import { validateExerciseState } from './store';
+
+describe('ExerciseState', () => {
+    it('validates a fresh state', () => {
```
Dassderdie, line 5:
```suggestion
    // If this fails, either the created state is invalid, or a validator is incorrect/missing in a model.
    it('creates a valid state', () => {
```

Dassderdie (approved):
A simple test that is very useful.  I like it.

shared/src/state.spec.ts
Dassderdie: This is optional for you to decide - approval from me either way.

PR #483, opened 2022-07-25
Increase lint timeout to 3 mins (from 2)

ClFeSc:
This keeps failing in the last few days. Not sure why, probably our code base is too large now.


Dassderdie (approved):
LGTM

PR #490, opened 2022-07-25
Patient model (from bacherlor thesis)

### LINKED ISSUE #335: Allow changing the treatment factor
Currently, all patients have a treatment factor of `1`. Trainers should be able to adjust these, either per patient, per viewport, or per exercise.

### LINKED ISSUE #402: Change Patientsystem from life-point-based to state-based

hpistudent72:
Closes #335 
Closes #402


frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,305 @@
...
+                return { latestTime: -1, matchingHealthStateId: 'Phase1State' };
+        }
+    }
+    throw new Error('Wrong condition format in SK entries');
+}
+
+/**
+ * Creates a new State in healthStateData if possible and returns the id of that state
+ */
+function createHealthState(
```
ClFeSc, line 292: Rename this method. I expect the whole `HealthState` from such a name.

shared/src/data/default-state/patient-templates.ts
```diff
@@ -1,1534 +1,1605 @@
...
+                ),
+            },
+            defaultMaleImage,
+            'redInstantTransportState'
+        ),
+    ]),
+];
 
-const phaseTime = 12 * 60 * 1000;
+// const noChangesState = PatientHealthState.create(
```
ClFeSc, line 99: Remove all the commented-out templates.

shared/src/models/exercise-configuration.ts
```diff
@@ -9,6 +9,9 @@ export class ExerciseConfiguration {
     @IsBoolean()
     public readonly bluePatientsEnabled: boolean = false;
 
+    @IsNumber()
```
ClFeSc, line 12:
```suggestion
    @IsNumber()
    @Min(0)
```

shared/src/models/patient-health-state.ts
```diff
@@ -62,54 +21,86 @@ export class ConditionParameters {
...
     }
 
     static readonly create = getCreate(this);
 }
 
 export class PatientHealthState {
-    @IsUUID(4, uuidValidationOptions)
-    public readonly id: UUID = uuid();
+    @IsString()
+    public readonly id: string;
```
ClFeSc, line 96: Rename to `name`. This should also include all variables and properties named after this field, but I'd be fine with letting them as they are for sake of the remaining time.

shared/src/models/patient-template.ts
```diff
@@ -24,41 +17,30 @@ export class PatientTemplate {
...
-    public readonly pretriageInformation: PretriageInformation;
-
     @ValidateNested()
     @Type(() => ImageProperties)
     public readonly image: ImageProperties;
 
     @IsDefined()
     public readonly healthStates: {
-        readonly [stateId: UUID]: PatientHealthState;
+        readonly [stateId: string]: PatientHealthState;
```
ClFeSc, line 26:
```suggestion
        readonly [stateName: string]: PatientHealthState;
```

shared/src/models/patient-template.ts
```diff
@@ -24,41 +17,30 @@ export class PatientTemplate {
...
     };
 
-    @IsUUID(4, uuidValidationOptions)
-    public readonly startingHealthStateId: UUID;
-
-    @IsValidHealthPoint()
-    public readonly health: HealthPoints;
+    @IsString()
+    public readonly startingHealthStateId: string;
```
ClFeSc, line 30:
```suggestion
    public readonly startingHealthStateName: string;
```

shared/src/models/patient.ts
```diff
@@ -110,21 +90,14 @@ export class Patient {
 
     @IsDefined()
     public readonly healthStates: {
-        readonly [stateId: UUID]: PatientHealthState;
+        readonly [stateId: string]: PatientHealthState;
```
ClFeSc, line 93:
```suggestion
        readonly [stateName: string]: PatientHealthState;
```

shared/src/models/patient.ts
```diff
@@ -110,21 +90,14 @@ export class Patient {
...
-    public readonly currentHealthStateId: UUID;
-    /**
-     * See {@link HealthPoints} for context of this property.
-     */
-    @IsNumber()
-    @Max(healthPointsDefaults.max)
-    @Min(healthPointsDefaults.min)
-    public readonly health: HealthPoints;
+    @IsString()
+    public readonly currentHealthStateId: string;
```
ClFeSc, line 100:
```suggestion
    public readonly currentHealthStateName: string;
```

shared/src/store/action-reducers/configuration.ts
```diff
@@ -28,6 +28,14 @@ export class SetBluePatientsEnabledFlagAction implements Action {
     public readonly bluePatientsEnabled!: boolean;
 }
 
+export class SetGlobalPatientChangeSpeedAction implements Action {
+    @IsString()
+    public readonly type = '[Configuration] Set globalPatientChangeSpeed';
+
+    @IsNumber()
```
ClFeSc, line 35:
```suggestion
    @IsNumber()
    @Min(0)
```

shared/src/store/action-reducers/exercise.ts
```diff
@@ -100,10 +99,24 @@ export namespace ExerciseActionReducers {
             patientUpdates.forEach((patientUpdate) => {
                 const currentPatient = draftState.patients[patientUpdate.id];
                 currentPatient.currentHealthStateId = patientUpdate.nextStateId;
-                currentPatient.health = patientUpdate.nextHealthPoints;
                 currentPatient.stateTime = patientUpdate.nextStateTime;
                 currentPatient.treatmentTime = patientUpdate.treatmentTime;
-                currentPatient.realStatus = getStatus(currentPatient.health);
+                if (currentPatient.treatmentHistory.length === 0) {
+                    for (let i = 0; i <= 60 / tickInterval; i++) {
```
ClFeSc, line 105: Are you sure that this condition is correct? `tickInterval` is in ms.

ClFeSc (changes requested):
I haven't tested it out yet.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/patient-popup/patient-popup.component.html
```diff
@@ -98,7 +102,36 @@ <h5 class="popover-header">
...
-                                        &times;{{ patient.timeSpeed }}
+                                        <select
+                                            [ngModel]="patient.changeSpeed"
+                                            (ngModelChange)="
+                                                setPatientChangeSpeed($event)
+                                            "
+                                            class="form-select d-inline-block no-validation"
+                                            style="max-width: fit-content"
+                                        >
+                                            <option [value]="0.1">
```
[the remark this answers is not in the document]
hpistudent72: Not yet. And since there are cases, where patients still can change even with a speed of zero, I would argue this should be a new issue.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/patient-popup/patient-popup.component.html
```diff
@@ -148,61 +188,95 @@ <h5 class="popover-header">
...
+                                    {{ pretriageInformation.dDisability.eyes }}
+                                    Spr.
+                                    {{
+                                        pretriageInformation.dDisability
+                                            .speaking
+                                    }}
+                                    Mo.
+                                    {{
+                                        pretriageInformation.dDisability.motoric
+                                    }}
```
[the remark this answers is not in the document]
hpistudent72: Yes these abbreviations are directly taken from the patient cards in the FüSim MANV

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,305 @@
...
+            });
+        } else {
+            currentCategory.templates.push(patientTemplate);
+        }
+    }
+
+    return patientData;
+}
+
+type ColorCategory = 'black' | 'blue' | 'green' | 'red' | 'yellow';
```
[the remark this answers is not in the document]
hpistudent72: We have a similar type that includes white as a color

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,305 @@
...
+        );
+    }
+
+    for (let i = 0; i < phaseChanges; i++) {
+        const conditionColor = colorRegExp.exec(
+            healthStateInformation[3 * i + 2]
+        );
+        const nextPhaseColor = getColor(healthStateInformation[3 * i + 3]);
+        // Finish all States and transfer them to two new states based on the condition
+        if (conditionColor!) {
```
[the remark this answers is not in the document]
hpistudent72: Because of the way the regex works. If it doesn't find anything you don't get an empty object but rather nothing at all

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/patient-popup/patient-popup.component.html
```diff
@@ -98,7 +102,36 @@ <h5 class="popover-header">
                                         Verletzungs-Geschwindigkeit:
                                     </th>
                                     <td class="font-monospace">
-                                        &times;{{ patient.timeSpeed }}
+                                        <select
```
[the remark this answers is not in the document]
hpistudent72: That would mean writing code, just to end up with a self implemented select input.

frontend/src/app/pages/exercises/exercise/shared/exercise-settings/exercise-settings-modal/exercise-settings-modal.component.html
```diff
@@ -75,6 +75,21 @@ <h4 class="modal-title">Einstellungen</h4>
                     type="checkbox"
                 />
             </div>
+            Globaler Verletzungsgeschwindigkeit-Modifikator
+            <select
```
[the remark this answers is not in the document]
hpistudent72: Same as above ("That would mean writing code, just to end up with a self implemented select input.")

backend/src/exercise/patient-ticking.ts
```diff
@@ -48,29 +37,37 @@ export function patientTick(
...
+                              patient.changeSpeed *
+                              state.configuration.globalPatientChangeSpeed
                         : 0;
                 return {
                     id: patient.id,
-                    nextHealthPoints,
                     nextStateId,
                     nextStateTime,
                     treatmentTime,
+                    newTreatment,
```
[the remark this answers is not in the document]
hpistudent72: Yes you're right. The intention was that newTreatment isn't modfied though

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/patient-popup/patient-popup.component.html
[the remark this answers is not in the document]
ClFeSc: @Dassderdie What do you think?

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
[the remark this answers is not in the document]
ClFeSc: Ah, then use `type ColorCategory = Exclude<{Type from Shared}, 'white'>` here.

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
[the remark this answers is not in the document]
ClFeSc: I know. But `conditionColor !== null` would be much better. With the `!` you're indicating that, regardless of the actual type, `conditionColor` is *never* `null` or `undefined`, which is definitely not the case here.

ClFeSc (commented):
I still haven't tried it out, but the code looks much better now.

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,305 @@
...
+    aspectRatio: 1,
+};
+
+export function parsePatientData(importString: string) {
+    const patientData: PatientData[] = [];
+
+    const splitString = importString.split('|');
+    // first 106 entries are the headers
+    for (let i = 106; i + 100 < splitString.length; i += 106) {
+        // 0, 1, 2 are orginasational data for anaolg training
```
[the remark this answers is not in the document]
ClFeSc: We decided that this takes too much time for now.

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.html
```diff
@@ -136,6 +136,18 @@ <h5>
                             height: 240px;
                         "
                     >
+                        <label class="btn btn-outline-primary ms-5">
```
ClFeSc, line 139: This input is below the other buttons and texts in this row:
image

This breaks the whole layout for me when there are more patients:
image

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,296 @@
...
+    // first 106 entries are the headers
+    for (let i = columns; i + columns - 1 < splitString.length; i += columns) {
+        // 0, 1, 2 are orginasational data for anaolg training
+        // 3 and 4 are personalInformation that will be randomly generated
+        const biometricInformation = BiometricInformation.create(
+            splitString[i + 7]!,
+            Number(splitString[i + 5]),
+            splitString[i + 6] === 'M'
+                ? 'male'
+                : splitString[i + 6] === 'F'
```
ClFeSc, line 41:
```suggestion
                : splitString[i + 6] === 'W'
```

Or something like this; currently, all patients are either male or undefined.

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,296 @@
...
+                ? 'female'
+                : 'diverse'
+        );
+
+        // 8, 9 are more organisational data
+        const pretriageInformation = PretriageInformation.create(
+            splitString[i + 28]!,
+            splitString[i + 29]! + splitString[i + 30]!,
+            !(splitString[i + 56] === 'G'),
```
ClFeSc, line 50: This seems to be inverted as all green patients are non-walkable and all yellow and red patients are walkable.

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,296 @@
...
+function getColor(colorString: string): ColorCategory {
+    if (colorString === '') {
+        return 'black';
+    }
+    const colorRegExp = /grün|gelb|rot|schwarz|blau/u;
+    const color = colorRegExp.exec(colorString.toLowerCase());
+    if (color === null) {
+        throw new Error('Wrong color format in SK entries');
+    }
+    return color[0] as ColorCategory;
```
ClFeSc, line 236: You have to translate the entries here, as otherwise, the types don't match up.

ClFeSc (changes requested):
Please fix these visual bugs (especially the one with the colors), then I can review this further.

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.html
ClFeSc: Now the button doesn't work at all. And I'm not sure I like that the text is less descriptive now.

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -233,6 +233,19 @@ function getColor(colorString: string): ColorCategory {
...
+            return 'green';
+        case 'gelb':
+            return 'yellow';
+        case 'rot':
+            return 'red';
+        case 'schwarz':
+            return 'black';
+        case 'blau':
+            return 'blue';
+        default:
```
ClFeSc, line 247: Maybe throw something here? Or use `as 'blau' | 'gelb' | 'grün' | 'rot' | 'schwarz'` in the `switch (color[0])` line?  
But don't have an empty `default` and then just return something you know will never match. Probably `as` is the better solution. @Dassderdie Any thoughts on this?

Dassderdie: Instead of a switch statement you should use an object literal as a map.
This should then result in something like:
```ts
colorTranslationMap[colors[0]]!;
```

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,311 @@
...
+    constructor(
+        public name: string,
+        public status: ColorCategory,
+        public conditions: ConditionParameters[],
+        public startingPhase: number,
+        public finished: boolean
+    ) {}
+}
+
+function generateHealthStates(
```
ClFeSc, line 126: Something about the generation of the condition parameters seems to be wrong, only two of the templates of the CSV actually have conditions. For this reason, the patients don't change their state.

ClFeSc (changes requested):

shared/src/models/patient.ts
```diff
@@ -139,21 +119,22 @@ export class Patient {
...
+    public readonly changeSpeed: number = 1;
 
     /**
-     * This can be any arbitrary string. It gives trainers the freedom to add additional functionalities that are not natively supported by this application (like an hospital ticket system)
+     * The time a patient has been treated for
      */
-    @IsString()
-    @MaxLength(65535)
-    public readonly remarks: string;
-
```
Dassderdie, line 150: This should not get deleted

ClFeSc: I think this only got moved and not removed.

anonym-HPI:
@Dassderdie @ClFeSc One of you interested in continuing this? I could also try to get this done.

ClFeSc:
I could try to do it.

anonym-HPI:
> I could try to do it.

Ok, if you need help, just write.

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.html
```diff
@@ -90,6 +90,18 @@ <h5>Ansicht/Transferpunkt</h5>
...
                         Patienten
+                        <div class="btn btn-outline-primary ms-5">
+                            <i class="bi-cloud-upload me-2"> </i>
+                            CSV-Import
+                            <input
+                                (input)="
+                                    importPatientsCSV($any($event.target).files)
+                                "
+                                type="file"
+                                accept="application/csv"
```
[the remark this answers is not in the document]
Dassderdie: `text/csv` should work (https://en.wikipedia.org/wiki/Media_type#Common_examples_[10])

lukasrad02:
@ClFeSc are you still working on this? This feature would be interesting for our future development.

We could also try to support you on this, this way we could also get familiar with the new features quickly.

shared/src/store/action-reducers/exercise.ts
```diff
@@ -80,11 +79,26 @@ export namespace ExerciseActionReducers {
...
+                            notSan: 0,
+                            rettSan: 0,
+                            san: 0,
+                        });
+                    }
+                }
+                currentPatient.treatmentHistory.shift();
+                currentPatient.treatmentHistory.push(
+                    patientUpdate.newTreatment
+                );
```
Dassderdie, line 101: This is probably a big performance problem as this means that for every tick for every patient, a new array with 60 items is created (immutability via the immer.js proxy).

Here is a proposal for a better solution:

This value probably doesn't change very often. Therefore, one could compress intervals with the same treatment status.

```ts
interface Interval {
    /**
     * Corresponds to the exerciseTime
     */
    startTime: number,
    value: { gf: number, .... }
}

class TreatmentHistory {
    // in seconds
    static readonly maxLength = 60;

    // If the newTreatment is different to the treatment in the last interval, add it (with the current exerciseTime) 
    // and remove all intervals that are older than (exerciseTime - maxLength)
    // else do nothing
    static addTreatment(intervalls: Intervall[], newTreatment: {gf: number, ...}, exerciseTime)

    static getAverageTreatment(...)
}
```

This would reduce write time (expensive due to immutability) and slightly increase read time (not done very often).

One could also either remove the PatientUpdates from the Tick action or at least only send those for patients that changed their treatment. This would reduce the state-history size by a lot and could either slightly reduce or increase the performance (more calculation on the client vs. less parsing, reduced usage of the WebSocket connection, and less overhead on the server). This would also have big repercussions on the migrations.


Guidelines: "Premature optimization is the root of all evil.", Use the benchmark script.

ClFeSc:
> @ClFeSc are you still working on this? This feature would be interesting for our future development.
> 
> We could also try to support you on this, this way we could also get familiar with the new features quickly.

Generally, yes. However, currently, I'm a bit short of time. But I'll look into resolving the merge conflicts this weekend, and we can see where we'll go from there.

ClFeSc:
There is now #614 that further tracks the progress of this. Note that the remarks and comments made here are currently still relevant. Any further remarks and comments are added to #614, please.

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.html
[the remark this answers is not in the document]
ClFeSc: Fixed in #614.

PR #494, opened 2022-07-26
Change the patientCapacity of an RTW to 1

Dassderdie:


Dassderdie:
@christianzoellner Is it correct that a `KTW` should be able to transport two patients?

christianzoellner:
it depends. 

In the context of disaster assistance there are special KTWs for two patients. This would be useful for the Malteser.

In the context of regular emergency management (as practiced by the BABZ) regular KTWs have only one slot.

Is it maybe possible to have two types of KTW, e.g. `KTW` and `KTW (KatSchutz)`, with the former having 1 and the latter having 2 patients?

Dassderdie:
@ClFeSc TODO: 
* write migration for patientCapacity

ClFeSc (commented):
Changes look good, migration missing.

ClFeSc (approved):

PR #496, opened 2022-07-27
Update personnel images

### LINKED ISSUE #487: Show personnel type
Currently, nobody knows which personnel-image represents which personnel type.

Dassderdie:
Partially fixes #487 

These are the updated images (at the maximum zoom level we allow):

image

This solution is not perfect, but good enough.


ClFeSc (approved):
Not tested, but image looks good.

Issue #500, opened 2022-07-28
Vehicles can have different images depending on how many patients are loaded in


anonym-HPI:

christianzoellner:
This could be generalized to a "elements have different states with different images". 
E.g. it would be useful to see whether a vehicle is full or parked, and this state could be used for permissions (i.e. can the vehicle be moved?) during an exercise.

anonym-HPI:
This could be generalized, yes, but could get more complicated as most people would use it, clicking on a vehicle to get information, could still be a valid way (like participant walking to the vehicle and looking - not giving too much information). As you tagged it, needs discussion and so on.

lukasrad02:
Solved by having loading indicators for vehicles (#1147)

PR #501, opened 2022-07-29
vehicle image changing depending on number of patients loaded

anonym-HPI:
fix https://github.com/hpi-sam/digital-fuesim-manv/issues/500

### DIFF
--- a/shared/src/models/vehicle-template.ts
+++ b/shared/src/models/vehicle-template.ts
@@ -5,7 +5,9 @@ import {
     ValidateNested,
     IsNumber,
     IsArray,
+    IsOptional,
 } from 'class-validator';
+import { ReducerError } from '../store';
 import { uuidValidationOptions, UUID, uuid } from '../utils';
 import type { PersonnelType } from './utils';
 import { CanCaterFor, ImageProperties, getCreate } from './utils';
@@ -20,9 +22,23 @@ export class VehicleTemplate {
     @IsString()
     public readonly name: string;
 
+    /**
+     * If an array, at position 0 it means zero patients are loaded
+     * at position 1 one patient is loaded, etc.
+     */
     @ValidateNested()
     @Type(() => ImageProperties)
     public readonly image: ImageProperties;
+
+    /**
+     * If an array, at position 0 it means zero patients are loaded
+     * at position 1 one patient is loaded, etc.
+     */
+    @IsArray()
+    @ValidateNested()
+    @Type(() => ImageProperties)
+    @IsOptional()
+    public readonly imagesPatientsLoaded?: ImageProperties[];
 
     @IsNumber()
     public readonly patientCapacity: number;
@@ -45,7 +61,8 @@ export class VehicleTemplate {
         image: ImageProperties,
         patientCapacity: number,
         personnel: readonly PersonnelType[],
-        materials: readonly CanCaterFor[]
+        materials: readonly CanCaterFor[],
+        imagesPatientsLoaded?: ImageProperties[]
     ) {
         this.vehicleType = vehicleType;
         this.name = name;
@@ -53,6 +70,15 @@ export class VehicleTemplate {
         this.patientCapacity = patientCapacity;
         this.personnel = personnel;
         this.materials = materials;
+
+        if (
+            imagesPatientsLoaded !== undefined &&
+            imagesPatientsLoaded.length + 1 !== patientCapacity
+        )
+            throw new ReducerError(
+                `vehicle was tried to be created, but imagesPatientsLoaded.length is not equal to patientCapacity`
+            );
+        this.imagesPatientsLoaded = imagesPatientsLoaded;
     }
 
     static readonly create = getCreate(this);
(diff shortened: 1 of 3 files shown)


shared/src/models/vehicle-template.ts
```diff
@@ -20,10 +22,24 @@ export class VehicleTemplate {
     @IsString()
     public readonly name: string;
 
+    /**
+     * If an array, at position 0 it means zero patients are loaded
+     * at position 1 one patient is loaded, etc.
+     */
     @ValidateNested()
     @Type(() => ImageProperties)
     public readonly image: ImageProperties;
```
ClFeSc, line 31: This can, in contrast to the comment, never be an array.

shared/src/models/vehicle-template.ts
```diff
@@ -20,10 +22,24 @@ export class VehicleTemplate {
...
 
+    /**
+     * If an array, at position 0 it means zero patients are loaded
+     * at position 1 one patient is loaded, etc.
+     */
+    @IsArray()
+    @ValidateNested()
+    @Type(() => ImageProperties)
+    @IsOptional()
+    public readonly imagesPatientsLoaded?: ImageProperties[];
```
ClFeSc, line 41: Prefer `null` over `undefined` for new code (@Dassderdie I think we've come to that conclusion?):

```suggestion
    public readonly imagesPatientsLoaded: ImageProperties[] | null;
```

The validators should work the same, I've not tested it, though.

shared/src/models/vehicle-template.ts
```diff
@@ -45,14 +61,24 @@ export class VehicleTemplate {
...
         this.image = image;
         this.patientCapacity = patientCapacity;
         this.personnel = personnel;
         this.materials = materials;
+
+        if (
+            imagesPatientsLoaded !== undefined &&
+            imagesPatientsLoaded.length + 1 !== patientCapacity
+        )
+            throw new ReducerError(
```
ClFeSc, line 78: I'm not sure a `ReducerError` is the best choice here. @Dassderdie What do you think?

shared/src/models/vehicle-template.ts
```diff
@@ -45,14 +61,24 @@ export class VehicleTemplate {
...
+                `vehicle was tried to be created, but imagesPatientsLoaded.length is not equal to patientCapacity`
```
ClFeSc, line 79:
```suggestion
                `Cannot create a vehicle with ${imagesPatientsLoaded.length} images and a capacity of ${patientCapacity}.`
```

shared/src/models/vehicle.ts
```diff
@@ -64,13 +76,22 @@ export class Vehicle {
...
         this.materialIds = materialIds;
         this.patientCapacity = patientCapacity;
         this.image = image;
+        if (
+            imagesPatientsLoaded !== undefined &&
+            imagesPatientsLoaded.length + 1 !== patientCapacity
+        )
+            throw new ReducerError(
+                `vehicle was tried to be created, but imagesPatientsLoaded.length is not equal to patientCapacity`
+            );
```
ClFeSc, line 93: Same comments as in the template.

shared/src/store/action-reducers/vehicle.ts
```diff
@@ -308,6 +310,25 @@ export namespace VehicleActionReducers {
...
+                    undefined
+                ) {
+                    // TODO: why needing "!"" at the end?
+                    vehicle.image =
+                        vehicle.imagesPatientsLoaded[
+                            numberOfPatientsInVehicle
+                        ]!;
+                } else {
+                    throw new ReducerError(
+                        `vehicle was tried to be unloaed while not having correct imagesPatientsLoaded`
```
ClFeSc, line 328:
```suggestion
                        `vehicle was tried to be loaded while not having correct imagesPatientsLoaded`
```

shared/src/store/action-reducers/vehicle.ts
```diff
@@ -308,6 +310,25 @@ export namespace VehicleActionReducers {
...
             }
+            if (vehicle.imagesPatientsLoaded !== undefined) {
+                const numberOfPatientsInVehicle = Object.keys(
+                    vehicle.patientIds
+                ).length;
+                if (
+                    vehicle.imagesPatientsLoaded[numberOfPatientsInVehicle] !==
+                    undefined
+                ) {
+                    // TODO: why needing "!"" at the end?
```
ClFeSc, line 321: Because TS has no idea that this element exists. It can't relate the check above to here. You can use an extra `const` for this.

ClFeSc (changes requested):
@anonym-HPI Does this work for you locally? I can't start the backend with `ReferenceError: Cannot access 'Vehicle' before initialization`, but this may be caused by my local setup.

Edit: The CI would like to agree with me that this does not work.

Dassderdie:
This won't be merged anytime soon, I guess. I, therefore, close this PR but mention it in the corresponding issue #500 .

PR #502, opened 2022-07-29
Alternative: Fix/500 vehicles different images how many patients are loaded in

anonym-HPI:
could also fix https://github.com/hpi-sam/digital-fuesim-manv/issues/500


ClFeSc (changes requested):
Please fix the CI.

PR #504, opened 2022-07-29
Add notes to `backend/README.md` about #183

### LINKED ISSUE #183: Find mechanism for cleanup in long-running server setups
It is possible to delete any exercise with trainer id id using DELETE /api/exercise/id, and there is #179 to find a way to invoke this with the frontend.

However, there sould also be an automatic deletion/achtiving mechanism, unless you can convince me that database / persistance https://github.com/hpi-sam/BP2021HG1/commit/f8dfcbc7ec26e744d541414f861142a16785b984 will solve this implicitly.

This is not so high priority, but it should be done before a must-have version is delivered to our customers (and run long-term on servers).

ClFeSc:
Closes #183 (for now)

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -106,8 +106,8 @@ Look at [`migrations.ts`](./src/database/state-migrations/migrations.ts) for mor
 
 ### Note on long term storage
 
-The current setup when using a database is that no exercises get deleted unless anyone deletes them from the UI (or, to be more precisely, using the HTTP request `DELETE /api/exercises/:exerciseId`).
-An average exercise with four sections being actively played and about 45 minutes of exercise time seems to take up about 10 MB storage in the database.
-This is in itself not much but in case many exercises of this size are run each day, it can scale quickly.
-Also note that there can never be more than 10,000 exercises because then the id generator fails.
-Finally note that it may be an issue of data protection laws like the [GDPR](https://gdpr.eu/) when user data never gets deleted, even after years of not using it.
+The current setup when using a database is that no exercises get deleted unless anyone deletes them from the UI (or, more precisely, using the HTTP request `DELETE /api/exercises/:exerciseId`).
+An average exercise with four sections being actively played and about 45 minutes of exercise time seems to take up about 10 MB of storage in the database.
+This is not much in itself, but if many exercises of this size are run each day, it can scale quickly.
+Also, note that there can never be more than 10,000 exercises because then the id generator fails.
+Finally, note that it may be an issue of data protection laws like the [GDPR](https://gdpr.eu/) when user data never gets deleted, even after years of not using it.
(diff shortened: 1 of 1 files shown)


backend/README.md
```diff
@@ -103,3 +103,11 @@ Note however that this results in a) all history being saved in memory instead o
 
 We use [state migrations](./src/database/state-migrations/) to convert outdated states to new versions.
 Look at [`migrations.ts`](./src/database/state-migrations/migrations.ts) for more information.
+
+### Note on long term storage
+
+The current setup when using a database is that no exercises get deleted unless anyone deletes them from the UI (or, to be more precisely, using the HTTP request `DELETE /api/exercises/:exerciseId`).
+An average exercise with four segments and about 45 minutes of exercise time seems to take up about 10 MB storage in the database.
```
christianzoellner, line 110:
```suggestion
An average exercise with four viewports being actively played and about 45 minutes of exercise time seems to take up about 10 MB storage in the database.
```

instead of viewport you could also use "section", which is the domain term rather than the technical term.

christianzoellner (approved):
looks good to me, except for one minor note on terminology.

backend/README.md
```diff
@@ -103,3 +103,11 @@ Note however that this results in a) all history being saved in memory instead o
...
 We use [state migrations](./src/database/state-migrations/) to convert outdated states to new versions.
 Look at [`migrations.ts`](./src/database/state-migrations/migrations.ts) for more information.
+
+### Note on long term storage
+
+The current setup when using a database is that no exercises get deleted unless anyone deletes them from the UI (or, to be more precisely, using the HTTP request `DELETE /api/exercises/:exerciseId`).
+An average exercise with four sections being actively played and about 45 minutes of exercise time seems to take up about 10 MB storage in the database.
+This is in itself not much but in case many exercises of this size are run each day, it can scale quickly.
+Also note that there can never be more than 10,000 exercises because then the id generator fails.
+Finally note that it may be an issue of data protection laws like the [GDPR](https://gdpr.eu/) when user data never gets deleted, even after years of not using it.
```
Dassderdie, line 113:
```suggestion
The current setup when using a database is that no exercises get deleted unless anyone deletes them from the UI (or, more precisely, using the HTTP request `DELETE /api/exercises/:exerciseId`).
An average exercise with four sections being actively played and about 45 minutes of exercise time seems to take up about 10 MB of storage in the database.
This is not much in itself, but if many exercises of this size are run each day, it can scale quickly.
Also, note that there can never be more than 10,000 exercises because then the id generator fails.
Finally, note that it may be an issue of data protection laws like the [GDPR](https://gdpr.eu/) when user data never gets deleted, even after years of not using it.
```

Dassderdie: -> from Grammarly

PR #507, opened 2022-07-29
Only validate actions once

### LINKED ISSUE #478: Actions are two times validated in the backend
Each action that is proposed to the backend is validated two times.

The validation in the `ExcerciseWrapper` should probably not be done in the function the action is reduced.

ClFeSc:
Closes #478

### DIFF
--- a/backend/src/exercise/exercise-wrapper.spec.ts
+++ b/backend/src/exercise/exercise-wrapper.spec.ts
@@ -105,7 +105,8 @@ describe('Exercise Wrapper', () => {
 
             exercise.applyAction(
                 { type: '[Exercise] Start' },
-                (exercise as any).emitterUUID
+                (exercise as any).emitterUUID,
+                true
             );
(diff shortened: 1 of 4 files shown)


Dassderdie (commented):
@ClFeSc This isn't exactly what I had in mind.

The action was and still is validated internally in the ExerciseWrapper in the `reduce` function.

Instead, I propose the following:
* There are sections of code we trust and some we don't.
* We *don't* trust an action that comes from outside -> validate it before in the WebSocket handler, the import endpoint - and maybe the database?
* We *do* trust the actions that are directly proposed in the `ExerciseWrapper`. This means they don't have to be checked in there. Currently, we are doing defensive programming for the `Tick` as well as multiple `Start` and `Pause` and `removeClient` actions. The behavior, if these checks should fail, is also unspecified. THis should rather be done via tests instead of runtime checks.

Dassderdie (approved):
Good job.

PR #508, opened 2022-07-29
Remove `Action.timestamp`

### LINKED ISSUE #498: Remove nondeterministic behavior from the actions
Currently, `Action`s support nondeterministic behavior via the `timestamp: number` field, which gets replaced by `Date.now()` on the server. This should be removed.

ClFeSc:
Closes #498


Dassderdie (approved):
I haven't tested the migrations, but else it works fine and the code looks good.

ClFeSc:
> I haven't tested the migrations, but else it works fine and the code looks good.

There are no new migrations as there were no actions that still used `timestamp`. The change in the migration was only for TS type checking.

PR #513, opened 2022-07-30
Add LF endings for `docker/docker-entrypoint.sh` to `.gitattributes`

### LINKED ISSUE #479: The docker-entrypoint.sh line endings must be LF
Probably https://stackoverflow.com/a/10357545/12698757

ClFeSc:
Closes #479 

Please verify that this works on Windows. On Linux, this is the default behavior, so I can't check it.


Dassderdie (approved):
Tested and works on Windows.

PR #514, opened 2022-07-30
Sort lists

### LINKED ISSUE #432: Transfer points should be alphabetically ordered
I noticed the chaotic representation of the listed transfer points in the following views:
* when adding a transfer target to a transfer point
* when selecting a traget for an alert group

Example: The three highlighted potential transfer targerts should be displayed next to each other and thus be easier to find as a group...
grafik

ClFeSc:
Closes #432 

Currently sorted:

* transfer points to be added in transfer point popup
* hospitals to be added in transfer point popup
* transfer points as target in eoc

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/send-alarm-group-interface/send-alarm-group-interface.component.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/send-alarm-group-interface/send-alarm-group-interface.component.ts
@@ -1,12 +1,13 @@
 import type { OnDestroy } from '@angular/core';
 import { Component } from '@angular/core';
 import { Store } from '@ngrx/store';
-import type { AlarmGroup, UUID } from 'digital-fuesim-manv-shared';
+import type { UUID, AlarmGroup } from 'digital-fuesim-manv-shared';
 import {
-    AlarmGroupStartPoint,
+    sortSet,
     createVehicleParameters,
+    AlarmGroupStartPoint,
 } from 'digital-fuesim-manv-shared';
-import { Subject, takeUntil } from 'rxjs';
+import { map, Subject, takeUntil } from 'rxjs';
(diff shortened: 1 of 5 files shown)


ClFeSc:
@Dassderdie Is it ok to do it this way?

Dassderdie:
> @Dassderdie Is it ok to do it this way?

I'm unsure about the general approach. I think I would rather create an `orderBy`-pipe with the following specification:

```ts
// TODO: also accept null/undefined to be easier used with async pipe
public transform<Item>(array: Immutable<Item[]>, getOrderByValue: (item: Immutable<Item>) => string | number, order: 'asc' | 'desc' | null): Immutable<Item[]>
```


To work in a pure pipe, there are some additional constraints on the parameters, but they should in our case, always be fulfilled, because our state is immutable.

The `getOrderByValue` function would have to be defined in the component. Probably like this:

```ts
public getOrderByValue: (transferPoint: TransferPoint) => string | number = (transferPoint) => TransferPoint.getFullName(transferPoint);
```

The main idea is that this pipe could be reused to sort in tables. This could be done like this:

```ts
public readonly getValueFunctions = {
    name: (transferPoint) => TransferPoint.getFullName(transferPoint),
   duration: (transferPoint) => transferPoint.duration
} as const;
public orderByKey: keyof typeof this.getValueFunctions = 'name';
```

```html
*ngFor="transferPoint of transferPoints$ | async | values | orderBy: getValueFunctions[orderByKey]: order; 
trackBy: ...
"
```


```html
(click)="orderByKey = 'name'"
```

Additional logic (maybe encapsulated in a directive) could do the ascending and descending part.

The gamble is that the boilerplate for adding a `switchMap` and additional `BehaviorSubjects` to implement table sorting purely via rxjs would be more than doing it with the pipe. A disadvantage is that more logic is in the template.

ClFeSc:
@Dassderdie I've implemented this in #516, and it seems to be way easier.

ClFeSc:
Superseded by #516.

PR #515, opened 2022-07-30
Improve shared documentation

ClFeSc:


shared/README.md
```diff
@@ -6,14 +6,35 @@ Keep in mind to add new exports to the `index.ts` file in the folder.
...
+
+When writing new `Action`s, note the comments in [src/store/action-reducer.ts](./src/store/action-reducer.ts) and [src/store/action-reducers/action-reducers.ts](./src/store/action-reducers/action-reducers.ts).
+You can orient yourself at the already existing `Action`s in [src/store/action-reducers](./src/store/action-reducers/).
+Note especially that new `Action`s have to be registered in `actionReducers` in [src/store/action-reducers/action-reducers.ts](./src/store/action-reducers/action-reducers.ts).
+
+## Adding new models
+
+When adding new models note that they must expose a `static` `create` method that calls the constructor, removes the prototype and freezes the object.
+This can be achieved by using `static create = getCreate(this)`, like in existing models.
+Also mark the `constructor` `deprecated` via a JSDoc comment.
```
Dassderdie, line 32:
```suggestion
Also, mark the `constructor` as `deprecated` via a JSDoc comment.
```

shared/src/store/action-reducer.ts
```diff
@@ -31,6 +31,7 @@ export interface Action {
 
 /**
  * A pure function that applies the action on the provided state.
+ * Pure especially means that no non-deterministic behavior (like `Date.now()` or `uuid()`) may be used.
```
Dassderdie, line 34:
```suggestion
 * A [pure function](https://en.wikipedia.org/wiki/Pure_function) is a function that produces the same return value for identical arguments (doesn't use non-deterministic functions like `Date.now()`, `uuid()` or `Math.random()`) and is side-effect free.
```

Dassderdie (approved):
See my two comments. Also, note that I'll add more documentation this week (the main focus will be on the whole synchronization stuff).

PR #516, opened 2022-07-31
Add `orderBy` pipe

### LINKED ISSUE #432: Transfer points should be alphabetically ordered

ClFeSc:
Closes #432 

Currently sorted:

* transfer points to be added in transfer point popup
* hospitals to be added in transfer point popup
* transfer points as target in eoc

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/transfer-point-popup/transfer-point-popup.component.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/transfer-point-popup/transfer-point-popup.component.ts
@@ -1,7 +1,8 @@
 import type { OnInit } from '@angular/core';
 import { Component, EventEmitter, Output } from '@angular/core';
 import { Store } from '@ngrx/store';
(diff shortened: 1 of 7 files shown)


frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/transfer-point-popup/transfer-point-popup.component.ts
```diff
@@ -44,6 +45,14 @@ export class TransferPointPopupComponent implements PopupComponent, OnInit {
 
     public hospitals$ = this.store.select(selectHospitals);
 
+    public getOrderByValueTransferPoints: (
```
Dassderdie, line 48: Either 
```suggestion
    public getOrderByValueTransferPoint: (
```

```suggestion
    public getTransferPointOrderByValue: (
```

- the names are still not pretty, but I'm unable to come up with better ones.

Dassderdie (approved):
Only small stuff

PR #517, opened 2022-07-31
Add test for import

ClFeSc:
@Dassderdie FYI

Not tested on CI (timeouts etc.)


ClFeSc:
@Dassderdie Could you take a look at ESLint? Maybe you're not as stupid as I am. I can't get ESLint to ignore the new files which are too large for the CI.

backend/test/http-exercise.spec.ts
```diff
@@ -28,6 +31,51 @@ describe('exercise', () => {
...
+                    exampleExerciseFinalState as Mutable<ExerciseState>;
+                receivedState.participantId = '123456';
+                expectedState.participantId = '123456';
+                expect(Object.keys(receivedState.clients).length).toBe(
+                    Object.keys(expectedState.clients).length
+                );
+                receivedState.clients = {};
+                expectedState.clients = {};
+
+                expect(receivedState).toStrictEqual(expectedState);
```
benn02, line 76: Does this still work after a migration? Or do we have to manually update the final state then?

ClFeSc:
This has been superseded by #774, even though there the test for a correct import has been dropped. This would, however, be a very tedious task to maintain and has never posed a problem, so this should be fine, at least for now.

PR #495, opened 2022-07-27
Fix actions not being saved to databse on import from file

ClFeSc:
There was a person who wrote code that was never able to work.

PR #519, opened 2022-07-31
Fix database migrations

ClFeSc:
Previously, they overrode actions from all exercises with the same id due to a missing where clause


Dassderdie (approved):
Code lgtm.

If you have tested it, I'm ok with it.

ClFeSc:
> Code lgtm.
> 
> If you have tested it, I'm ok with it.

Yes, I've downloaded the production database and verified that it works locally. One exercise had to be deleted as it's broken due to #495, but the other ones loaded.

Issue #464, opened 2022-07-18
Increase import/upload size and editable via ENV


anonym-HPI:
Right now, the upload size is limited to 1mb
via http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size (when using docker)
and to 200mb in backend\src\exercise\http-server.ts (without docker, as no nginx)

at least setting it to 200MB in nginx and better make it editable via ENV

checking: the different writing, in nginx it is `200M` and in http-server.ts `200mb` - checking if both could use `MB` or so

ClFeSc:
@anonym-HPI Any progress on this? It would be quite important for a usable experience.

PR #520, opened 2022-08-19
upload limit via ENV

anonym-HPI:
fixes https://github.com/hpi-sam/digital-fuesim-manv/issues/464

### DIFF
--- a/backend/src/exercise/http-server.ts
+++ b/backend/src/exercise/http-server.ts
@@ -2,6 +2,7 @@ import type { Server as HttpServer } from 'node:http';
 import cors from 'cors';
 import type { Express } from 'express';
 import express from 'express';
+import { Config } from 'config';
 import type { DatabaseService } from '../database/services/database-service';
 import {
     deleteExercise,
@@ -17,16 +18,11 @@ export class ExerciseHttpServer {
     /**
      * @param uploadLimit in Megabyte can be set via ENV DFM_UPLOAD_LIMIT
      */
-    constructor(
-        app: Express,
-        port: number,
-        uploadLimit: number,
-        databaseService: DatabaseService
-    ) {
+    constructor(app: Express, databaseService: DatabaseService) {
         // TODO: Temporary allow all
         app.use(cors());
 
-        app.use(express.json({ limit: String(`${uploadLimit}mb`) }));
+        app.use(express.json({ limit: `${Config.uploadLimit}mb` }));
 
         // This endpoint is used to determine whether the API itself is running.
         // It should be independent from any other services that may or may not be running.
@@ -48,7 +44,7 @@ export class ExerciseHttpServer {
             )
         );
 
-        this.httpServer = app.listen(port);
+        this.httpServer = app.listen(Config.httpPort);
     }
 
     public close(): void {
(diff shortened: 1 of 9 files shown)


backend/src/exercise/http-server.ts
```diff
@@ -14,11 +14,19 @@ import { secureHttp } from './http-handler/secure-http';
...
+        app: Express,
+        port: number,
+        uploadLimit: number,
+        databaseService: DatabaseService
+    ) {
         // TODO: Temporary allow all
         app.use(cors());
 
-        app.use(express.json({ limit: '200mb' }));
+        app.use(express.json({ limit: String(`${uploadLimit}mb`) }));
```
ClFeSc, line 29: Two things here:

1. Are you sure you need to call `String()` here? Wouldn't it be enough to just use the format string?
2. I think it would be better to directly access the config here instead of passing the limit as a parameter. I'm not sure though.

ClFeSc (commented):
Works like a charm, just two small comments.

backend/src/exercise/http-server.ts
anonym-HPI: String( ) is probably not needed.

For the orher, then also the port should be reqd directly from config or not?

ClFeSc: I agree with 2, so only remove the `String()` call and this is good to go.

anonym-HPI: @ClFeSc removing the port (httpPort with or without webSocketPort) seems to break cypress, I am not sure why, it should work, maybe something about initialization missing in cypress test?

ClFeSc: I don't know what you mean. You can't just remove parts of the program and expect it to work like before.

backend/src/fuesim-server.ts
```diff
@@ -71,18 +71,10 @@ export class FuesimServer {
         this.saveTickInterval
     );
 
-    constructor(
-        private readonly databaseService: DatabaseService,
-        websocketPort: number = Config.websocketPort,
-        httpPort: number = Config.httpPort
```
anonym-HPI, line 77: @ClFeSc I removed these parameters and called them inside the websocket and http server directly from the config, I haven't seen that the constructor ever gets called with `websocketPort` and `httpPort` set (therefore always being the default, maybe cypress does it.

anonym-HPI: Maybe the initialization is not called correctly and therefore the ports aren't set to the values in test environment?

ClFeSc: A look at the logs would tell you that you're incorrectly importing `Config`, you have to use relative imports.

anonym-HPI: @ClFeSc have given not the reletavie path in the imports, auto merge is on.

anonym-HPI: @ClFeSc have given the relative path in the imports, auto merge is on.

ClFeSc (approved):
I'm not sure whether I 100% agree with the changes around the ports, but I think it's good for now.

anonym-HPI:
> I'm not sure whether I 100% agree with the changes around the ports, but I think it's good for now.

What would you have liked differently?

ClFeSc:
> > I'm not sure whether I 100% agree with the changes around the ports, but I think it's good for now.
> 
> What would you have liked differently?

I'm not sure whether it would be better to gather the `Config` at once and pass it along, as now you can't deviate from the ports and specifications defined by the config. However, this is the only use case we currently have, so I'm fine with it but I noticed that it might be problematic in the future.

anonym-HPI:
> 

Ok, if/when this is needed, this can be changed.

Issue #392, opened 2022-05-22
Export & Import (full & partial)


ClFeSc:
Trainers should be able to export either

* [x] the current state...
  * [x] without the action history (= snapshot)
  * [x] with action history and initial state,

or any combination of

* [ ] the existing viewports,
* [ ] the patient templates,
* [ ] the vehicle templates,
* [ ] the mapImage templates, and
* [ ] the tileMapProperties.

When importing such an export
* [x] a new exercise should be created when the state is included, or
* [ ] for the other points offer overriding or appending existing data.

There must be something to handle id clashes.

ClFeSc:
Note https://github.com/hpi-sam/digital-fuesim-manv/pull/407/files#r885098984

christianzoellner:
Added [x] to the issue description to allow tracking of progress on this issue.

PR #522, opened 2022-09-16
Allow partial export

ClFeSc:
This addresses the partial exports as described in #392.

* [x] Allow exporting a partial export
* [x] Allow importing a partial export
* [x] Find a way to migrate partial exports without too much hassle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 - this software has now a license (with some exceptions), see LICENSE-README.md
     - the software includes a file with third party acknoledgments, a license.html and the docker container will include a git archive of the source code downloadable in zip format
     - new npm run commands for licensing (third party acknowledgements) and creating source code archive
+- Add possibility to export only templates from an excercise and import them in another one.
+    - Patients, vehicles, and images templates can be selected.
+    - Imports can either overwrite existing templates or be added to the excercise.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,9 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 - this software has now a license (with some exceptions), see LICENSE-README.md
     - the software includes a file with third party acknoledgments, a license.html and the docker container will include a git archive of the source code downloadable in zip format
     - new npm run commands for licensing (third party acknowledgements) and creating source code archive
-- Add possibility to export only templates from an excercise and import them in another one.
-    - Patients, vehicles, and images templates can be selected.
-    - Imports can either overwrite existing templates or be added to the excercise.
+- Add possibility to export only templates from an exercise and import them in another one.
+    - Patients, vehicles, and image templates can be selected.
+    - Imports can either overwrite existing templates or be added to the exercise.
 
 ### Changed
 
--- a/frontend/src/app/pages/exercises/exercise/shared/partial-import/partial-import-modal/partial-import-modal.component.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/partial-import/partial-import-modal/partial-import-modal.component.ts
@@ -14,7 +14,7 @@ import { MessageService } from 'src/app/core/messages/message.service';
 export class PartialImportModalComponent {
     public importingPartialExport = false;
     constructor(
-        public activeModal: NgbActiveModal,
+        private readonly activeModal: NgbActiveModal,
         private readonly messageService: MessageService,
         private readonly exerciseService: ExerciseService
     ) {}
--- a/frontend/src/app/pages/exercises/exercise/shared/partial-import/partial-import-modal/partial-import-modal.component.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/partial-import/partial-import-modal/partial-import-modal.component.ts
@@ -1,0 +1,55 @@
+import { Component } from '@angular/core';
+import { NgbActiveModal } from '@ng-bootstrap/ng-bootstrap';
+import type { PartialExport } from 'digital-fuesim-manv-shared';
+import { preparePartialExportForImport } from 'digital-fuesim-manv-shared';
+import { ExerciseService } from 'src/app/core/exercise.service';
+import { MessageService } from 'src/app/core/messages/message.service';
+
+@Component({
+    selector: 'app-partial-import-modal',
+    templateUrl: './partial-import-modal.component.html',
+    styleUrls: ['./partial-import-modal.component.scss'],
+    standalone: false,
+})
+export class PartialImportModalComponent {
+    public importingPartialExport = false;
+    constructor(
+        public activeModal: NgbActiveModal,
+        private readonly messageService: MessageService,
+        private readonly exerciseService: ExerciseService
+    ) {}
+
+    public partialExport!: PartialExport;
+
+    public close() {
+        this.activeModal.close();
+    }
+
+    public async partialImportOverwrite(mode: 'append' | 'overwrite') {
+        this.importingPartialExport = true;
+        try {
+            const result = await this.exerciseService.proposeAction({
+                type: '[Exercise] Import Templates',
+                mode,
+                partialExport: preparePartialExportForImport(
+                    this.partialExport
+                ),
+            });
+            if (!result.success) {
+                throw new Error((result as { message?: string }).message);
+            }
+            this.messageService.postMessage({
+                title: 'Vorlagen erfolgreich importiert',
+                color: 'success',
+            });
+            this.close();
+        } catch (error: unknown) {
+            this.messageService.postError({
+                title: 'Fehler beim Importieren von Vorlagen',
+                error,
+            });
+        } finally {
+            this.importingPartialExport = false;
+        }
+    }
+}
--- a/frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.ts
@@ -1,22 +1,32 @@
 import { Component } from '@angular/core';
 import { NgbModal } from '@ng-bootstrap/ng-bootstrap';
 import { Store } from '@ngrx/store';
+import { plainToInstance } from 'class-transformer';
+import {
+    Viewport,
+    TransferPoint,
+    PartialExport,
+    migratePartialExport,
+    validateExerciseExport,
+} from 'digital-fuesim-manv-shared';
+import { ExerciseService } from 'src/app/core/exercise.service';
+import { MessageService } from 'src/app/core/messages/message.service';
 import type {
     ColorCode,
     UUID,
     VehicleTemplate,
 } from 'digital-fuesim-manv-shared';
-import { TransferPoint, Viewport } from 'digital-fuesim-manv-shared';
 import type { AppState } from 'src/app/state/app.state';
 import {
-    selectMapImagesTemplates,
-    selectPatientCategories,
     selectVehicleTemplates,
+    selectPatientCategories,
+    selectMapImagesTemplates,
 } from 'src/app/state/application/selectors/exercise.selectors';
 import { DragElementService } from '../core/drag-element.service';
 import { TransferLinesService } from '../core/transfer-lines.service';
 import { openCreateImageTemplateModal } from '../editor-panel/create-image-template-modal/open-create-image-template-modal';
 import { openEditImageTemplateModal } from '../editor-panel/edit-image-template-modal/open-edit-image-template-modal';
+import { openPartialImportOverwriteModal } from '../partial-import/open-partial-import-overwrite-modal';
 import { simulatedRegionDragTemplates } from '../editor-panel/templates/simulated-region';
 
 @Component({
@@ -52,7 +62,9 @@ export class TrainerMapEditorComponent {
         private readonly store: Store<AppState>,
         public readonly dragElementService: DragElementService,
         public readonly transferLinesService: TransferLinesService,
-        private readonly ngbModalService: NgbModal
+        private readonly ngbModalService: NgbModal,
+        private readonly messageService: MessageService,
+        private readonly exerciseService: ExerciseService
     ) {}
 
     public readonly simulatedRegionTemplates = simulatedRegionDragTemplates;
@@ -84,5 +96,36 @@ export class TrainerMapEditorComponent {
             type: 'vehicle',
             template: vehicleTemplate,
         });
+    }
+
+    public importingTemplates = false;
+    public async importPartialExport(fileList: FileList) {
+        try {
+            this.importingTemplates = true;
+            const importedText = await fileList.item(0)?.text();
+            if (importedText === undefined) {
+                // The file dialog has been aborted.
+                return;
+            }
+            const importedPlainObject = JSON.parse(
+                importedText
+            ) as PartialExport;
+            const migratedPartialExport =
+                migratePartialExport(importedPlainObject);
+            const validation = validateExerciseExport(
+                plainToInstance(PartialExport, migratedPartialExport)
+            );
+            if (validation.length > 0) {
+                throw Error(
+                    `PartialExport is invalid:\n${validation.join('\n')}`
+                );
+            }
+            openPartialImportOverwriteModal(
+                this.ngbModalService,
+                migratedPartialExport
+            );
+        } finally {
+            this.importingTemplates = false;
+        }
     }
 }
--- a/frontend/src/app/pages/landing-page/landing-page/landing-page.component.ts
+++ b/frontend/src/app/pages/landing-page/landing-page/landing-page.component.ts
@@ -72,7 +72,6 @@ export class LandingPageComponent {
                 (type === 'complete'
                     ? StateExport
                     : PartialExport) as Constructor<
-                    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-arguments
                     PartialExport | StateExport
                 >,
                 importPlain
@@ -98,7 +97,7 @@ export class LandingPageComponent {
                 }
                 case 'partial': {
                     throw new Error(
-                        'Dieser Typ kann zur Zeit nicht importiert werden.'
+                        'Dieser Typ kann nur innerhalb einer Übung importiert werden.'
                     );
                 }
             }
--- a/frontend/src/app/pages/landing-page/landing-page/landing-page.component.ts
+++ b/frontend/src/app/pages/landing-page/landing-page/landing-page.component.ts
@@ -96,9 +96,12 @@ export class LandingPageComponent {
                     break;
                 }
                 case 'partial': {
-                    throw new Error(
-                        'Dieser Typ kann nur innerhalb einer Übung importiert werden.'
-                    );
+                    this.messageService.postMessage({
+                        color: 'danger',
+                        title: 'Unerlaubter Importtyp',
+                        body: 'Dieser Typ kann nur innerhalb einer Übung importiert werden.',
+                    });
+                    break;
                 }
             }
         } catch (error: unknown) {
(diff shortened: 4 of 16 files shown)


ClFeSc:
>     * [ ]  Find a way to migrate partial exports without too much hassle

This is currently depending on #581 as I need to run a migration in the reducer (or the frontend) but not the backend.

ClFeSc:
@Dassderdie Apart from the missing call of the migration this should now be ready for review.

Dassderdie:
1. Skimming through the code, it seemed like having a `complete` and `partial` `StateExport` results in more code than just having a `StateExport` and a `TemplateExport`, which are completely separate from each other.

2. In general, only allowing up-to-date imports in the backend and doing the migration in the frontend, could be advantageous. 
    * Less strain on the server (DOS)
    * Old migrations can't be used for attacks.
    * The user can be provided with better feedback during the migration (e.g. "Running migration 1 of 11")

ClFeSc:
> 1. Skimming through the code, it seemed like having a `complete` and `partial` `StateExport` results in more code than just having a `StateExport` and a `TemplateExport`, which are completely separate from each other.
> 
>     2. In general, only allowing up-to-date imports in the backend and doing the migration in the frontend, could be advantageous.
>        
>        * Less strain on the server (DOS)
>        * Old migrations can't be used for attacks.
>        * The user can be provided with better feedback during the migration (e.g. "Running migration 1 of 11")

I agree with both points made. However, I'd stick with the current solution for now. The second point depends on #581, and the first point would be a refactoring I'd like to do separately.  Maybe we could even disallow migrations for partial imports for now, depending on your plan with #581.

shared/src/store/action-reducers/exercise.ts
```diff
@@ -151,3 +215,26 @@ function refreshTransfer(
...
+    // `patientCategories` don't have an `id`
+    const properties = ['mapImageTemplates', 'vehicleTemplates'] as const;
+    for (const property of properties) {
+        const value = copy[property];
+        if (value !== undefined) {
+            for (const entry of value) {
+                entry.id = uuid();
+            }
+        }
+    }
```
Dassderdie, line 238: This would be easier to understand, if instead of `property`, `value` and `entry` this would be `templateType`, `templates`, and `template`.

YAGNI

Dassderdie (changes requested):

frontend/src/app/pages/exercises/exercise/shared/partial-import-overwrite/partial-import-overwrite-modal/partial-import-overwrite.component.html
```diff
@@ -0,0 +1,38 @@
...
+    <div>
+        Wollen Sie die vorhandenen Vorlagen mit den importierten Vorlagen
+        ersetzen oder ergänzen?
+    </div>
+</div>
+<div class="modal-footer">
+    <label *ngIf="importingPartialExport" class="btn-outline-primary">
+        <span class="spinner-border spinner-border-sm"></span>
+        Import läuft
+    </label>
```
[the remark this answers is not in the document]
Dassderdie: I made a complete code proposal for a different UI below. The changes I propose here are already incorporated into it.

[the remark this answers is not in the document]
Dassderdie: See the proposal below.

shared/src/store/action-reducers/exercise.ts
```diff
@@ -128,6 +143,55 @@ export namespace ExerciseActionReducers {
         },
         rights: 'server',
     };
+
+    export const templateImport: ActionReducer<ImportTemplatesAction> = {
+        action: ImportTemplatesAction,
+        reducer: (draftState, { mode, partialExport }) => {
+            // TODO: Use `migratePartialExport` function that is currently defined in the backend.
```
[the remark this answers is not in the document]
ClFeSc: Yes, I'll move it to the frontend.

shared/src/store/action-reducers/exercise.ts
```diff
@@ -128,6 +143,55 @@ export namespace ExerciseActionReducers {
...
+                }
+            }
+            if (mutablePartialExport.vehicleTemplates !== undefined) {
+                if (mode === 'append') {
+                    draftState.vehicleTemplates.push(
+                        ...mutablePartialExport.vehicleTemplates
+                    );
+                } else {
+                    // Remove all alarm groups as all existing vehicle templates are being removed
+                    draftState.alarmGroups = {};
```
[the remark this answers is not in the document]
ClFeSc: Good idea, I'll update this.

frontend/src/app/pages/exercises/exercise/shared/partial-import-overwrite/partial-import-overwrite-modal/partial-import-overwrite.component.html
```diff
@@ -0,0 +1,38 @@
...
+        type="button"
+        class="btn btn-warning"
+        (click)="partialImportOverwrite('append')"
+    >
+        Ergänzen
+    </button>
+    <button
+        type="button"
+        class="btn btn-outline-secondary"
+        autofocus
```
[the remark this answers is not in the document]
ClFeSc: Btw. you're using `autofocus` yourself in `frontend/src/app/core/confirmation-modal/confirmation-modal.component.html`. I think I might have copied this from there.

shared/src/store/action-reducers/exercise.ts
```diff
@@ -128,6 +143,55 @@ export namespace ExerciseActionReducers {
...
+        reducer: (draftState, { mode, partialExport }) => {
+            // TODO: Use `migratePartialExport` function that is currently defined in the backend.
+            const mutablePartialExport = cloneDeepMutable(partialExport);
+            const validationErrors =
+                validateExerciseExport(mutablePartialExport);
+            if (validationErrors.length > 0) {
+                throw new ReducerError(
+                    `Cannot import templates due to validation errors: ${validationErrors}`
+                );
+            }
```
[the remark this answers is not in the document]
ClFeSc: Ah, I see what you mean. Yes, you're right, I'll remove it here.

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.ts
```diff
@@ -82,4 +87,17 @@ export class TrainerMapEditorComponent {
...
         });
     }
+
+    public importingTemplates = false;
+    public async importPartialExport(fileList: FileList) {
+        try {
+            this.importingTemplates = true;
+            openPartialImportOverwriteModal(
+                this.ngbModalService,
+                await fileList.item(0)?.text()
```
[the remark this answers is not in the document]
ClFeSc: This would result in a double validation, though, as the action will be validated later on.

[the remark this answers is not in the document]
Dassderdie: The idea is to validate (and migrate) it here **instead**.
This would not only improve the code (no confusing string that is passed to the modal) but also improve the UX as you get an error message if you try to import another JSON file before opening the modal.

shared/src/store/action-reducers/exercise.ts
```diff
@@ -128,6 +142,54 @@ export namespace ExerciseActionReducers {
...
+                } else {
+                    // Remove all vehicles from all alarm groups as all existing vehicle templates are being removed
+                    draftState.alarmGroups = Object.fromEntries(
+                        Object.entries(draftState.alarmGroups).map(
+                            ([id, alarmGroup]) => {
+                                alarmGroup.alarmGroupVehicles = {};
+                                return [id, alarmGroup];
+                            }
+                        )
+                    );
```
Dassderdie, line 184:
```suggestion
                    for (const alarmGroup of draftState.alarmGroups) {
                        alarmGroup.alarmGroupVehicles = {};
                    }
```

shared/src/store/action-reducers/exercise.ts
```diff
@@ -151,3 +213,26 @@ function refreshTransfer(
...
+    // `patientCategories` don't have an `id`
+    const templateType = ['mapImageTemplates', 'vehicleTemplates'] as const;
+    for (const property of templateType) {
+        const templates = copy[property];
+        if (templates !== undefined) {
+            for (const template of templates) {
+                template.id = uuid();
+            }
+        }
+    }
```
Dassderdie, line 236:
```suggestion
    const templateTypes = ['mapImageTemplates', 'vehicleTemplates'] as const;
    for (const templateType of templateTypes) {
        const templates = copy[templateType];
        if (templates !== undefined) {
            for (const template of templates) {
                template.id = uuid();
            }
        }
    }
```

frontend/src/app/pages/exercises/exercise/shared/partial-import/partial-import-modal/partial-import-modal.component.ts
```diff
@@ -0,0 +1,59 @@
...
+        this.activeModal.close();
+    }
+
+    public async partialImportOverwrite(mode: 'append' | 'overwrite') {
+        this.importingPartialExport = true;
+        try {
+            const importInstance = plainToInstance(
+                PartialExport,
+                this.partialExport
+            );
```
Dassderdie, line 36:
```suggestion
```

You should be able to use `partialExport` directly, as this is just data.

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.ts
```diff
@@ -82,4 +90,28 @@ export class TrainerMapEditorComponent {
...
+            if (importedText === undefined) {
+                // The file dialog has been aborted.
+                return;
+            }
+            const importedPlainObject = JSON.parse(importedText) as object;
+            const importedInstance = plainToInstance(
+                PartialExport,
+                importedPlainObject
+            );
+            const migratedInstance = migratePartialExport(importedInstance);
```
Dassderdie, line 108:
```suggestion
            const importedPlainObject = JSON.parse(importedText) as object;
            const migratedPartialExport = migratePartialExport(importedPlainObject);
            if (!isValid(plainToInstance(
                PartialExport,
                migratedPartialExport 
            ))) {
                 throw Error('PartialExport is invalid');
            }
```

Dassderdie: At the moment, you are passing an instance to the modal (instead of a plain object) and, therefore, to the action. I would suspect this to crash, if this instance is later modified in a reducer (e.g., a template gets renamed) as `immer.js` would find a prototype in the template object.

Remember that all the `.create()-factory-functions were there to *not* have instances in the state.

FYI: Instead of the factory functions, one could also add a token to those classes in which immer should ignore the prototypes. But having plain objects without prototypes could be more performant (https://github.com/immerjs/immer/pull/941) and maybe less error-prone. #531

Dassderdie (changes requested):
I haven't tested it out again. I will do it after these changes are incorporated.

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.ts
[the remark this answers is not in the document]
ClFeSc: I just noticed that this could not fail. However, I don't know whether this is good behavior. When importing an invalid file, the migration function will choose `undefined` for all three categories, which is valid, just not useful. I'm not quite sure how to proceed there.

ClFeSc:
@lukasrad02 This should also now be ready for review. I've rebased this PR as getting the `--signoff` message to all commits across multiple merge commits was something I'd rather not do.

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 - this software has now a license (with some exceptions), see LICENSE-README.md
     - the software includes a file with third party acknoledgments, a license.html and the docker container will include a git archive of the source code downloadable in zip format
     - new npm run commands for licensing (third party acknowledgements) and creating source code archive
+- Add possibility to export only templates from an excercise and import them in another one.
+    - Patients, vehicles, and images templates can be selected.
```
lukasrad02, line 16: Either "image[s]" or "image[] templates"

lukasrad02:
When I try to join a new exercise, I receive the following error in the browser console and stay on the landing page:

> ERROR SyntaxError: The requested module 'http://localhost:4200/@fs/[redacted]/digital-fuesim-manv/frontend/.angular/cache/19.2.7/digital-fuesim-manv/vite/deps/digital-fuesim-manv-shared.js?v=b4b91550' doesn't provide an export named: 'preparePartialExportForImport'

shared/src/state-migrations/migrations.ts
```diff
@@ -42,6 +47,44 @@ export function migrateStateExport(
...
+            mapImageTemplates: mutablePartialExport.mapImageTemplates ?? [],
+            patientCategories: mutablePartialExport.patientCategories ?? [],
+            vehicleTemplates: mutablePartialExport.vehicleTemplates ?? [],
+        })
+    );
+    stateExport.fileVersion = mutablePartialExport.fileVersion;
+    stateExport.dataVersion = mutablePartialExport.dataVersion;
+    const migratedStateExport = migrateStateExport(stateExport as StateExport);
+    const mapImageTemplates =
+        mutablePartialExport.mapImageTemplates !== undefined
```
lukasrad02, line 67: Why are you checking for `undefined` here (same for patients and vehicles)? In the construction of the state export from the partial export, all undefined/null values are replaced by empty arrays.

I haven't tested it yet, but I think think that this can also cause unexpected overwrites, so that importing an export that only contains patients in overwrite-mode will delete all vehicles and map images from the current exercise.

shared/src/store/action-reducers/exercise.ts
```diff
@@ -197,6 +255,29 @@ function refreshTransfer(
...
+ * Prepare a {@link PartialExport} for import.
+ *
+ * This includes resetting UUIDs as this cannot be done in the reducer.
+ * @param partialExport The {@link PartialExport} to prepare.
+ */
+export function preparePartialExportForImport(
+    partialExport: PartialExport
+): PartialExport {
+    const copy = cloneDeepMutable(partialExport);
+    // `patientCategories` don't have an `id`
```
lukasrad02, line 268: They don't, but the categories contain templates and these templates have an id

lukasrad02:
> > When I try to join a new exercise, I receive the following error in the browser console and stay on the landing page:
> > > ERROR SyntaxError: The requested module '[http://localhost:4200/@fs/[redacted]/digital-fuesim-manv/frontend/.angular/cache/19.2.7/digital-fuesim-manv/vite/deps/digital-fuesim-manv-shared.js?v=b4b91550](http://localhost:4200/@fs/%5Bredacted%5D/digital-fuesim-manv/frontend/.angular/cache/19.2.7/digital-fuesim-manv/vite/deps/digital-fuesim-manv-shared.js?v=b4b91550)' doesn't provide an export named: 'preparePartialExportForImport'
> 
> I've also encountered this problem several times now. It seems to be an outdated cached version of `shared` that gets used in the frontend for some reason. Removing old build artifacts and rebuilding always fixed the issues for me, though.

Doing `rm -rf shared/dist/ frontend/dist/`, restarting the tasks and a force reload in the browser does not work in my case. Is there something I forgot to clean?

lukasrad02 (commented):
Only code-based review so far, as the frontend is currently broken for me

ClFeSc:
> > > When I try to join a new exercise, I receive the following error in the browser console and stay on the landing page:
> > > > ERROR SyntaxError: The requested module '[http://localhost:4200/@fs/[redacted]/digital-fuesim-manv/frontend/.angular/cache/19.2.7/digital-fuesim-manv/vite/deps/digital-fuesim-manv-shared.js?v=b4b91550](http://localhost:4200/@fs/%5Bredacted%5D/digital-fuesim-manv/frontend/.angular/cache/19.2.7/digital-fuesim-manv/vite/deps/digital-fuesim-manv-shared.js?v=b4b91550)' doesn't provide an export named: 'preparePartialExportForImport'
> > 
> > 
> > I've also encountered this problem several times now. It seems to be an outdated cached version of `shared` that gets used in the frontend for some reason. Removing old build artifacts and rebuilding always fixed the issues for me, though.
> 
> Doing `rm -rf shared/dist/ frontend/dist/`, restarting the tasks and a force reload in the browser does not work in my case. Is there something I forgot to clean?

I've done `git clean -dfX` (and then recreated the `.env` file). It's quite the "Holzhammermethode" but worked for me.

lukasrad02:
> > > > When I try to join a new exercise, I receive the following error in the browser console and stay on the landing page:
> > > > > ERROR SyntaxError: The requested module '[http://localhost:4200/@fs/[redacted]/digital-fuesim-manv/frontend/.angular/cache/19.2.7/digital-fuesim-manv/vite/deps/digital-fuesim-manv-shared.js?v=b4b91550](http://localhost:4200/@fs/%5Bredacted%5D/digital-fuesim-manv/frontend/.angular/cache/19.2.7/digital-fuesim-manv/vite/deps/digital-fuesim-manv-shared.js?v=b4b91550)' doesn't provide an export named: 'preparePartialExportForImport'
> > > 
> > > 
> > > I've also encountered this problem several times now. It seems to be an outdated cached version of `shared` that gets used in the frontend for some reason. Removing old build artifacts and rebuilding always fixed the issues for me, though.
> > 
> > 
> > Doing `rm -rf shared/dist/ frontend/dist/`, restarting the tasks and a force reload in the browser does not work in my case. Is there something I forgot to clean?
> 
> I've done `git clean -dfX` (and then recreated the `.env` file). It's quite the "Holzhammermethode" but worked for me.

Works. If this occurs more often and/or on other branches, we should take a look at this. Maybe the new, vite-based Angular build system behaves differently…

shared/src/state-migrations/migrations.ts
ClFeSc: I'm checking to explicitly not have the circumstances described in your second paragraph. Mind you, I'm checking against the original partial export, not the state created from it. I will add a comment explaining this. This is explicitly to prevent overriding other categories with `undefined`.

shared/src/store/action-reducers/exercise.ts
ClFeSc: Good catch, added this.

frontend/src/app/pages/landing-page/landing-page/landing-page.component.ts
```diff
@@ -98,7 +97,7 @@ export class LandingPageComponent {
                 }
                 case 'partial': {
                     throw new Error(
-                        'Dieser Typ kann zur Zeit nicht importiert werden.'
+                        'Dieser Typ kann nur innerhalb einer Übung importiert werden.'
```
lukasrad02, line 100: The user does not see this error message, but only some complex stack trace (at least in development mode, may be different in prod):

image

shared/src/state-migrations/migrations.ts
lukasrad02: Yes, I also just noticed this. Thanks for the clarifying comment.

lukasrad02:
> If this occurs more often and/or on other branches, we should take a look at this. Maybe the new, vite-based Angular build system behaves differently…

I've found a potential solution for this: <https://github.com/angular/angular-cli/issues/27552>. Will create a PR for this soon, but have to resolve some issues with indirect dependencies before.

frontend/src/app/pages/landing-page/landing-page/landing-page.component.ts
ClFeSc: I've redone this so that we post a message containing a more simple error, and no longer throw there. Throwing should be reserved for actual errors where a stack trace etc. is relevant.

lukasrad02 (approved):

lukasrad02:
> > If this occurs more often and/or on other branches, we should take a look at this. Maybe the new, vite-based Angular build system behaves differently…
> 
> I've found a potential solution for this: [angular/angular-cli#27552](https://github.com/angular/angular-cli/issues/27552). Will create a PR for this soon, but have to resolve some issues with indirect dependencies before.

Fix provided in #1055

ClFeSc:
@Dassderdie Unless you want to review anything again here, please resolve your open conversations etc. to allow for a non-admin-only merge :)

PR #523, opened 2022-09-17
Feature/improve treatment system

Dassderdie:
Continuation from #458


ClFeSc:
I'll look into it in the next few days.

anonym-HPI:
I will look into this PR at the end of next week, probably. 

About the capacity model, would it be easier, if I Implemented it in this PR or would it be similiar to write the migrations (this question seems to be irrelevant, I know, as migrations are written, but still)?

Dassderdie:
> I will look into this PR at the end of next week, probably.
> 
> About the capacity model, would it be easier, if I Implemented it in this PR or would it be similiar to write the migrations (this question seems to be irrelevant, I know, as migrations are written, but still)?

**Don't** add even more features in here! This should be merged, and all other features can be added in separate PRs. Else this will be a never-ending story ;)

THe migrations will not be easier/need less effort if made here.

backend/src/database/state-migrations/8-treatment-system-improvements.ts
```diff
@@ -0,0 +1,118 @@
...
+    Mutable,
+} from 'digital-fuesim-manv-shared';
+import {
+    cloneDeepMutable,
+    SpatialTree,
+    StrictObject,
+} from 'digital-fuesim-manv-shared';
+import type { Migration } from './migrations';
+
+export const treatmentSystemImprovements8: Migration = {
```
ClFeSc, line 13: I don't really like that you're exclusively using `any` here.

backend/src/database/state-migrations/8-treatment-system-improvements.ts
```diff
@@ -0,0 +1,118 @@
...
+        }
+    },
+    state: (state: any) => {
+        // Spatial tree
+        state.spatialTrees = cloneDeepMutable({
+            materials: SpatialTree.create(),
+            patients: SpatialTree.create(),
+            personnel: SpatialTree.create(),
+        });
+        for (const material of StrictObject.values(state.materials)) {
```
ClFeSc, line 38: Any particular reason for `StrictObject` over `Object`?

backend/src/database/state-migrations/8-treatment-system-improvements.ts
```diff
@@ -0,0 +1,118 @@
...
+        }
+        // VehicleTemplates
+        for (const vehicleTemplate of state.vehicleTemplates) {
+            vehicleTemplate.materials = vehicleTemplate.materials.map(
+                () => 'standard'
+            );
+        }
+        // PatientCategories
+        for (const patientCategory of state.patientCategories as Mutable<
+            ExerciseState['patientCategories']
```
ClFeSc, line 90: This is not safe to do at this point. You cannot assume anything about the state because it may have changed in later versions. When doing it like this you have to change this section every time this part of the state changes. Are you aware of that?

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/transfer-point-feature-manager.ts
```diff
@@ -80,8 +77,8 @@ export class TransferPointFeatureManager extends ElementFeatureManager<TransferP
             return false;
         }
         if (
-            droppedElement.type !== 'vehicle' &&
-            droppedElement.type !== 'personnel'
+            droppedElement.type !== 'vehicles' &&
+            droppedElement.type !== 'personnels'
```
ClFeSc, line 81: Why `personnels` instead of `personnel`?

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/style-helper/line-style-helper.ts
```diff
@@ -1,28 +1,32 @@
...
+    /**
+     * @param getOptions returns options that can be adjusted for each feature
+     * @param width the width of the line (in a non standardized unit).
+     */
     constructor(
-        private readonly scale: number,
-        private readonly color: ColorLike
+        private readonly getOptions: (
+            feature: Feature
+            // The width in the options would only be respected
```
ClFeSc, line 15: This comment seems to be incomplete.

ClFeSc (commented):
For now my review of `backend` and `frontend`.

backend/src/database/state-migrations/8-treatment-system-improvements.ts
Dassderdie: I believe `any` is the best solution here. We don't have typings for the previous state. Adding the typings manually like you did for the previous migrations is not a very nice way (more code, more complexity, creating the typings is also error prone). In addition, you will constantly run into problems because you need to e.g rename properties etc.
Migrations won't (or shouldn't) break if you change other parts of the codebase. Therefore one of typescripts biggest features (automatic refactoring) isn't of any use here.

TLDR: the cost/benefit for completely typed out code is not very good for the migrations.

backend/src/database/state-migrations/8-treatment-system-improvements.ts
Dassderdie: I would interpret this a bit different:

"You can't use this typing here, as the State type could change in later versions. You would notice this if `ExerciseState['patientCategories']` changes because then there would be a type error."

The same also goes for the `Mutable<ImageProperties>` at the end of this file.

backend/src/database/state-migrations/8-treatment-system-improvements.ts
Dassderdie: I believe `StrictObject` should be the default to use in this codebase. It is correct that the stricter typings don't provide a benefit in this case.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/transfer-point-feature-manager.ts
Dassderdie: Because `personnel` is a bad variable name ^^

But you are correct. This currently prohibits a personnel from being transferred.

In retrospect, we should have chosen a different variable name or even explicitly used `personnels` instead of `personnel` for the plural (of course, documented somewhere). Having the same name for singular and plural leads to such mistakes...

```suggestion
            droppedElement.type !== 'personnel'
```

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/style-helper/line-style-helper.ts
Dassderdie:
```suggestion
            // The width in the options would only be used on the initial render and then changed due to the zoom
```

backend/src/database/state-migrations/8-treatment-system-improvements.ts
ClFeSc: Yes, true. I agree that the whole system is not very good. Also that you get the initial state for action migrations is worthless as this is the state after all migrations, which could be vastly different than the state you expected when writing the migration. At some point I'll refactor this a bit, I think.

backend/src/database/state-migrations/8-treatment-system-improvements.ts
ClFeSc: Sounds reasonable, but we should document this somewhere.

backend/src/database/state-migrations/8-treatment-system-improvements.ts
ClFeSc: Yes, this was what I meant. You can leave it like this, but you must be aware that this might break at some point.

backend/src/database/state-migrations/8-treatment-system-improvements.ts
Dassderdie: > Yes, true. I agree that the whole system is not very good. Also that you get the initial state for action migrations is worthless as this is the state after all migrations, which could be vastly different than the state you expected when writing the migration. At some point I'll refactor this a bit, I think.

Shouldn't you get the state with the same version as the actions (so in this case 7?) instead of "the state after all migrations" (a state version >= 7)?
Or do I misunderstand your comment?

ClFeSc: This is the behavior you'd expect, correct. But this is not the current behavior. Currently, you get the state after the last migration.

Dassderdie: > This is the behavior you'd expect, correct. But this is not the current behavior. Currently, you get the state after the last migration.

That should be changed. This not only makes it unintuitive but could also introduce bugs if the state is changed and makes the code much harder to maintain.

Is there already an issue for this?

ClFeSc: I've added #524.

backend/src/exercise/patient-ticking.ts
```diff
@@ -86,43 +67,30 @@ function getDedicatedResources(
...
+    const cateringTypes: Catering = {
+        notarzt: 0,
+        notSan: 0,
+        rettSan: 0,
+        san: 0,
+        gf: 0,
+        material: 0,
+    };
+
     if (!patient.isBeingTreated) {
```
anonym-HPI, line 79: Maybe putting a comment here or changing it.
Right now there is a dependency as a patient can't be treated by material alone, but this choice shouldn't be done here probably (or at least be mentioned via comments). Right now this returns zero material, even if a patient has material.

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,330 @@
...
+ *     - if an element was moved: {@link element.position} needs the new position
+ *     - if treatment of an element should be removed: set {@link element.position} to `undefined`
+ * Also sets {@link Patient.visibleStatusChangedSinceTreatment} back to `false` (if element is a patient)
+ */
+export function updateTreatments(
+    state: Mutable<ExerciseState>,
+    element: Mutable<Material | Patient | Personnel>
+) {
+    // The requirement of this function is neither to result in a "perfect" treatment pattern
+    // nor to be independent from its history (calculating the treatments from scratch is allowed to differ
```
anonym-HPI, line 189: The comment is correct, but right now (at least what I implemented).
But recalculating everything from scratch should right now result in the same assignments as with the incremental calculation (as every personnel/material) does not take into account other personnel or material and how much a patient is already treated.

Only (which right now is not the case) if there is a dependency created between personnel/material and wether a patient has already enough treatment, a dependence is created.

shared/src/store/action-reducers/utils/spatial-elements.ts
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Adds an element with a position and executes side effects to guarantee the consistency of the state.
+ * Must be called if an element is added to the state
+ */
+export function addElementPosition(
+    state: Mutable<ExerciseState>,
+    elementType: SpatialElementType,
+    element: Mutable<Material | Patient | Personnel>
+) {
+    if (element.position === undefined) {
```
anonym-HPI, line 26: should this throw a reducerError?

shared/src/store/action-reducers/utils/spatial-elements.ts
```diff
@@ -0,0 +1,86 @@
...
+ * The position of the element must be changed via one of the function in this file.
+ * In addition, the respective functions must be called when an element gets added or removed.
+ */
+export type SpatialElementType = 'materials' | 'patients' | 'personnel';
+
+/**
+ * Adds an element with a position and executes side effects to guarantee the consistency of the state.
+ * Must be called if an element is added to the state
+ */
+export function addElementPosition(
```
anonym-HPI, line 21: this function is not needed (except for the name or so), as updateElementPosition could do exactly the same or not?

shared/src/store/action-reducers/vehicle.ts
```diff
@@ -139,11 +144,13 @@ export namespace VehicleActionReducers {
                 );
             }
             draftState.vehicles[vehicle.id] = cloneDeepMutable(vehicle);
-            for (const material of materials) {
-                draftState.materials[material.id] = cloneDeepMutable(material);
+            for (const material of cloneDeepMutable(materials)) {
+                draftState.materials[material.id] = material;
+                addElementPosition(draftState, 'materials', material);
```
anonym-HPI, line 149: this should be wrong, as material is initially loaded in the vehicle and its `position === undefined`

shared/src/store/action-reducers/vehicle.ts
```diff
@@ -139,11 +144,13 @@ export namespace VehicleActionReducers {
...
             }
-            for (const person of personnel) {
-                draftState.personnel[person.id] = cloneDeepMutable(person);
+            for (const person of cloneDeepMutable(personnel)) {
+                draftState.personnel[person.id] = person;
+                addElementPosition(draftState, 'personnel', person);
```
anonym-HPI, line 153: this should be wrong, as personnel is initially loaded in the vehicle and its `position === undefined`

anonym-HPI (changes requested):
First feedback round (no visual checking, no checking of migrations and not enough in tests). See also my two commits (at least one comment is linked to the commits)

backend/src/database/state-migrations/8-treatment-system-improvements.ts
Dassderdie: Added in 8e3439e5b9623ae14530cfd879e8a4a6bd11cc2d

backend/src/exercise/patient-ticking.ts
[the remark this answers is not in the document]
Dassderdie: Your sentence doesn't make any sense to me.

I assume (solely based on where the comment is made) you mean that `isBeingTreated` only considers personnel. Therefore this function doesn't produce the expected return value if the patient is only treated by material.

Can't we just remove the `isBeingTreated` check here?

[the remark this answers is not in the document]
anonym-HPI: > Your sentence doesn't make any sense to me.
> 
> I assume (solely based on where the comment is made) you mean that `isBeingTreated` only considers personnel. Therefore this function doesn't produce the expected return value if the patient is only treated by material.
> 
> Can't we just remove the `isBeingTreated` check here?

I would not remove the `isBeingTreated` check, if we remove this check we could discuss if the variable `isBeingTreated` should be removed, as it is just kind of a performance caching value.

Right now `isBeingTreated === false` means no personnel is treating. If the material calculation is set to after this check we will return that this patient has `zero` material associated with him, even if he does. 
This comes from the dependence that we know that it doesn't matter if a patient has material or not when there is no personnel treating him/her, this is an indirect dependence.
But this is a dependence that doesn't need to be there. The dependence of `isBeingTreated === false` when no personnel is treating a patient is there as this is the whole purpose of this variable (to tell if a patient has at least one personnel treating him/her), but `isBeingTreated` doesn't say right now anything about how much material a patient has assigned.
This whole thing is right now not important as a patient without personnel will change his/her state independently from material assigned to him/her.
With putting it before we remove this dependence, otherwise I would advise at least writing a comment or description.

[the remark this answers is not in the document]
Dassderdie: Removing the `isBeingTreated`-check here would fix the return value of the function. It would also not impact the performance of this function by a significant amount as in the case that `isBeingTreated` is false, `patient.assignedPersonnelIds` is an empty object.

Your current solution with moving the `material.length` stuff up is not very elegant and could lead to future confusion/bugs.

So my two proposals:
1. Remove the `isBeingTreated`-check in this function
    - Fixes `getDedicatedResources`
    - Less code
2. Replace `isBeingTreated` everywhere with `Patient.isTreatedByPersonnel = (patient)=> !isEmpty(patient.assignedPersonnelIds)` 
    - Better name
    - The calculation is very fast and should therefore be preferred over setting the variable each time it changes
    - Performance impact should be minimal.

1. Should be done in this pull request
2. Could maybe be done by the next team (easy task to get familiar with refactoring (state, writing a migration, ...))

frontend/src/app/state/exercise/exercise.selectors.ts
```diff
@@ -118,34 +118,43 @@ export const selectExerciseStatus = (state: AppState) =>
...
+                        .filter((patient) => patient.position !== undefined)
+                        .map((patient) => ({
+                            id: `${element.id}:${patient!.id}` as const,
+                            catererPosition: element.position!,
+                            patientPosition: patient!.position!,
+                        }));
+                })
+                // To improve performance, all not needed lines are removed
+                // Lines where both ends are not in the viewport are not relevant for the user
+                .filter(
```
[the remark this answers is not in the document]
Dassderdie: I believe this would be premature optimisation.

shared/src/store/action-reducers/utils/spatial-elements.ts
Dassderdie: The general idea for the API here was: 
* `addElementPosition` must be called when an element is added to the state.
* `updateElementPosition` must be called when the position of an (previously) added element changes
* `removeElementPosition` must be called when an element that was added gets deleted

The knowledge that elements without a position are not in the SpatialTree should be kept internally and not exposed outside this API.

I'll look at these three comments about it again later.

shared/src/models/patient.ts
```diff
@@ -141,6 +124,14 @@ export class Patient {
     @Min(0)
     public readonly timeSpeed: number = 1;
 
+    /**
+     * Whether the {@link getVisibleStatus} of this patient has changed
+     * since the last time it was updated which personnel and materials treat him/her.
+     * Use this to prevent unnecessary recalculations for patients that didn't change -> performance optimization.
+     */
+    @IsBoolean()
+    public readonly visibleStatusChangedSinceTreatment: boolean = false;
```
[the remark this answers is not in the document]
Dassderdie: I couldn't come up with a better name. Therefore I would leave it like this.

shared/src/store/action-reducers/utils/calculate-treatments.ts
Dassderdie: > The comment is correct.

So this can be resolved?

shared/src/store/action-reducers/utils/spatial-elements.ts
anonym-HPI: I see, but should `addElementPosition` be called when the `Position === undefined` (see here https://github.com/hpi-sam/digital-fuesim-manv/pull/523#discussion_r978822501) ?

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/transfer-lines-feature-manager.ts
```diff
@@ -19,6 +19,7 @@ export class TransferLinesFeatureManager
     >
     implements FeatureManager<Feature<LineString>>
 {
+    readonly type = 'transferLines';
```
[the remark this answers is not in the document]
Dassderdie: A single instance of a FeatureManager represents all elements of this type.
To use the plural for types is, from my experience, a best practice. Using the plural form, you can easily reuse the type for routes or as a key in data structures without having to remap it manually. In this concrete case, the `type` in the `FeatureManager` is only a temporary solution. The next step would be to give each element (`Material`, `Personnel`, `Patient`, ...) its own `type`, thereby making the type here obsolete.
We could then leverage the fact that `state[element.type][element.id] = element`.

shared/src/store/action-reducers/utils/spatial-elements.ts
Dassderdie: >

> I see, but should `addElementPosition` be called when the `Position === undefined` (see here [#523 (comment)](https://github.com/hpi-sam/digital-fuesim-manv/pull/523#discussion_r978822501)) ?

According to my API definition: Yes.
> `addElementPosition` must be called when an element is added to the state.

frontend/src/app/state/exercise/exercise.selectors.ts
```diff
@@ -118,34 +118,43 @@ export const selectExerciseStatus = (state: AppState) =>
...
+                    if (element.position === undefined) {
+                        return [];
+                    }
+                    return Object.keys(element.assignedPatientIds)
+                        .map((patientId) => patients[patientId]!)
+                        .filter((patient) => patient.position !== undefined)
+                        .map((patient) => ({
+                            id: `${element.id}:${patient!.id}` as const,
+                            catererPosition: element.position!,
+                            patientPosition: patient!.position!,
```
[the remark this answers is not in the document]
Dassderdie: Yeah, this was probably a merge error somewhere in the history of this branch...

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/element-manager.ts
```diff
@@ -19,11 +19,17 @@ export abstract class ElementManager<
         UnsupportedChangeProperties
     > = Exclude<ReadonlySet<keyof Element>, UnsupportedChangeProperties>
 > {
+    /**
+     * This type is used to identify the type of an element, e.g., when it was (drag &) dropped on something
```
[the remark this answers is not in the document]
Dassderdie: "make it more descriptive" is not a good review comment. In such cases, you should try to come up with your own guess/proposal of what should be there because I now can only guess what exactly wasn't clear to you...

Keep in mind that this will probably be replaced in the near future with a `type` property in each element.

```
When an element gets (dragged &) dropped, this identifies the type of the dropped element.
@example `materials`
```

shared/src/store/action-reducers/utils/spatial-elements.ts
Dassderdie: A reducer error should only be thrown if the action is invalid. Else this would be defensive programming (an anti-pattern).

shared/src/store/action-reducers/utils/spatial-elements.ts
Dassderdie: Another proposal for an API:

```ts
/**
 * @param previousPosition if an element gets added, this must be `undefined`
 * @param newPosition if an element gets removed, this must be `undefined`
 * Must be called every time the position of an element changes (including if it gets added or removed)
 * Updates the Spatial tree and the element's position property to {@link newPosition}.
 */
function updateElementPosition(state, elementType, elementId, previousPosition: Position | undefined, newPosition: Position | undefined)
```

Dassderdie: Or: 
```ts
/**
 * @param element the element whose position should be updated (the position property must still be the position before the update - exception: the element just got added and the previous position is the same as the new position)
 * @param elementType TODO: incorporate this into element
 * @param newPosition if an element gets removed, this must be `undefined`
 * Must be called every time the position of an element changes (including if the element gets added or removed)
 * Updates the Spatial tree and the element's position property to {@link newPosition}.
 */
function updateElementPosition(state, element: Element, elementType, newPosition: Position | undefined)
```

shared/src/store/action-reducers/utils/calculate-treatments.ts
anonym-HPI: > > The comment is correct.
> 
> So this can be resolved?

You mean in GitHub clicking "resolved"?
The comment says something that is technically correct, but not needed.

Right now you can recalculate everything from scratch and it should result in the exact same assignements as with this incremental calculation.

backend/src/exercise/patient-ticking.ts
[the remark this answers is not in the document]
anonym-HPI: First proposal is good, can be renamed later, maybe putting a TODO to the variable `isBeingTreated` and pointing to your second proposal?

shared/src/store/action-reducers/utils/spatial-elements.ts
anonym-HPI: Something like that the two proposal could be better than to have 3 different functions, you only need to remember "always call this function to change the position of an element".
But having three different functions could also make more clear what is happining. 

What about making it impossible to change the position in any other way than through this (or three) function(s)?
(Using position setter and getter)?

Dassderdie: Getter and setter do not work in this context. I think a comment `Reducers should only change the position via {@link ...}` is best one could do...

shared/src/models/personnel.ts
```diff
@@ -39,12 +43,31 @@ export class Personnel {
...
+    public readonly treatmentRange: number;
+
     @ValidateNested()
     @Type(() => ImageProperties)
     public readonly image: ImageProperties;
 
     /**
-     * if undefined, is in vehicle with {@link vehicleId} or in transfer
+     * If undefined, the personnel is either in the vehicle with {@link this.vehicleId} or in transfer.
+     * Exclusive-or to {@link transfer}
```
Dassderdie, line 70:
```suggestion
```

If the personnel is in a vehicle, neither `transfer` nor `position` has a value. Therefore this is not an exclusive or.

Additionally, I would change the comment on transfer to 
```
     * If undefined, the personnel is either in the vehicle with {@link this.vehicleId} or has a {@link position}.
```

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,330 @@
...
- * Maximum distance to treat the exact nearest patient
+ * @returns wether a material or personnel could treat a patient with {@link status} if it already {@link catersFor} patients
  */
-const specificThreshold = 0.5;
+function couldCaterFor(
+    status: Exclude<PatientStatus, 'white'>,
+    cateringElement: Mutable<Material | Personnel>,
+    catersFor: Mutable<CatersFor>
+) {
+    // black can't be treated (anymore) and blue patients should not (yet) be treated
```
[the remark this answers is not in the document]
Dassderdie: Not in this pull request!

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,330 @@
...
-            bluePatientsEnabled
-        );
-    });
+
+    if (isPersonnel(element) || isMaterial(element)) {
+        updateCatering(state, element);
+        return;
+    }
+
+    const updatedElements = new Set<UUID>();
```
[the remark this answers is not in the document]
Dassderdie: Renamed.
I changed the class comment to a normal one because this variable is only used in the next few lines.

backend/src/exercise/patient-ticking.ts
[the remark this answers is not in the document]
Dassderdie: > First proposal is good, can be renamed later, maybe putting a TODO to the variable `isBeingTreated` and pointing to your second proposal (to do that later)?

Done

shared/src/store/action-reducers/utils/calculate-treatments.ts
Dassderdie: > > > The comment is correct.
> > 
> > 
> > So this can be resolved?
> 
> You mean in GitHub clicking "resolved"?
Yes.

> Right now...

Just because it is like this now doesn't mean that it will stay like this in the future.
This comment's goal was to clarify that it isn't necessary to watch out for this when refactoring or trying to understand this algorithm.

It also depends a bit on whether one only looks at the semantic level or the strict representation of the end states. Our current JSON-Set implementations (e.g. `assignedMaterialIds`) are objects (and therefore have a key order).  Changing the order in which `updateTreatments` is called would change the order of the keys in these objects. The "normal" es6 JS-Set also guarantees to return all elements in insertion order. Though in reality, this shouldn't matter.

TLDR: I'd also be fine with removing this comment if you think it won't be necessary for future programmers.

Dassderdie:
The next team of students will start working on this project next wTherefore, this This branch should be merged until then.

This PR doesn't have to be perfect. Just good enough. Else it will never be ready to be merged...

frontend/src/app/state/exercise/exercise.selectors.ts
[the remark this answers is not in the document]
anonym-HPI: Not necessary, true (unless this would be a bottleneck)

shared/src/models/patient.ts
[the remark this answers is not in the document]
anonym-HPI: The name is long, true, but the name should not let somebody think something happens since treatment (treatment happens probably every second in the tick or not when patient has nobody assigned).

The name I wrote is more long, not coming up with a better name that is shorter, maybe:
`treatmentRecalculationSensible` (or usefull or ...), as before the recalculation wouldn't result in anything different.
But this name doesn't show that this is dependent on the visibleStatus

shared/src/store/action-reducers/utils/calculate-treatments.ts
anonym-HPI: The comment could stay there.
But the problem when this happens:
Right now the treatment calculation is independent from the history, resulting in when people drag something around it doesn't matter the order they drag personnel, material and patients around. Each time they will get the exact same.
Right now the assignment algorithm is stable and can therefore do incremental updates without any dependence on the history of updates and getting the same result.

If it is okay for people to have different assignments when the history is different, then it can stay. It is reasonable for this to happen, as personnel will probably not switch between two patients having the same category and stay with the one they already treat (maybe until the patient is stable or so).

But I would at least write in the comment that the algorithm is right now stable and that if things change so it is not anymore, it should be done not just carelessly but a be a design choice and implications should be considered.

E.g. an export could happen with all assigments deleted (as long as the spatialTrees are filled, which actually could also be rebuild at import with the mass import function of rbush) and recalculated on import and would result in the exact same assignements. Before export

Dassderdie: This was a design choice. I asked you whether this would be ok when I started reviewing this PR.

This comment should be removed if this isn't a design choice anymore (because now the algorithm is seen as by design stable and not just by chance).

anonym-HPI: I would remove the part, but keep the part with not being perfectly realistic assignment (or how it is right now written)

shared/src/models/personnel-template.ts
```diff
@@ -1,12 +1,14 @@
...
+import { maxTreatmentRange } from '../state-helpers/max-treatment-range';
 import {
     CanCaterFor,
-    ImageProperties,
     getCreate,
+    ImageProperties,
     PersonnelType,
 } from './utils';
 
+// TODO: These are not ?yet? saved in the state -> Decide whether they should and if not move this file from the models folder away
```
ClFeSc, line 11:
```suggestion
// TODO: These are not (yet) saved in the state -> Decide whether they should and if not move this file from the models folder away
```

shared/src/models/personnel-template.ts
```diff
@@ -15,6 +17,24 @@ export class PersonnelTemplate {
     @Type(() => CanCaterFor)
     public readonly canCaterFor: CanCaterFor;
 
+    /**
+     * Patients in this range are preferred over patients that are more far away (even if they are less injured).
```
ClFeSc, line 21:
```suggestion
     * Patients in this range are preferred over patients farther away (even if they are less injured).
```

shared/src/models/personnel-template.ts
```diff
@@ -15,6 +17,24 @@ export class PersonnelTemplate {
...
+     * Guaranteed to be <= {@link maxTreatmentRange}.
+     */
+    @IsNumber()
+    @Min(0)
+    @Max(maxTreatmentRange)
+    public readonly overrideTreatmentRange: number;
+
+    /**
+     * Only patients in this range around the material's position can be treated.
```
ClFeSc, line 30:
```suggestion
     * Only patients in this range around the personnel's position can be treated.
```

shared/src/models/utils/cater-for.ts
```diff
@@ -1,17 +1,35 @@
...
+    /**
+     * if {@link logicalOperator}  `=== 'and'` it is cumulative,
+     * how many {@link green} patients can the catering treat
+     * to the number already written in the {@link yellow} and {@link red} value
+     */
     @IsNumber()
+    @Min(0)
     public readonly green: number;
 
-    // TODO
```
ClFeSc, line 14: Why is this comment removed? I think I've marked this spot in a different way in #125 but I'd like to keep this comment here.

shared/src/models/utils/spatial-tree.ts
```diff
@@ -0,0 +1,176 @@
...
+     * @returns the ids of the elements in the search-circle, sorted by distance to {@link circlePosition}
+     */
+    public static findAllElementsInCircle(
+        spatialTree: Mutable<SpatialTree>,
+        circlePosition: Position,
+        radius: number
+    ): UUID[] {
+        if (radius <= 0) {
+            return [];
+        }
```
ClFeSc, line 117: Is this correct? Wouldn't `radius === 0` find elements that have exactly the same coordinates?

shared/src/store/action-reducers/patient.ts
```diff
@@ -117,9 +124,14 @@ export namespace PatientActionReducers {
     export const movePatient: ActionReducer<MovePatientAction> = {
         action: MovePatientAction,
         reducer: (draftState, { patientId, targetPosition }) => {
+            // Check whether the patient exists
             const patient = getElement(draftState, 'patients', patientId);
```
ClFeSc, line 128: This is already done in the spatial function.

shared/src/store/action-reducers/transfer.ts
```diff
@@ -143,14 +151,21 @@ export namespace TransferActionReducers {
...
+            // Remove the position of the element
+            if (elementType === 'personnel') {
+                removeElementPosition(draftState, 'personnel', element.id);
+            } else {
+                element.position = undefined;
+            }
+
             // Set the element to transfer
-            delete element.position;
             element.transfer = {
```
ClFeSc, line 162: Isn't there a difference between `element.position = undefined` (new) and `delete element.position` (old)?

shared/src/store/action-reducers/utils/calculate-treatments.spec.ts
```diff
@@ -23,38 +25,49 @@ interface Catering {
...
-            draftState.patients[patientId]!.isBeingTreated = false;
-        });
+            for (const patientId of catering.patientIds) {
+                cateringElement.assignedPatientIds[patientId] = true;
+            }
+        }
         return draftState;
     });
-    expect(shouldState).toStrictEqual(beforeState);
+    expect(newState).toStrictEqual(shouldState);
```
ClFeSc, line 66: `expect(received).toStrictEqual(expected)` Is this still the correct order here?

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,333 @@
...
+ */
 interface CatersFor {
     red: number;
     yellow: number;
     green: number;
 }
 
 /**
- * Maximum distance to treat the exact nearest patient
+ * @returns wether a material or personnel could treat a patient with {@link status} if it already {@link catersFor} patients
```
ClFeSc, line 23:
```suggestion
 * @returns whether a material or personnel could treat a patient with {@link status} if it already {@link catersFor} patients
```

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,333 @@
...
+function isPersonnel(
+    element: Material | Patient | Personnel
+): element is Personnel {
+    return (element as Personnel).personnelType !== undefined;
+}
+
+function isMaterial(
+    element: Material | Patient | Personnel
+): element is Material {
+    // as Material does not include any distinguishable, we will check if it is not of type Personnel or Patient
```
ClFeSc, line 112:
```suggestion
    // as Material does not include any distinguishable properties, we will check if it is not of type Personnel or Patient
```

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,333 @@
...
+    return (
+        (element as Personnel).personnelType === undefined &&
+        (element as Patient).personalInformation === undefined
```
ClFeSc, line 115:
```suggestion
        !isPersonnel(element) &&
        !isPatient(element)
```

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,333 @@
...
-                catering,
-                catersFor,
+    const patientsInTreatmentRange: Mutable<Patient>[] =
+        SpatialTree.findAllElementsInCircle(
+            state.spatialTrees.patients,
+            cateringElement.position,
+            cateringElement.treatmentRange
+        )
+            // Filter out every patient in the overrideTreatmentRange
+            .filter((patientId) => !cateredForPatients.has(patientId))
```
ClFeSc, line 300: You can filter by `status` here as you're ignoring every non `'red' | 'yellow' | 'green'` patient later on.

shared/src/store/action-reducers/utils/spatial-elements.ts
ClFeSc: But isn't this already defensive programming? The docs comment clearly states that the `element` must have a position.

ClFeSc (changes requested):
Now I've gone through all of the changes and left some comments. I still have to try it out, but I don't expect any surprises there (ideally, I wouldn't even notice it).

shared/src/store/action-reducers/utils/spatial-elements.ts
anonym-HPI: @ClFeSc what do you think and more intuitive/logic to use of the two proposals or the one that is right now in the PR (with three functions)?

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,333 @@
...
+        updateCatering(state, getElement(state, elementType, elementId));
+    }
+}
+
+function removeTreatmentsOfElement(
+    state: Mutable<ExerciseState>,
+    element: Material | Patient | Personnel
+) {
+    if (isPatient(element)) {
+        const patient = getElement(state, 'patients', element.id);
```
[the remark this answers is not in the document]
anonym-HPI: @ClFeSc What exactly do you mean?

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,333 @@
...
-                catersFor,
-                patient,
-                pretriageEnabled,
-                bluePatientsEnabled
-            )
-        ) {
-            break;
-        }
+    if (
+        cateringElement.treatmentRange <= 0 ||
```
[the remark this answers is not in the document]
anonym-HPI: In this model right now treating a patient with range `0` is not allowed, they would have to put them exactly on the same position, resulting in no cateringLine and therefore no feedback. 
Right now setting a range to `0` is equivilant of deactivating this circle (have thought of `-1` or so).

Probably a comment explaining setting this to `0` means this is deactivated/skipped for this personnel/material would probably be good.

shared/src/models/utils/spatial-tree.ts
anonym-HPI: See the other comment in `calculate-treatments.ts`.

Also `knn` handles `radius === 0` as `radius === infinity` resulting in not the correct result, thefore I had implemented before something like this:
https://github.com/hpi-sam/digital-fuesim-manv/commit/7c7254f33875a6ff542fbb727110ce0d1cfa5457

Maybe putting a comment here about what to consider

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
anonym-HPI: Gets longer, yes, I will write this into backlog.

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
anonym-HPI: @Dassderdie @ClFeSc I tried removing the `getElement` and just write like `const patient = element;`.
This worked for `patient`, but not for `material` or for `personnel` (as the isXXX function does return not a `Mutable` type).
I could fix it but would be not good, I am also not sure why it works for `patient`.
Maybe writing the `isXXX` function to something that returns `Mutable` when it is really `Mutable` or something like that? Or we just keep it that way and maybe write `TODO` comment that this should be fixed when every element gets a type variable.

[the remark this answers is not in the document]
ClFeSc: > @Dassderdie @ClFeSc I tried removing the `getElement` and just write like `const patient = element;`. This worked for `patient`, but not for `material` or for `personnel` (as the `isXXX` functions do return not a `Mutable` type). I could fix it but would be not good (@Dassderdie maybe you know how to write this, like in `getElement`?), I am also not sure why it works for `patient`. Maybe writing the `isXXX` function to something that returns `Mutable` when it is really `Mutable` or something like that? Or we just keep it that way and maybe write `TODO` comment that this should be fixed when every element gets a type variable.

I don't know what you mean with the `isXXX` functions as they return a `boolean`, but you're completely right that this gets the mutable version of the `element` while `element` itself is immutable. So leave it like it is, my bad.

shared/src/models/utils/spatial-tree.ts
ClFeSc: I get you're reasoning, so please add a comment.

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
ClFeSc: I get you're reasoning, so please add a comment.

shared/src/models/personnel.ts
anonym-HPI: @Dassderdie done

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
anonym-HPI: > > @Dassderdie @ClFeSc I tried removing the `getElement` and just write like `const patient = element;`. This worked for `patient`, but not for `material` or for `personnel` (as the `isXXX` functions do return not a `Mutable` type). I could fix it but would be not good (@Dassderdie maybe you know how to write this, like in `getElement`?), I am also not sure why it works for `patient`. Maybe writing the `isXXX` function to something that returns `Mutable` when it is really `Mutable` or something like that? Or we just keep it that way and maybe write `TODO` comment that this should be fixed when every element gets a type variable.
> 
> I don't know what you mean with the `isXXX` functions as they return a `boolean`, but you're completely right that this gets the mutable version of the `element` while `element` itself is immutable. So leave it like it is, my bad.

I changed `element` in the input to be `Mutable` but typescript is clever (or not) enough to see that I check with `isPatient`, ... if the element is `Patient` or `Material` or `Personnel` (but not checking if `Mutable<Patient>` or `Mutable<Materiall>` and so on.
Even restarting TypeScript didn't help.
I don't get why it works for `patient` and not the rest, the `isXXX` are similiar enough that this shouldn't happen.

Should I change `element` in the input already to be `Mutable` and just write a comment like this?:
```
// TODO: when elements have their own type saved don't use const patient = getElement(state, 'patients', element.id);
// instead use const patient = element;
// same for personnel and material in the other if statements
```

`element` right now is always put into as `Mutable` anyway.

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
anonym-HPI: Done

shared/src/store/action-reducers/utils/calculate-treatments.ts
anonym-HPI: Right now `blue` patients (when `blue` patients are deactivated) are treated as `red` (else as `blue`) and `white` are treated as `yellow`, therefore `Patient.getVisibleStatus` is needed to be called before filtering patients by their status category/color.

shared/src/models/patient.ts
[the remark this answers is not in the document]
Dassderdie: The variable name should state what currently is and not what should be because you need the former information for a refactoring.

What about `visibleStatusChanged`?
There is a class comment that explains this variable more in detail, therefore this should be enough.

shared/src/models/utils/cater-for.ts
Dassderdie: A TODO comment should always state what should be done. I could imagine I removed it because during some refactoring in this bloated PR because I could not identify what it meant. With the context of the stricter validation it makes sense.

I'll add it back again.

shared/src/store/action-reducers/transfer.ts
Dassderdie: Using `delete` also removes the key whereas setting it to `undefined` only changes the value.
I would expect the key to still be there but just the value set to undefined.

To remove those concerns and also to avoid having different objects after serialisation (`Json.parse(Json.stringify({ a: undefined}))`)  we should replace all occasions of `undefined` in the models with `null`. But this should be done at a later point in time.

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -1,255 +1,330 @@
...
+ *     - if treatment of an element should be removed: set {@link element.position} to `undefined`
+ * Also sets {@link Patient.visibleStatusChangedSinceTreatment} back to `false` (if element is a patient)
+ */
+export function updateTreatments(
+    state: Mutable<ExerciseState>,
+    element: Mutable<Material | Patient | Personnel>
+) {
+    // The requirement of this function is neither to result in a "perfect" treatment pattern
+    // nor to be independent from its history (calculating the treatments from scratch is allowed to differ
+    // from a solution that has been incrementally produced by calls to this function)
```
anonym-HPI, line 188:
```suggestion
    // The requirement of this function is neither to result in a "perfect" treatment pattern
    // but more in a treatment pattern that makes it intuitive and realistic enough
    // without doing everything for the participants.
    // The calculation is right now stable - meaning when completely done from scratch it would result in the exact outcome 
    // and it is not bound to the order of elements being moved.
    // This could be changed later, but keep in mind to make it still intuitive 
    // (participants moving different elements around in a different order and getting a [complete] different result)
```

Maybe something like that?

anonym-HPI: @Dassderdie what do you think?

shared/src/store/action-reducers/transfer.ts
ClFeSc: Right. So the new behavior is better than the old one? Or is it good (or bad) either way?

shared/src/store/action-reducers/utils/calculate-treatments.spec.ts
ClFeSc: So it's correct now, right? I've always had trouble thinking about this specific order in this test.

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
ClFeSc: Ah, I see. You could add the same as with `State` for `getElement` to these `isXXX` functions.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/element-manager.ts
```diff
@@ -19,11 +19,19 @@ export abstract class ElementManager<
         UnsupportedChangeProperties
     > = Exclude<ReadonlySet<keyof Element>, UnsupportedChangeProperties>
 > {
+    /**
+     * When an element gets (dragged &) dropped, this identifies the type of the dropped element.
+     * @example `patients`
+     */
+    abstract readonly type: string;
```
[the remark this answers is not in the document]
Dassderdie: This class doesn't know anything about `Material` or `Personnel`. It only expects some string here.
The child classes make this string then specific.

In addition, this property will be removed soon. We will then think about higher typesafety again.

shared/src/store/action-reducers/utils/calculate-treatments.ts
ClFeSc: Right. You could filter after the `map`, though.

shared/src/store/action-reducers/transfer.ts
Dassderdie: I would say it is a bit better but still not great.

shared/src/store/action-reducers/utils/calculate-treatments.spec.ts
Dassderdie: I think it is correct now (I believe it was the other way around before and confused me a lot ^^)

shared/src/store/action-reducers/utils/spatial-elements.ts
ClFeSc: I'd stick with the current three functions (`add`, `update`, `delete`) as it's the most flexible for the future (while not being overly complicated - you could call `update` from `add` for example).

shared/src/store/action-reducers/vehicle.ts
```diff
@@ -100,8 +105,8 @@ export class LoadVehicleAction implements Action {
 
     @IsString()
     public readonly elementToBeLoadedType!:
-        | 'material'
-        | 'patient'
+        | 'materials'
+        | 'patients'
```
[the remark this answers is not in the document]
Dassderdie: Yes! Good catch.

But I'm a bit confused that I didn't encounter any error when trying to migrate any of the states...

shared/src/store/action-reducers/transfer.ts
ClFeSc: Ok, so we can leave it as it is.

shared/src/store/action-reducers/utils/calculate-treatments.spec.ts
ClFeSc: I think you're correct. The test was much more confusing before this PR as it removed part of the state and checked that one.

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
anonym-HPI: That was an idea, do you know how to that (that it automatically results in `Mutable` when `Mutable` is entered and not when not?
Else, the `isXXX` functions could be changed to just accept `Mutable` (as it is only called with `Mutable` elements).

shared/src/store/action-reducers/vehicle.ts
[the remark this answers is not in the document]
ClFeSc: There is no error because the validation is broken, and it is translated to no behavior, so nothing gets loaded (apart from `personnel` which didn't change).

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
ClFeSc: As I said, you can do the same thing as with `State` in `getElement` (using a generic).

shared/src/store/action-reducers/utils/calculate-treatments.ts
anonym-HPI: Do you want directly to use `getCateringStatus` and `Patient.getVisibleStatus` after the map and groupBy directly?
Else wouldn't this result in calling multiple times the two mentioned functions?
(Or how would you calculate `patientsPerStatus` or do the loops?)

shared/src/models/patient.ts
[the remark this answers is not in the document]
anonym-HPI: > The variable name should state what currently is and not what should be because you need the former information for a refactoring.
> 
> What about `visibleStatusChanged`? There is a class comment that explains this variable more in detail, therefore this should be enough.

If the comment explains the rest (`visibleStatusChangedSinceLastTreatmentCalculation`), I would agree with this name, not the best but is not super long and better to finish than to discuss more time on this name.

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
anonym-HPI: > As I said, you can do the same thing as with `State` in `getElement` (using a generic).

I did this

```
function isPatient<PatientType extends Mutable<Patient> | Patient>(
    element: Material | PatientType | Personnel
): element is PatientType {
    return (element as PatientType).personalInformation !== undefined;
}

function isPersonnel<PersonnelType extends Mutable<Personnel> | Personnel>(
    element: Material | Patient | PersonnelType
): element is PersonnelType {
    return (element as PersonnelType).personnelType !== undefined;
}
```

For patient it results in element being after the `isPatient` being of `Mutable<Patient> | Mutable<Patient>` (yes twice).
and for personnel it results in `Personnel | Mutable<Personnel>` which does not work.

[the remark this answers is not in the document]
anonym-HPI: I would just make these `isXXX` functions only take `Mutable` as this is what works right now, if somebody wants to use these functions with immutable elements before every element gets a type, this person can rewrite these functions.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/element-manager.ts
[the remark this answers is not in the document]
anonym-HPI: Ok, is there anywhere a TODO written about the typesafety? When yes, this can be resolved.

shared/src/store/action-reducers/utils/spatial-elements.ts
anonym-HPI: @Dassderdie from the three functions and your two proposals which do you prefer?

Three functions could be more flexible, but having only one and knowing you have to call exactly this function and have not to think which one you need right now, could be good as well.
Maybe doing the three functions but giving `update` the option to catch intended deletions or adding and doing them or disallowing this and throwing a reducerError as people should use the correct function for the correct action (update meaning having old and new position being defined).

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
anonym-HPI: @ClFeSc are you ok with that or do you have a proposal to make it with generics work correctly?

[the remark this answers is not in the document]
Dassderdie: > ```ts
> function isPatient<PatientType extends Mutable<Patient> | Patient>(
>     element: Material | PatientType | Personnel
> ): element is PatientType {
>     return (element as PatientType).personalInformation !== undefined;
> }
> 
> function isPersonnel<PersonnelType extends Mutable<Personnel> | Personnel>(
>     element: Material | Patient | PersonnelType
> ): element is PersonnelType {
>     return (element as PersonnelType).personnelType !== undefined;
> }
> ```

I think typescript can't automatically infer the correct value for `PatientType` because the argument's type is a union. This should work by calling it like this `isPatient<Mutable<Patient>>(element)` or `isPatient<Patient>(element)`.

Also, keep in mind that this will soon be obsolete because of the `type` property in the element itself.

shared/src/store/action-reducers/utils/calculate-treatments.ts
Dassderdie: Continuation in https://github.com/hpi-sam/digital-fuesim-manv/pull/523/files/76eaac75c386627cb4ad0d9bb9f8da05540ccb92#r983319421

shared/src/store/action-reducers/utils/calculate-treatments.ts
Dassderdie: > and it is not bound to the order of elements being moved.
I believe this is wrong. Changing the order of the elements being moved could result in a different order in  `SpatialTree.findAllElementsInCircle` for elements that have the same distance to the passed position. Therefore a different element could be treated.
I believe only the order of the `updateTreatments` and the already present treatments don't matter.

```suggestion
The requirement of this function is not to result in a "perfect" treatment pattern. Instead, it should just find an intuitive and reasonably realistic treatment pattern.
Currently, the treatment pattern algorithm is stable. This means that completely done from scratch, the result would semantically be the same. 
This could be changed later.
```

shared/src/store/action-reducers/utils/calculate-treatments.ts
Dassderdie: I think this would be premature optimization. Keep in mind that we are only looking at the patients between the `overrideTreatmentRange` and `treatmentRange` around the `cateringElement`.
Additionally, we would have to iterate over all elements again to filter them and would only save the few calculations needed in `groupBy` and save a tiny bit of memory.

shared/src/store/action-reducers/vehicle.ts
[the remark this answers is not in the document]
Dassderdie: > There is no error because the validation is broken, and it is translated to no behavior, so nothing gets loaded (apart from `personnel` which didn't change).

I wouldn't say the validation is broken - it just isn't implemented yet ^^
But your explanation makes complete sense.

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
Dassderdie: After looking at your changes, I think you can leave it like this. 👍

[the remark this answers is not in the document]
anonym-HPI: Yes, as this is obsolete when an element is changed, also making this compatible with immutable elements, when right now it is called everytime with `Mutable` would make it more complicated than it needs to be (KISS)

shared/src/store/action-reducers/utils/calculate-treatments.ts
anonym-HPI: > I think this would be premature optimization. Keep in mind that we are only looking at the patients between the `overrideTreatmentRange` and `treatmentRange` around the `cateringElement`. Additionally, we would have to iterate over all elements again to filter them and would only save the few calculations needed in `groupBy` and save a tiny bit of memory.

I would also leave it as it is, if you @ClFeSc have any suggestion about how to write it in an easier way, resulting in the same outcome, please make a suggestion.
Otherwise mark this as resolved if you are agreeing that this optimization is not necessarly needed.

shared/src/models/personnel.ts
anonym-HPI: Marking this as resolved, @Dassderdie if you think otherwise undo or ping me in this conversation.

shared/src/models/patient.ts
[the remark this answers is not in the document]
Dassderdie: Renamed to `visibleStatusChanged`

shared/src/models/utils/cater-for.ts
Dassderdie: Fixed with 4e542b4f842f7fd1c7afa51a2d20846181845808

Dassderdie:
I would propose this branch should be squashed when merging. This branch had an overproportional number of `Fix`, `Apply feedback`, and big refactoring commits.

anonym-HPI:
> I would propose this branch should be squashed when merging. This branch had an overproportional number of `Fix`, `Apply feedback`, and big refactoring commits.

Agreed, in the squash probably a few informations could be written what this commit does. I will write it (and enable auto-merge (squash).

shared/src/store/action-reducers/utils/calculate-treatments.ts
ClFeSc: Yes, I agree. I forgot the fact that only patients nearby are included.

ClFeSc (approved):
I think I'm happy with it now. Not sure whether I missed something, but I just cannot thoroughly check every single line again.

anonym-HPI:
Here the proposed commit text, would edit this live in this comment:
```
* introducing spatial-tree - a datastructure to more efficiently calculate elements around a position

* spatial-tree is used to make the treatment calculation more efficient

* spatial-tree can also be used to find every patient, personnel or material (or in the future any other type of elements with a position) in a radius or in a rectangle (e.g. a viewport)

* treatment calculation now happens incrementally and only locally - no calculation from scratch every time something changed or every 20th tick

* making refreshTreatments every 20th tick more efficient (only for patients there visibleStatusChanged - saving this in variable)

* every personnel and material now has its own override and general treatment range

* having a maximumTreatmentRange (important for calculation of treatments when a patient changes his/her position)

* changed how many patients personnel + material can treat (changing templates)

* logialOperator `and` is now commulative

* patients now have also a list of assigned material and personnel to them

* removed redundant PatientTickResult - replaced with PatientUpdate

* Refactor getElementFromFeature

* getImmutableElement + used in patient-ticking.ts

* getElement now automatically returns Mutable or an Immutable element, dependent of the state 

* Add material-template (having `standard` and `big` material)

* Change GwSan to have 4 `big` materials instead of 8 `standard`

* Add migrations

* Update allowedCommonJsDependencies for angular

Co-authored-by: Julian Schmidt <julian.r.schmidt.js@gmail.com>
Co-authored-by: anonym-HPI <68286419+anonym-HPI@users.noreply.github.com>
Co-authored-by: Clemens <68013019+ClFeSc@users.noreply.github.com>
```

If I missed something important tell me or add it or if something is there that doesn't need to be, you can delete it (have a local copy e.g. `* Update allowedCommonJsDependencies for angular` not sure if this should be removed)

shared/src/store/action-reducers/utils/spatial-elements.ts
```diff
@@ -0,0 +1,86 @@
...
+import { cloneDeepMutable } from '../../../utils';
+import { updateTreatments } from './calculate-treatments';
+import { getElement } from './get-element';
+
+/**
+ * The element types for which a spatial tree exists in the state to improve the performance.
+ * The position of the element must be changed via one of the function in this file.
+ * In addition, the respective functions must be called when an element gets added or removed.
+ */
+export type SpatialElementType = 'materials' | 'patients' | 'personnel';
```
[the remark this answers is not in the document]
anonym-HPI: Or was the discussion that it stays `personnel` (I couldn't find `personnels` anywhere in the code)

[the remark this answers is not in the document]
Dassderdie: The plural of `personnel` is `personnel`.

We decided nearly a year ago that we use it in keys in the grammatically correct form.
In some isolated instances, we still use `personnels` as a variable name.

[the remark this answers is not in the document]
anonym-HPI: I know that this is the plural, will resolve this.

ClFeSc:
> * making refreshTreatments every 20th tick more efficient (only for patients there visibleStatusChanged - saving this in variable)

What do you mean by `only for patients there visibleStatusChanged`? Otherwise I'm fine with the text.

shared/src/store/action-reducers/utils/spatial-elements.ts
Dassderdie: > Maybe doing the three functions but giving update the option to catch intended deletions or adding and doing them or disallowing this and throwing a reducerError as people should use the correct function for the correct action (update meaning having old and new position being defined).

Throwing a `ReducerError` when a function is called with the wrong arguments is defensive programming -> bad practice.

I would leave it as it currently is.

anonym-HPI: Then we will leave it, as it is.
But I would make the inputs consistent (where it is possible/sensible).
Either always directly give the element in or only the elementId (and correctly add or delete potential getElements where these functions are called) and call it always `state` or `draftState`

anonym-HPI:
> 

* making refreshTreatments every 20th tick more efficient (only for patients where their `visibleStatusChanged` - saving this in variable)

shared/src/store/action-reducers/utils/spatial-elements.ts
anonym-HPI: @Dassderdie commited changes, if you agree with them, this is resolved.

shared/src/store/action-reducers/utils/calculate-treatments.spec.ts
```diff
@@ -263,34 +270,30 @@ describe('calculate treatment', () => {
...
-                ids.redPatient = redPatient.id;
-                state.patients[greenPatient.id] = greenPatient;
-                state.patients[redPatient.id] = redPatient;
-                state.materials[material.id] = material;
+                    Position.create(3, 3)
+                ).id;
+                ids.material = addMaterial(state, Position.create(0, 0)).id;
             }
         );
+        console.log(ids);
```
ClFeSc, line 296: Just saw this log in a test run. Is it required for anything, or could we remove this sometime?

Dassderdie: I think this can be removed. This PR was so big that we completely missed that.

anonym-HPI: :thumbsup: This should probably not be in there.

PR #527, opened 2022-09-30
Pass the correct state version to action migrations

### LINKED ISSUE #524: Use specific version of state in action migrations
Action migrations currently get the most up-to-date state whereas it would be much better to receive the state migrated to the version of the action migration. E.g. for a migration from version 4 to 5, the action migration would receive the initial state migrated to version 5, regardless of the current latest possible version.

ClFeSc:
Fixes #524

### DIFF
--- a/backend/src/database/state-migrations/migrations.ts
+++ b/backend/src/database/state-migrations/migrations.ts
@@ -1,9 +1,10 @@
 import type {
-    ExerciseState,
+    ExerciseAction,
     Mutable,
     StateExport,
     UUID,
 } from 'digital-fuesim-manv-shared';
+import { applyAllActions, ExerciseState } from 'digital-fuesim-manv-shared';
 import type { EntityManager } from 'typeorm';
 import { RestoreError } from '../../utils/restore-error';
 import { ActionWrapperEntity } from '../entities/action-wrapper.entity';
@@ -52,76 +53,67 @@ export const migrations: {
     8: treatmentSystemImprovements8,
 };
 
-export async function migrateInDatabaseTo(
-    targetStateVersion: number,
-    currentStateVersion: number,
+export async function migrateInDatabase(
     exerciseId: UUID,
     entityManager: EntityManager
 ): Promise<void> {
-    const migrationFunctions = getMigrationFunctions(
-        currentStateVersion,
-        targetStateVersion
-    );
-    const stateMigrationFunctions = migrationFunctions.state;
-    const originalStates = await entityManager.findOne(ExerciseWrapperEntity, {
+    const exercise = await entityManager.findOne(ExerciseWrapperEntity, {
         where: { id: exerciseId },
-        select: { initialStateString: true, currentStateString: true },
     });
-    if (originalStates === null) {
+    if (exercise === null) {
         throw new RestoreError(
             'Cannot find exercise to convert in database',
             exerciseId
         );
     }
-    let initialState: object | null = null,
-        currentState: object | null = null;
-    if (stateMigrationFunctions.length > 0) {
-        initialState = JSON.parse(originalStates.initialStateString);
-        currentState = JSON.parse(originalStates.currentStateString);
-        stateMigrationFunctions.forEach((stateMigrationFunction) => {
-            initialState = stateMigrationFunction(initialState!);
-            currentState = stateMigrationFunction(currentState!);
-        });
-    }
-    const actionsMigrationFunctions = migrationFunctions.actions;
-    let actions: (object | null)[] | null = null;
-    if (actionsMigrationFunctions.length > 0) {
-        const originalActions = await entityManager.find(ActionWrapperEntity, {
+    // const targetVersion = ExerciseState.currentStateVersion;
+    const initialState = JSON.parse(exercise.initialStateString);
+    const currentState = JSON.parse(exercise.currentStateString);
+    const actions = (
+        await entityManager.find(ActionWrapperEntity, {
             where: { exercise: { id: exerciseId } },
             select: { actionString: true },
             order: { index: 'ASC' },
-        });
-        actions = originalActions.map((originalAction) =>
-            JSON.parse(originalAction.actionString)
-        );
-        initialState ??= JSON.parse(originalStates.initialStateString);
-        actionsMigrationFunctions.forEach((actionsMigrationFunction) => {
-            actionsMigrationFunction(initialState!, actions!);
-        });
-    }
+        })
+    ).map((action) => JSON.parse(action.actionString));
+    const {
+        currentStateVersion: newStateVersion,
+        currentState: newCurrentState,
+        history,
+    } = applyMigrations(exercise.stateVersion, currentState, {
+        initialState,
+        actions,
+    });
+    // Save exercise
     const patch: Partial<ExerciseWrapperEntity> = {
-        stateVersion: targetStateVersion,
+        stateVersion: newStateVersion,
     };
-    if (currentState !== null) {
-        patch.initialStateString = JSON.stringify(initialState);
-        patch.currentStateString = JSON.stringify(currentState);
-    }
+    // if (newCurrentState !== null) {
+    patch.initialStateString = JSON.stringify(history!.initialState);
+    patch.currentStateString = JSON.stringify(newCurrentState);
+    // }
     await entityManager.update(
         ExerciseWrapperEntity,
         { id: exerciseId },
         patch
     );
-    if (actions !== null) {
+    // Save actions
+    if (history!.actions !== undefined) {
         let patchedActionsIndex = 0;
         const indicesToRemove: number[] = [];
-        const actionsToUpdate: { index: number; actionString: string }[] = [];
-        actions.forEach((action, i) => {
+        const actionsToUpdate: {
+            previousIndex: number;
+            newIndex: number;
+            actionString: string;
+        }[] = [];
+        history!.actions.forEach((action, i) => {
             if (action === null) {
                 indicesToRemove.push(i);
                 return;
             }
             actionsToUpdate.push({
-                index: patchedActionsIndex++,
+                previousIndex: i,
+                newIndex: patchedActionsIndex++,
                 actionString: JSON.stringify(action),
             });
         });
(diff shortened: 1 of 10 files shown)


backend/src/database/state-migrations/migrations.ts
```diff
@@ -52,55 +54,59 @@ export const migrations: {
...
-    targetStateVersion: number,
-    currentStateVersion: number,
+/**
+ *
+ * @param targetStateVersion The new state version the exercise should have.
+ *          `'latest'` to always migrate as far as possible.
+ *          Must be at least as big as the current sate version.
+ */
+export async function migrateInDatabase(
+    targetStateVersion: number | 'latest',
```
Dassderdie, line 64: I think we could just omit this parameter and always migrate to the latest version. YAGNI

The same for the other migrate function.

backend/src/database/state-migrations/migrations.ts
```diff
@@ -52,55 +54,59 @@ export const migrations: {
...
-        });
-    }
-    const actionsMigrationFunctions = migrationFunctions.actions;
-    let actions: (object | null)[] | null = null;
-    if (actionsMigrationFunctions.length > 0) {
-        const originalActions = await entityManager.find(ActionWrapperEntity, {
+    const targetVersion =
+        typeof targetStateVersion === 'number'
+            ? targetStateVersion
+            : Math.max(...StrictObject.keys(migrations));
```
Dassderdie, line 80: Either 
* Change the `migrations` dictionary to an array where the index is the version and use `migrations.length`
* Assume (maybe with comment) that Object.keys().at(-1) is the highest stateNumber
* Make a public variable on the outer scope under `migrations` `const latestStateVersion = Math.max(...StrictObject.keys(migrations));`

backend/src/database/state-migrations/migrations.ts
```diff
@@ -52,55 +54,59 @@ export const migrations: {
...
+        if (typeof stateMigration === 'function') {
+            initialState = stateMigration(initialState);
+            currentState = stateMigration(currentState);
+        }
+        if (!actions) continue;
+        const actionMigration = migrations[currentStateVersion]?.actions;
+        if (typeof actionMigration === 'function') {
+            actionMigration(initialState, actions);
+        }
     }
```
Dassderdie, line 106: This should be an extra function as it is reused in the other migration function.

backend/src/database/state-migrations/migrations.ts
```diff
@@ -52,55 +54,59 @@ export const migrations: {
...
+    ).map((action) => JSON.parse(action.actionString));
+    for (
+        let currentStateVersion = exercise.stateVersion + 1;
+        currentStateVersion <= targetVersion;
+        currentStateVersion++
+    ) {
+        const stateMigration = migrations[currentStateVersion]?.state;
+        if (typeof stateMigration === 'function') {
+            initialState = stateMigration(initialState);
+            currentState = stateMigration(currentState);
```
Dassderdie, line 99: I believe this will create problems in the future.

It should always be that `applyAllActions(initialState, actions) == currentState`.
In the latest migrations in #523 we had to create a SpatialTree. The specific SpatialTree depends on the order of the actions that were performed. Therefore `applyAllActions(initialState, actions) == currentState` could be false. Besides failing checks with `isEqual()` (what could be very confusing) it also effects the state itself as the treatment algorithm could potentially have a different result if e.g. two patients have the same distance from a personnel and, due to them changing the order in the spatial tree, the personnel now treats someone else.

I would therefore recommend to first, migrate the `initialState` and all the actions to the wanted version, and then applying all the actions to get the `currentState`. If there is an Error during the application of the actions, we know that the migration is incorrect and this exercise could not be migrated.

Dassderdie: In `shared\src\store\time-travel-utilities.ts`
```ts
/**
 * @param actions all the actions that were applied to the {@link initialState} to get to the present state
 * @returns the state after applying all the actions
 */
export function applyAllActions(
    initialState: ExerciseState,
    actions: readonly ExerciseAction[]
): ExerciseState {
    // TODO: find a heuristic for whether using produce or deepCloning it
    // Default is produce
    return produce(initialState, (draftState) => {
        for (const action of actions) {
            applyAction(draftState, action);
        }
    });
}
```

Dassderdie (changes requested):
The few bigger things I noticed.

backend/src/database/state-migrations/migrations.ts
ClFeSc: No, mich simpler: `ExerciseState.currentStateVersion` contains the current version. Its correctness is asserted by the test.

backend/src/database/state-migrations/migrations.ts
ClFeSc: Hm, good point. Although I didn't change it in this PR, I believe it was incorrect even before (but the function was mich more complex so it was more difficult to see). I will change it to only migrate the `initialState` and the `actions` and then use them to create a new `currentState` (unless we have a `StateExport` without `history`).

backend/src/database/state-migrations/migrations.ts
```diff
@@ -52,76 +53,83 @@ export const migrations: {
...
+    //     if (typeof stateMigration === 'function') {
+    //         initialState = stateMigration(initialState);
+    //         currentState = stateMigration(currentState);
+    //     }
+    //     if (!actions) continue;
+    //     const actionMigration = migrations[currentStateVersion]?.actions;
+    //     if (typeof actionMigration === 'function') {
+    //         actionMigration(initialState, actions);
+    //     }
+    // }
```
anonym-HPI, line 102: This is commented out, does this get removed or included?

ClFeSc: Oh, forgot to remove this. Will do it later this evening I think.

ClFeSc: Done.

backend/src/database/state-migrations/migrations.ts
```diff
@@ -136,71 +128,86 @@ export async function migrateInDatabaseTo(
...
+    for (let i = currentStateVersion + 1; i <= targetVersion; i++) {
+        const stateMigration = migrations[i]?.state;
+        if (typeof stateMigration === 'function') {
+            if (history)
+                history.initialState = stateMigration(history.initialState);
+            else newCurrentState = stateMigration(newCurrentState);
+        }
+        if (!history) continue;
+        const actionMigration = migrations[i]?.actions;
+        if (typeof actionMigration === 'function') {
```
Dassderdie, line 192: Checking for `typeof function` is weird. I propose to make `actions` and `state` optional (`?`) or a union with `undefined` in `Migration` instead of a union with `null`.
The migrations are never part of the state/get stringified, therefore I would not prefer `null` here (Sorry, if I previously proposed `null` for this).

backend/src/database/state-migrations/migrations.ts
```diff
@@ -136,71 +128,86 @@ export async function migrateInDatabaseTo(
...
+    const { currentState, history } = applyMigrations(
+        stateExport.dataVersion,
+        stateExport.currentState,
+        stateExport.history
+            ? {
+                  initialState: stateExport.history.initialState,
+                  actions: stateExport.history.actionHistory,
+              }
+            : undefined
     );
```
Dassderdie, line 157: `stateExport` is immutable. I believe this mutates `stateExport`.

Either change the input to `Mutable<StateExport>` or use `cloneDeep()`.

Dassderdie (changes requested):
There seem to be some issues with Immutability (only in the typings - not the runtime).

Also, please don't force push. I always have to look at the whole PR instead of just the changes of the last commit during my reviews...

backend/src/database/state-migrations/migrations.ts
```diff
@@ -52,76 +53,67 @@ export const migrations: {
...
+    ).map((action) => JSON.parse(action.actionString));
+    const {
+        currentStateVersion: newStateVersion,
+        currentState: newCurrentState,
+        history,
+    } = applyMigrations(exercise.stateVersion, currentState, {
+        initialState,
+        actions,
+    });
+    // Save exercise
```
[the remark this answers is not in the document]
ClFeSc: Well, we also save the new state version. Maybe

```suggestion
    // Save exercise wrapper
```

?

backend/src/database/state-migrations/migrations.ts
ClFeSc: I'd keep the `null` because I want the person who writes a migration to explicitly say that some part of it is not required. I've changed the `?` to a `!` and then replaced this `typeof` check with `!== null` which also works.

backend/src/database/state-migrations/migrations.ts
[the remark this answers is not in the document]
Dassderdie: > Well, we also save the new state version. Maybe
> 

👍

backend/src/database/state-migrations/migrations.ts
```diff
@@ -136,71 +128,86 @@ export async function migrateInDatabaseTo(
...
+    };
+} {
+    const targetVersion = ExerciseState.currentStateVersion;
+    let newCurrentState = currentState;
+    for (let i = currentStateVersion + 1; i <= targetVersion; i++) {
+        const stateMigration = migrations[i]?.state;
+        if (typeof stateMigration === 'function') {
+            if (history)
+                history.initialState = stateMigration(history.initialState);
+            else newCurrentState = stateMigration(newCurrentState);
```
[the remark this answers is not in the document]
ClFeSc: I hope I've not missed something.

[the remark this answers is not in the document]
Dassderdie: > I hope I've not missed something.

These two points:
* Change this function to return void
* Add a comment explaining that the history and `currentState` get mutated as a side effect by this function.

I meant `applyMigrations()` with "this function".

````ts
/**
 * Migrates {@link currentState} and {@link history} to the newest version ({@link ExerciseState.currentStateVersion})
 * by mutating them
 */
function applyMigrations(
    currentStateVersion: number,
    currentState: Mutable<object>,
    history?: {
        initialState: Mutable<object>;
        actions: (Mutable<object> | null)[];
    }
): void {
````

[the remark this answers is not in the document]
ClFeSc: Oh, I just forgot that. Added it now. I had to return the new state version from `applyMigration` as I couldn't mutate the number directly (obviously).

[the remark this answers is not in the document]
Dassderdie: By contract, this version would always be `ExerciseState.currentStateVersion`.

But I'm fine with returning the new version separately.

Dassderdie (approved):
This was more of a refactoring than I thought (mainly because we made other improvements along the way).

I haven't reviewed the final state now in one go, but the commits looked good and the migrations seemed to have worked fine.

Good work.

ClFeSc:
I just noticed a problem with the approach of recalculating the `currentState` while working on the validation: We apply invalidated actions there. I don't know whether this could be harmful. I worst thing I currently can think of is crashing the HTTP request of importing from a file (without modifying the database itself, this can only happen for file imports), which is not that bad.

PR #528, opened 2022-10-01
Fix performance leak in show-more-component

Dassderdie:
Currently, every message with a body (e.g., when creating a new exercise) creates an interval that is never cleared. Not only can these add up over time, but they also trigger a ChangeDetectionCycle for Angular.

This PR
* clears the interval after the components destruction and
* disables polling by default

Background:

The `ShowMoreComponent` limits the message-body to a default height and provides the ability to expand it entirely when the user clicks on a button. To determine whether this button should be shown, we need the current height of the body. There is no way to observe this height directly. Therefore this component has multiple strategies to do this. The polling strategy would only be needed if the height changes after, e.g., an asynchronous request. A cleaner solution could be to use angulars changeDetection (via the `OnCheck`-hook?) or hide the button via CSS.

We currently only need this in the `MessageBodyComponent`, therefore this should be good enough.

### DIFF
--- a/frontend/src/app/feature/messages/show-more/app-show-more.component.ts
+++ b/frontend/src/app/feature/messages/show-more/app-show-more.component.ts
@@ -55,7 +55,7 @@ export class AppShowMoreComponent
      */
     @Input()
     heightChangeObservationStrategies: HeightChangeObservationStrategies = {
-        polling: 1000,
+        polling: false,
         resizeObserver: true,
         mutationObserver: true,
     };
@@ -74,6 +74,7 @@ export class AppShowMoreComponent
 
     private resizeObserver?: ResizeObserver;
     private mutationObserver?: MutationObserver;
+    private pollingIntervallRef?: ReturnType<typeof setInterval>;
 
     ngOnChanges() {
         this.updateState();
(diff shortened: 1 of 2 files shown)


frontend/src/app/feature/messages/show-more/app-show-more.component.ts
```diff
@@ -86,7 +87,7 @@ export class AppShowMoreComponent
         // observe the height of the content and update the showingMore-state whenever it changes
         // therefore we make use of up to three different strategies to detect changes in the scrollHeight
         if (this.heightChangeObservationStrategies.polling) {
-            setInterval(
+            this.pollingIntervallRef = setInterval(
                 () => this.updateState(),
                 this.heightChangeObservationStrategies.polling
             );
```
ClFeSc, line 93: Doesn't `setInterval` expect a number as its second argument? `this.heightChangeObservationStrategies.polling` is `false`, or not?

Dassderdie: It was `number | false` (and `0` would have created a performance problem).

I changed it to `boolean`.

ClFeSc: Not what I meant, but you're change solved my issue.

ClFeSc (approved):

Issue #530, opened 2022-10-03
There is no easy way to differentiate between elements


Dassderdie:
### Problem

Currently, there is no "nice" way to differentiate between e.g. a `Patient` and a `Material` object.
At the moment, we use two different workarounds for this:

#### Type guards (that could break easily) 
`shared\src\store\action-reducers\utils\calculate-treatments.ts`
```ts
function isPatient(
    element: Mutable<Material | Patient | Personnel>
): element is Mutable<Patient> {
    return (element as Patient).personalInformation !== undefined;
}
```

#### Pass a type with the element

In addition to the element itself, the respective `type` is manually propagated to other functions (much more verbose).

1. 
* "Sender": `frontend\src\app\pages\exercises\exercise\shared\exercise-map\feature-managers\element-manager.ts` - `getElementFromFeature()`
* Receiver: `frontend\src\app\pages\exercises\exercise\shared\exercise-map\feature-managers\transfer-point-feature-manager.ts` - `onFeatureDrop()`

2. 
`frontend\src\app\pages\exercises\exercise\shared\core\drag-element.service.ts` - `TransferTemplate`

##### [`instanceof`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof?retiredLocale=de)
* Sadly, this doesn't work in our case, as objects saved in the state do not have a `prototype`. Remember that our state is an immutable JSON object that can be sent over the network.

##### Add a `type` property to the models
* Each model gets a readonly `type` property (similar to the `type` property in an action like in `shared\src\store\action-reducers\material.ts` - `MoveMaterialAction`)
* This type should be plural (`personnel`, `patients`, ...) to be reusable as the key in the respective Maps in the `ExerciseState` (`shared\src\state.ts`).
* TypeScript has a [discriminated-union feature](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions), this enables easy differentiation between the objects later.

1. Add the `type` property to the following models (`shared\src\models`):
* [ ] Viewport
* [ ] Vehicle
* [ ] Personnel
* [ ] Material
* [ ] MapImage
* [ ] TransferPoint
* [ ] Hospital
* [ ] HospitalPatient
* [ ] AlarmGroup
* [ ] Client
* [ ] PatientCategory
* [ ] VehicleTemplate
* [ ] MapImageTemplate
* [ ] PersonnelTemplate

[...]

PR #536, opened 2022-10-05
TransferTargetDropdown doesn't show hospitals

Dassderdie:
The root problem is that the `droppedElement.type` is of type `any` in `onFeatureDrop()`.

This type should probably be improved sometime in the future. I add it as a comment to #530 .


ClFeSc (approved):
Confirmed the problem existed and confirmed it's fixed with these changes.

PR #537, opened 2022-10-05
Add a launch config for firefox

Dassderdie:
* Adds a launch config for firefox
* Removes the deprecated npm extension from the recommended extensions

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -9,7 +9,6 @@
         "editorconfig.editorconfig",
         "bierner.jsdoc-markdown-highlighting",
         "ms-vsliveshare.vsliveshare",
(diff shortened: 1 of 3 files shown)


lukasrad02:
I've added `pathMappings` to the `launch.json` in 64e7e00 to enable debugging of frontend as well as shared code. Without this change, only code from the frontend package could be debugged.

However, code from the shared package still needs a reload of the page and can't be debugged immediately after launch. In Chrome, this works fine. Also, you will receive some error regarding the path mappings messages in VS Code if breakpoints are set before launch, but the breakpoints will work fine (immediately in the frontend package, after one reload in the shared package).  
Unfortunately, I could not fix this so far...

lukasrad02 (approved):

PR #538, opened 2022-10-05
Add Arrive now functionality for transfers

lukasrad02:
Adds a button to make a transfer arrive immediately


ClFeSc (approved):
It seems to be working as expected. It also has the advantage that the transfer is aborted immediately and not in the next tick as with the previous method of setting the time to `0`. So as a trainer, you can end a transfer without starting the exercise.

lukasrad02:
The action is renamed now.

I've checked the import for all files from [here](https://github.com/hpi-sam/digital-fuesim-manv_test-scenarios) and also created an own export which definitely contains the old action name and worked, too

backend/src/database/state-migrations/10-rename-delete-transfer-action.ts
```diff
@@ -0,0 +1,27 @@
...
+                    targetTransferPointId: string;
+                };
+                const newAction: DeleteTransferAction = {
+                    type: '[Transfer] Delete transfer',
+                    elementId: oldAction.elementId,
+                    elementType: oldAction.elementType,
+                    targetTransferPointId: oldAction.targetTransferPointId,
+                };
+
+                actions[i] = newAction;
```
ClFeSc, line 22: I think you can just reset the name of the action without creating a new one.

lukasrad02: I did not do it, since the type is readonly, but I think simply not casting to Action should do the trick, thank you for the hint.

I'm going to take a look at this tomorrow

frontend/src/app/pages/exercises/exercise/shared/transfer-overview/transfer-time-input/transfer-time-input.component.ts
```diff
@@ -40,4 +40,13 @@ export class TransferTimeInputComponent {
             elementId: this.elementId,
         });
     }
+
+    public transferArriveNow() {
```
anonym-HPI, line 44: maybe make this name consistent to `letElementArrive` (maybe even using the same name - could be confusing?)?

backend/src/database/state-migrations/10-rename-delete-transfer-action.ts
Dassderdie: The actions are actually `Mutable<object>` - sadly such a type doesn't exist.

````ts
        for (const action of actions as (Mutable<Action> | null)[]) {
            if (action?.type === '[Transfer] delete transfer') {
                action.type = '[Transfer] Delete transfer';
            }
        }
````

anonym-HPI:
> Converted to draft since I'd like to change the type from `[Transfer] delete transfer` to `[Transfer] Delete transfer` (uppercase 'D') to match the names of the other actions. This will take some time since I have to write a migration, too

Maybe rename it from `Delete` to `Remove`, `Remove` is chosen in the other reducers, except `[MapImageTemplate] Delete mapImageTemplate`. If you do so, rename the class and action, too.
Or we could completely rename it to something more meaningful, like `[Transfer] Finish transfer` or so.

lukasrad02:
> Or we could completely rename it to something more meaningful, like `[Transfer] Finish transfer` or so.

This seems like a good idea to me. The name "delete" confused me in the first moment and I had to check whether it really does what I expect, but since the action existed before I started this PR, I did not want to change it first.

However, since there seems to be no reason to stick with "delete", I'll rename it to "finish"

anonym-HPI:
> > Or we could completely rename it to something more meaningful, like `[Transfer] Finish transfer` or so.
> 
> This seems like a good idea to me. The name "delete" confused me in the first moment and I had to check whether it really does what I expect, but since the action existed before I started this PR, I did not want to change it first.
> 
> However, since there seems to be no reason to stick with "delete", I'll rename it to "finish"

I created this action before, finish is more meaningful and also gives the word `Delete` free for later, if something like a cancel button is created (which also deletes the thing on transfer, at least when it is a fully loaded vehicle).

Dassderdie (approved):
LGTM

anonym-HPI (changes requested):
Also tested the UI, what about putting the `sofort` button on the right and maybe even a bit with a gap or so or even on the left have a gap?

ClFeSc:
> Also tested the UI, what about putting the `sofort` button on the right and maybe even a bit with a gap or so or even on the left have a gap?

My first thought was the same, but I think it fits better on the left, as its the most extreme case of reducing the time, and it's directly next to the `- 5 mins` button.

anonym-HPI:
> > Also tested the UI, what about putting the `sofort` button on the right and maybe even a bit with a gap or so or even on the left have a gap?
> 
> My first thought was the same, but I think it fits better on the left, as it's the most extreme case of reducing the time, and it's directly next to the `- 5 mins` button.

It can stay there, what do you think about the two screenshots?

anonym-HPI:
One argument to at least create a gap is that it makes it less likely to accidently press the `sofort` button when you wanted as a trainer, to just reduce the time from like 60 minutes to 20 minutes and clicking multiple times on the `-5 mins`

lukasrad02:
We've discussed the placement of the button in our group yesterday and placing it on the left seemed to be the best option ("now" = "- infinity minutes").

We also thought about a different style for the button, to prevent mis-clicks, but there was no design that wasn't distracting.

Regarding the screenshots, I'd rather do this in a separate PR to keep things simple

anonym-HPI (approved):

lukasrad02:
I've just noticed that #527 changed the behavior of the migration system. Merging the current state of the `dev`-Branch worked without problems, but I'm not so familiar with the migration system to know whether this branch needs to be changed to work with the new migration system.

@ClFeSc Could you please take a look at this and tell me if (and what) I have to change something here?

ClFeSc:
> I've just noticed that #527 changed the behavior of the migration system. Merging the current state of the `dev`-Branch worked without problems, but I'm not so familiar with the migration system to know whether this branch needs to be changed to work with the new migration system.
> 
> @ClFeSc Could you please take a look at this and tell me if (and what) I have to change something here?

In this case, nothing as we've only changed the `state` migrations and not the `action` ones. And for reference, the only difference to before there is that we now don't return a new `state` but rather modify the passed in one.

PR #541, opened 2022-10-05
Fix importing the same exercise twice does not work

### LINKED ISSUE #534: Importing the same exercise twice does not work
When importing the same file twice without reloading the page in between, the second import has no noticeable effect.

Nils1729:
Resolves #534


Nils1729:
See #546

PR #544, opened 2022-10-06
Add new image for big material and material

benn02:
There is no migration to the new image since it was optional in the issue

image


Dassderdie:
@christianzoellner Are you happy with the new image?

anonym-HPI:
I would, if it stays the same design, at least maybe remove the white lines (at least inside the material), as the grey/... already creates a contrast to the red one. Maybe leave the white around the cross.

Also the red could maybe replaced with blue or orange (or which color is the most found) or ...
@christianzoellner what do you think?

anonym-HPI:
something like this maybe (but better looking)?
grafik

benn02:
I added Christian to review this since his opinion on the images is important

christianzoellner:
> @christianzoellner Are you happy with the new image?

> I added Christian to review this since his opinion on the images is important

sorry for just noticing this discussion now. :-) Let me get to that.

My idea for a big material would be a metal box as used on the GW San vehicles, which is also typical for any kind of emergency material storage. In contrast, the normal materials should be recognizable as a kind of backpack. 
Thus, my ideal goals for recognizable big materials are that they should be a) slightly larger than the backback-style nomal material, b) be a box that is broad instead of tall (also requires changes of the normal materials) and c) have the handles on the sides instead of on top.  

Thus, I think both the committed proposal and @anonym-HPI's proposal are not 100% recognizable by our target audience. Also, I personally do not really like the blue or red frame design. Maybe someone has a better idea?

I would propose something like this (tough I am not able to draw it in vector graphics):
image

Since such boxes (at least in the medical context) are generally used for disaster management materials, the Icon may be the orange-blue desaster-management symbol instead of the red cross

benn02:
I now changed the images of both big material and material according to @christianzoellner s comment. The material is now a backpack and the big material a metal box.

anonym-HPI:
> I now changed the images of both big material and material according to @christianzoellner s comment. The material is now a backpack and the big material a metal box.

@christianzoellner is it allowed to use this logo?

@benn02 can you please add which license works for the materials (without any license notice nobody is allowed to use it or edit it). Mayve look at the CC licenses to choose one. It should be (for the best usage in this project) be editable, used freely (commercially dont know) and the Logo should have a notice about the rights and license from the institution (if it falls under this @christianzoellner ?)

shared/src/data/default-state/material-templates.ts
```diff
@@ -23,7 +23,7 @@ const bigMaterialTemplate = MaterialTemplate.create(
     {
         url: '/assets/big-material.svg',
         // Roughly double the area of the standard material
```
Dassderdie, line 25:
```suggestion
```

I believe this isn't true or relevant anymore...

christianzoellner:
> @christianzoellner is it allowed to use this logo?

If you mean the blue/orange circle/triangle: That the international symbol for civil protection (and often also for desasster management in general), so you can use it freely.

btw: I did not yet try it in the application, but from the images I really like the current backpack + metal box.

Dassderdie:
> @benn02 can you please add which license works for the materials (without any license notice, nobody is allowed to use it or edit it). Maybe look at the CC licenses to choose one. It should be (for the best usage in this project) editable, used freely (commercially, don't know) [...]


@benn02  If you add a license, this should be ready to be merged.

benn02:
> CC license
Since I created the Images myself and I pushed them to this repo, they fall under the same license as the repo.

Dassderdie (approved):

PR #545, opened 2022-10-06
Lock mapImages

### LINKED ISSUE #332: Lock MapImages
Currently, when a mapImage takes up the whole screen, no movement for trainers with the mouse is possible. There should be a way to still move, e.g. by locking the mapImage.

Dassderdie:
Fixes #332

### DIFF
--- a/backend/src/database/state-migrations/11-add-map-image-is-locked.ts
+++ b/backend/src/database/state-migrations/11-add-map-image-is-locked.ts
@@ -1,0 +1,22 @@
+import type { Action, UUID } from 'digital-fuesim-manv-shared';
+import type { Migration } from './migrations';
+
+export const addMapImageIsLocked11: Migration = {
+    actions: (_initialState, actions) => {
+        actions.forEach((action) => {
+            if ((action as Action | null)?.type === '[MapImage] Add MapImage') {
+                Object.assign((action as { mapImage: object }).mapImage, {
+                    isLocked: false,
+                });
+            }
+        });
+    },
+    state: (state) => {
+        const typedState = state as {
+            mapImages: { [key: UUID]: object };
+        };
+        Object.values(typedState.mapImages).forEach((mapImage) => {
+            Object.assign(mapImage, { isLocked: false });
+        });
+    },
+};
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.html
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.html
@@ -59,7 +59,7 @@
                     [ngModel]="mapImage.locked"
                     (ngModelChange)="lock($event)"
                 />
-                <label class="form-check-label">Position Sperren</label>
+                <label class="form-check-label">Position sperren</label>
             </div>
         </ng-container>
     </div>
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.html
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.html
@@ -56,8 +56,8 @@
                 <input
                     type="checkbox"
                     class="form-check-input"
-                    [ngModel]="mapImage.locked"
-                    (ngModelChange)="lock($event)"
+                    [ngModel]="mapImage.isLocked"
+                    (ngModelChange)="setLocked($event)"
                 />
                 <label class="form-check-label">Position sperren</label>
             </div>
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.html
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.html
@@ -52,6 +52,15 @@
                     [ngModelInput]="heightInput"
                 ></app-display-validation>
             </div>
+            <div class="form-check form-switch pb-3">
+                <input
+                    type="checkbox"
+                    class="form-check-input"
+                    [ngModel]="mapImage.locked"
+                    (ngModelChange)="lock($event)"
+                />
+                <label class="form-check-label">Position Sperren</label>
+            </div>
         </ng-container>
     </div>
 </ng-container>
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
@@ -55,7 +55,7 @@ export class MapImagePopupComponent implements PopupComponent, OnInit {
 
     public setLocked(newLocked: boolean) {
         this.apiService.proposeAction({
-            type: '[MapImage] Lock MapImage',
+            type: '[MapImage] Set Locked MapImage',
             mapImageId: this.mapImageId,
             newLocked,
         });
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
@@ -55,7 +55,7 @@ export class MapImagePopupComponent implements PopupComponent, OnInit {
 
     public setLocked(newLocked: boolean) {
         this.apiService.proposeAction({
-            type: '[MapImage] Set Locked MapImage',
+            type: '[MapImage] Set isLocked',
             mapImageId: this.mapImageId,
             newLocked,
         });
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
@@ -52,4 +52,12 @@ export class MapImagePopupComponent implements PopupComponent, OnInit {
             newHeight,
         });
     }
+
+    public lock(newLocked: boolean) {
+        this.apiService.proposeAction({
+            type: '[MapImage] Lock MapImage',
+            mapImageId: this.mapImageId,
+            newLocked,
+        });
+    }
 }
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
@@ -53,7 +53,7 @@ export class MapImagePopupComponent implements PopupComponent, OnInit {
         });
     }
 
-    public lock(newLocked: boolean) {
+    public setLocked(newLocked: boolean) {
         this.apiService.proposeAction({
             type: '[MapImage] Lock MapImage',
             mapImageId: this.mapImageId,
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/feature-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/feature-manager.ts
@@ -27,6 +27,11 @@ export interface FeatureManager<ElementFeature extends Feature<any>> {
     ) => void;
 
     /**
+     * @returns whether the feature can be translated
+     */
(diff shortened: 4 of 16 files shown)


Dassderdie:
Hint: if you want the to get the mapImage in the new `isFeatureTranslatable()` function you can use `const mapImage = this.getElementFromFeature(feature)`

ClFeSc (approved):

anonym-HPI (commented):
Tested it, works, what I am not sure about, is it possible to make the default be false and not have to add this function in only the feature managers where it should be overriden? Would this be possible @Dassderdie ?

Dassderdie:
> Tested it, works, what I am not sure about, is it possible to make the default be false and not have to add this function in only the feature managers where it should be overriden? Would this be possible @Dassderdie ?

We have ten FeatureManagers. Most inherit the default true function from `ElementFeatureManager` - the others don't have a `FeatureManager` class from which they inherit. The `FeatureManager` interface could specify the function as optional, but I would prefer to leave it mandatory for discoverability reasons.
This decision could be revisited when there are stronger cases for either implementation.

shared/src/models/map-image.ts
```diff
@@ -15,12 +15,16 @@ export class MapImage {
...
+    public readonly locked: boolean;
+
     /**
      * @deprecated Use {@link create} instead
      */
-    constructor(topLeft: Position, image: ImageProperties) {
+    constructor(topLeft: Position, image: ImageProperties, locked?: boolean) {
         this.position = topLeft;
         this.image = image;
+        this.locked = locked ?? false;
```
Dassderdie, line 27: Is there a special reason you have made it optional?
I believe it is rather untypical in the models to have a property optional with a default value.

Btw. you would probably rather do it like this:

```suggestion
    constructor(topLeft: Position, image: ImageProperties, locked = false) {
        this.position = topLeft;
        this.image = image;
        this.locked = locked;
```

shared/src/models/map-image.ts
```diff
@@ -15,12 +15,16 @@ export class MapImage {
     @Type(() => ImageProperties)
     public readonly image: ImageProperties;
 
+    @IsBoolean()
+    public readonly locked: boolean;
```
Dassderdie, line 19: mountain

I think you must write a migration for this as previous states didn't have this property, and it isn't optional.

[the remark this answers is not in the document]
Dassderdie: Currently, it is possible to change the position of the `mapImage`, because only the frontend enforces it.
Either 
1. enforce it in the `MoveMapImage` reducer by throwing a `ReducerError` or
2. explain this via a comment.

I would prefer 2. so e.g.:

```
/**
 * Whether the UI should prevent position changes of the mapImage.
 */
```

Dassderdie:
@christianzoellner Currently,
* participants can never move images,
* trainers can only move unlocked images and
* every trainer can lock and unlock any image via a switch in the image's popup




Is this the desired behavior, or should participants also be able to move unlocked images?

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/feature-manager.ts
```diff
@@ -26,6 +26,11 @@ export interface FeatureManager<ElementFeature extends Feature<any>> {
         feature: ElementFeature
     ) => void;
 
+    /**
+     * @returns whether the feature can be translated
+     */
```
anonym-HPI, line 31: This comment is actually not giving anymore information as is already provided by the name. I would propose to maybe write into the comment what translatable means for a feature (or link to a documentation explaining it) or just remove it all together (but I would propose either explain or to put a link in)?

ClFeSc (changes requested):
I completely missed the missing migration, thanks @Dassderdie! So no approval anymore from me.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
```diff
@@ -52,4 +52,12 @@ export class MapImagePopupComponent implements PopupComponent, OnInit {
             newHeight,
         });
     }
+
+    public lock(newLocked: boolean) {
```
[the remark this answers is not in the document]
anonym-HPI: I think `lock` is okay, but I would propose two different Action Reducers, one doing `Lock` and one doing `Unlock` instead of a `Lock` action that can also unlock or call this action something like `Toggle lock` (but the problem there is you can't enforce it to be locked, when you don't know the current state, don't know what happens if something gets out of sync or if this out of sync would even be a possibility).

anonym-HPI:
> > Tested it, works, what I am not sure about, is it possible to make the default be false and not have to add this function in only the feature managers where it should be overriden? Would this be possible @Dassderdie ?
> 
> We have ten FeatureManagers. Most inherit the default true function from `ElementFeatureManager` - the others don't have a `FeatureManager` class from which they inherit. The `FeatureManager` interface could specify the function as optional, but I would prefer to leave it mandatory for discoverability reasons. This decision could be revisited when there are stronger cases for either implementation.

Ok, as long as this is a thought decision and you think it brings value and not just increases the code size without a reason.

christianzoellner:
> * participants can never move images,
>  * trainers can only move unlocked images and
> * every trainer can lock and unlock any image via a switch in the image's popup
> 
> Is this the desired behavior, or should participants also be able to move unlocked


lgtm, this is what I would have specified, thanks @Nils1729 and @Dassderdie . :-) 
We need to test it with actual users and trainers soon.

One question: are images locked or unlocked by default?

anonym-HPI:
> > * participants can never move images,
> > * trainers can only move unlocked images and
> > * every trainer can lock and unlock any image via a switch in the image's popup
> > 
> > Is this the desired behavior, or should participants also be able to move unlocked
> 
> lgtm, this is what I would have specified, thanks @Nils1729 and @Dassderdie . :-) We need to test it with actual users and trainers soon.
> 
> One question: are images locked or unlocked by default?

The default is unlocked.

shared/src/models/map-image.ts
lukasrad02: We had the hope that making it optional would be an alternative to writing a migration, but since you've said that we need the migration anyways, we're going to change this.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.ts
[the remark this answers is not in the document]
lukasrad02: To avoid having two actions for this and the trouble a `toggle`-action might have, we've decided to use `Set Locked` as the name for the action ([1611dfa](https://github.com/hpi-sam/digital-fuesim-manv/pull/545/commits/1611dfa760d14b6c98785c5169c1e58f02b05e58)). Does this look good to you?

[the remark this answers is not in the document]
anonym-HPI: What about something like `Set isLocked` (the variable name) or `Set LockedStatus` or `LockStatus`?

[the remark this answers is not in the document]
anonym-HPI: > > To avoid having two actions for this and the trouble a `toggle`-action might have, we've decided to use `Set Locked` as the name for the action ([1611dfa](https://github.com/hpi-sam/digital-fuesim-manv/pull/545/commits/1611dfa760d14b6c98785c5169c1e58f02b05e58)). Does this look good to you?
> 
> I'm fine with it.
> 
> I assume @anonym-HPI wanted two separate actions because of [this](https://redux.js.org/style-guide/#model-actions-as-events-not-setters) best practice?
> Because this action only sets the `lock` property and is completely independent of other actions, I think your setter approach is better.
> 
> This should be refactored to `lock` and `unlock` actions, though, as soon as the reducers do more.

Didn't knew this style guide, but I am fine with one action, but I would change the action name (see my previous comment).

shared/src/models/map-image.ts
lukasrad02: There is a migration now ([8f2ce60](https://github.com/hpi-sam/digital-fuesim-manv/pull/545/commits/8f2ce60b9dc2e73956792294684b65a352cbbf4e))

shared/src/models/map-image.ts
lukasrad02: There is a migration now ([8f2ce60](https://github.com/hpi-sam/digital-fuesim-manv/pull/545/commits/8f2ce60b9dc2e73956792294684b65a352cbbf4e))

backend/src/database/state-migrations/11-add-map-image-is-locked.ts
```diff
@@ -0,0 +1,22 @@
+import type { Action, UUID } from 'digital-fuesim-manv-shared';
+import type { Migration } from './migrations';
+
+export const addMapImageIsLocked11: Migration = {
+    actions: (_initialState, actions) => {
+        actions.forEach((action) => {
+            if ((action as Action | null)?.type === '[MapImage] Add MapImage') {
+                Object.assign((action as { mapImage: object }).mapImage, {
+                    isLocked: false,
+                });
```
Dassderdie, line 10: Dust

I'm personally still of the opinion that typings in the migrations aren't useful.

The `Object.assign()` syntax is also a bit strange but does make it easier to not use `any` typings.

-> Personal preferences.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/feature-manager.ts
Dassderdie:
```suggestion
    /**
     * @returns whether the feature can be moved by the user
     */
```

Dassderdie (commented):
Everything LGTM.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/feature-manager.ts
```diff
@@ -26,6 +26,11 @@ export interface FeatureManager<ElementFeature extends Feature<any>> {
         feature: ElementFeature
     ) => void;
 
+    /**
+     * @returns whether the feature can be moved by the user
```
anonym-HPI, line 30:
```suggestion
     * @returns whether the feature can be moved by the user (when having succifient rights to do so)
```

Not sure about that (I know it is only visual and moving objects is controlled by the API), what do you think @Dassderdie ?

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.html
```diff
@@ -52,6 +52,15 @@ <h5 class="popover-header">
...
                 ></app-display-validation>
             </div>
+            <div class="form-check form-switch pb-3">
+                <input
+                    type="checkbox"
+                    class="form-check-input"
+                    [ngModel]="mapImage.isLocked"
+                    (ngModelChange)="setLocked($event)"
+                />
+                <label class="form-check-label">Position sperren</label>
```
anonym-HPI, line 62: Maybe put the label in front of the toggle? I am thinking of e.g. settings in iOS or Android (first the name and then the toggle).

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/feature-manager.ts
Dassderdie: This would be wrong.

This is a completely independent API from the actions.
E.g. a transferLine or the trash bin at the top of the screen are not even in the state and could be made user moveable by this...

frontend/src/app/pages/exercises/exercise/shared/exercise-map/shared/map-image-popup/map-image-popup.component.html
lukasrad02: We've used a layout consistent to the existing "Verbindungen anzeigen" toggle in the sidebar of the editor.

I don't have a preference whether the toggle is on the left or the right of the label, but feel free to open a separate issue to discuss this.

By the way: At the moment, labels and input element have no logical connection. I've created #552 for this.

anonym-HPI: Ok, doing this later and making it consistent in the UI.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/feature-manager.ts
anonym-HPI: Ok, resolving this

anonym-HPI (approved):

PR #546, opened 2022-10-06
Add directive for consistent file inputs across browsers

### LINKED ISSUE #534: Importing the same exercise twice does not work

Dassderdie:
This fixes #534. It uses the same workaround as #541 but encapsulates it in a reusable directive instead of directly doing it in the component.

### DIFF
--- a/frontend/src/app/shared/directives/file-input.directive.ts
+++ b/frontend/src/app/shared/directives/file-input.directive.ts
@@ -1,6 +1,4 @@
-import type { OnDestroy } from '@angular/core';
 import { Directive, EventEmitter, HostListener, Output } from '@angular/core';
-import { Subject } from 'rxjs';
 
 /**
  * This directive must only be used on file inputs and provides an output similar to `(input)`,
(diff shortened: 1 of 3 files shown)


frontend/src/app/shared/directives/file-input.directive.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @example
+ * ````html
+ * <input type="file" (appFileInput)="processFileList($event)" />
+ * ````
+ */
+@Directive({
+    selector: 'input[appFileInput]',
+})
+export class FileInputDirective implements OnDestroy {
+    destroy$ = new Subject<void>();
```
Nils1729, line 18: Dust
Is the `destroy$` member required here? I could not find any observers to it.

Dassderdie: Very good observation!
I'll look into it this evening.

Dassderdie: I removed the (unused) destroy logic in dced4d474d0dc8d697386b23f6c801266f56b7f9.

I believe it is safe to assume that `@HostListener` automatically unsubscribes and, while `appFileInput` doesn't `complete()` when the directive is destroyed, angular automatically unsubscribes from Outputs in templates.

So there should hopefully be no memory leak.

Nils1729 (approved):
LGTM

PR #548, opened 2022-10-06
new capacity model proposal

anonym-HPI:


anonym-HPI:
This is a first proposal, showing the direction this could go in.
Of course the individual numbers are up for debate and should be editable via templates.

- I would also remove the `canCaterFor` and put `capacity` as a variable directly into personnel and material.
- Also I would give every patient a cost instead of this `catersCost` variable in `calculate-treatment.ts`, this cost should change when `realStatus` changes (or could later change in each state, e.g. some `yellow` patient could need more capacity than normal)

Two examples:
- We have a `red` patient that would cost `1`, but a participant marks the patient as `green`, now a `san` can treat a `red` patient, which right now it shouldn't be able to do.
- Let's take the same example, but the `red` patient was not triaged (so it is `white`), the participant shouldn't be able to drag & drop a `san` around and see which needs to be at least a `yellow` patient as no `cateringLine` is there.

I would propose the following:
- Every personnel/material can treat at least one patient, even when it is over there limit. How good a patient can be treated by this personnel should be handled by the health/state calculation. Of course this only would be true if the capacity is `>0`, so a `gf` still can't treat a patient.
- The `visibleStatus` still influences the priority calculation, but not the actual cost calculation (`couldCaterFor`).

When a participant says a patient with a big wound is `green` and a personnel comes to this patient, this personnel would even with this (wrong) triage have no more capacity because of that.

This would mitigate the second example.
About the second example, this would probably be more realistic, as every personnel could at least try to treat a patient (how sucessful this treatment is would/should be determent in the health/state calculation).

Actually right now there is an exploit possible, if I am not mistaken (of course a trainer could intervene, but would have to see it):
- participants triages every patient as `green` but keep track which are actual `green`, `yellow` and `red`, e.g. with geographic distancing
- now every notarzt and other personnel has more capacity to treat `red` patients, as they are triaged as `green` and as such treated as `green` in the calculation (e.g. a `notarzt` could suddenly treat `6` `red` patients).

This would exploit would be not anymore possible when `couldCaterFor` takes the `realStatus` (or then the `catersCost`) of a patient into account.

@christianzoellner @Dassderdie @ClFeSc What do you think (of course also other persons can write what they think)?

PR #550, opened 2022-10-10
Preload tiles

Dassderdie:
[Demonstration](https://openlayers.org/en/latest/examples/preload.html)

The number of requests didn't seem to change by a lot. I didn't do a proper benchmark of it, though.

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
@@ -96,7 +96,9 @@ export class OlMapManager {
         transferLinesService: TransferLinesService
     ) {
         // Layers
-        const satelliteLayer = new TileLayer();
+        const satelliteLayer = new TileLayer({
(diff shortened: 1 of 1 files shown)


benn02 (commented):
Why is it necessary to preload the tiles and doesnt it produce an unnecessary load to preload infinetly many?

Dassderdie:
> Why is it necessary to preload the tiles 

Look at the demo https://openlayers.org/en/latest/examples/preload.html and try zooming in and out + pan to see the difference. Preloading makes this less intrusive as the new tiles are not white but a lower resolution tile from the appropiate area.
Because we cache the tiles (`cacheSize: 1000` a few lines below), this is in our case only noticeable on the first zoom/pan.

> and doesnt it produce an unnecessary load to preload infinetly many?

The argument is not how many tiles should be preloaded, but the maximum zoom for which they should be preloaded. We want to preload for every zoom level (-> zoom level `Infinity` like in the demo). In our case, the maximum zoom is 23 (same file, when the `OlMap` is initialized), so in this specific case we could use any number >= 23

benn02 (approved):

PR #553, opened 2022-10-10
Allow join links as exercise ids

lukasrad02:


lukasrad02:
It seems like the previously used `onlyNumbersValidator` is not used in any other component. Shall I remove it or keep it so it doesn't need to be reimplemented if we need it some day?

frontend/src/app/pages/landing-page/landing-page/landing-page.component.html
```diff
@@ -9,12 +9,11 @@ <h1 class="display-3 m-3 text-center">Digitale FüSim MANV</h1>
         <div class="input-group">
             <input
                 [(ngModel)]="exerciseId"
+                (ngModelChange)="joinInputChanged()"
```
Dassderdie, line 12: I believe this would break if you switch the `[(ngModel)]` and `(ngModelChange)`.

frontend/src/app/shared/validation/join-id-or-url-validator.directive.ts
```diff
@@ -0,0 +1,20 @@
...
+        {
+            provide: NG_VALIDATORS,
+            useExisting: JoinIdOrUrlDirective,
+            multi: true,
+        },
+    ],
+})
+export class JoinIdOrUrlDirective implements Validator {
+    validate(control: AbstractControl) {
+        return CustomValidators.joinUrlOrIdValidator()(control);
```
Dassderdie, line 18: All validators could probably use some refactoring to not recreate the validator from the factory on every key press...

frontend/src/app/pages/landing-page/landing-page/landing-page.component.html
```diff
@@ -9,12 +9,11 @@ <h1 class="display-3 m-3 text-center">Digitale FüSim MANV</h1>
...
                 [appAutofocus]="true"
                 (keydown.enter)="joinExercise()"
-                [minlength]="6"
-                [maxlength]="8"
                 required
-                appOnlyNumbersValidator
+                appJoinUrlOrIdValidator
```
Dassderdie, line 16: General feedback:

Instead of allowing both, the id or the url, I would just replace the url with the id on paste.
(e.g. https://stackoverflow.com/questions/50138910/angular-on-paste-event-get-content)

Nobody is going to type in the url letter by letter here. Actually, the only occurence where someone pasted the url in here, was during development. None of our customers has (to my knowledge) during our testruns made this mistake, yet.

Also, we only allow ids with the lengths of 6 or 8. So if you are already at it, you could also change this.

frontend/src/app/pages/landing-page/landing-page/landing-page.component.html
Dassderdie: Background:

```
[(ngModel)]="exerciseId"
``` 
is the shorthand syntax for 
```
[ngModel]="exerciseId"
(ngModelChange)="exerciseId = $event"
```

lukasrad02: Thank you for the warning. Since I'm using the paste event instead of the change event now, this is resolved.

frontend/src/app/pages/landing-page/landing-page/landing-page.component.html
lukasrad02: I've renamed it to `appJoinIdValidator` and only allow 6 or 8 digits now. I've also reintroduced `[minLength]` and `[maxLength]`, since this automatically ignores every input after the 8th char, but pasting longer URLs still works fine (due to handling the paste event).

Does this look good to you or shall we use only one length validation mechanism (i.e., removing `[minLength]` and `[maxLength]`)?

frontend/src/app/shared/validation/join-id-or-url-validator.directive.ts
lukasrad02: Since I'm not that familiar with Angular, I'm not sure what is the best way to solve this.

My idea was using the factory in the constructor and storing the validator function, so the code would look like this:

```ts
export class JoinIdDirective implements Validator {
    private readonly validator: (control: AbstractControl) => any;
    constructor() {
        this.validator = CustomValidators.joinIdValidator();
    }
    validate(control: AbstractControl) {
        return this.validator(control);
    }
}
```

This would still create one validator per input element, but at least not per key press. I think it should also work for validators with inputs (e.g. `exactMatchValidator`) by using `implements OnChanges`/`ngOnChanges` to (re-)create the validator.

If you think this is okay, I would change the `joinIdValidator` as stated above and refactor the other validators in a separate PR

Dassderdie:
> It seems like the previously used `onlyNumbersValidator` is not used in any other component. Shall I remove it or keep it so it doesn't need to be reimplemented if we need it some day?

I doubt you'll need it somewhere else. So remove it.

frontend/src/app/shared/validation/join-id-or-url-validator.directive.ts
Dassderdie: You could simplify this a bit:

```ts
export class JoinIdDirective implements Validator {
    private readonly validator = CustomValidators.joinIdValidator();

    validate(control: AbstractControl) {
        return this.validator(control);
    }
}
```

For the factories with parameters, `OnChanges` should work:

```ts
export class ExactMatchValidatorDirective implements Validator, OnChanges {
    /**
     * The string to match
     */
    @Input() appExactMatchValidator!: string;

    private validator!: ValidatorFn;

    validate(control: AbstractControl) {
        return this.validator(control);
    }

    ngOnChanges(changes: SimpleChangesGeneric<this>) {
        this.validator = CustomValidators.exactMatchValidator(
            this.appExactMatchValidator
        );
    }
}
```

One could even think about changing the `CustomValidators` from only Validator-factory-functions to Validators or Validator-factory-functions. But the performance benefit would be negligible, and the typings would become more complicated.

> If you think this is okay, I would change the joinIdValidator as stated above and refactor the other validators in a separate PR

👍

frontend/src/app/pages/landing-page/landing-page/landing-page.component.html
Dassderdie: Add a comment in the html (above the input) about why you still want the maxLength and minLength validators, and I'm happy with it.

frontend/src/app/shared/validation/display-validation/display-validation.component.html
```diff
@@ -51,4 +51,7 @@
             >) </ng-container
         >.
     </div>
+    <div *ngIf="errors['joinId']">
+        Die Eingabe muss eine gültige Übungs-ID (6-8 Ziffern) sein.
```
Dassderdie, line 55:
```suggestion
        Die Eingabe muss eine gültige Übungs-ID (6 oder 8 Ziffern) sein.
```

frontend/src/app/shared/validation/join-id-or-url-validator.directive.ts
lukasrad02: Created #559 for this. Will start working on it after this PR is merged to prevent conflicts due to the addition and deletion of validators.

The `appJoinIdValidator` is refactored now.

Dassderdie (approved):
Good job.

PR #556, opened 2022-10-10
Fix incorrect image url for patient templates

### LINKED ISSUE #555: Patients added to imported exercices are not visible
Steps to reproduce:
1. Import an exercise from the [test scenarios](https://github.com/hpi-sam/digital-fuesim-manv_test-scenarios)
2. Add a patient to the map

While patients created in the imported exercise are visible, new patients only show their pretriage indicator but not the patient image itself.

The following image shows one patient from the imported exercise (green indicator, lower left corner) and one new patient (white indicator, upper right corner):
image

This bug only occurs for exercises with state version 2 and 4 but not for those with state version 9 or 10 (most recent version as of creating this issue), so it might be related to migrations between.

ClFeSc:
Fixes #555 

Bumps the state version with a new migration instead of modifying the incorrect one to perform the update for already migrated exercises.

Should it also change the images of placed patients once again in order to show images for placed patients with the incorrect image?


lukasrad02 (approved):
LGTM

lukasrad02:
> Should it also change the images of placed patients once again in order to show images for placed patients with the incorrect image?

What do you mean? Do you want to change `patient.svg` back to `male-patient.svg` or do you want to apply the changes from migration 8 in your new migration again?

ClFeSc:
> > Should it also change the images of placed patients once again in order to show images for placed patients with the incorrect image?
> 
> What do you mean? Do you want to change `patient.svg` back to `male-patient.svg` or do you want to apply the changes from migration 8 in your new migration again?

I want to update all `male-patient.svg` to `patient.svg` so that patients added with the incorrect image are visible once again.

lukasrad02:
> > > Should it also change the images of placed patients once again in order to show images for placed patients with the incorrect image?
> > 
> > 
> > What do you mean? Do you want to change `patient.svg` back to `male-patient.svg` or do you want to apply the changes from migration 8 in your new migration again?
> 
> I want to update all `male-patient.svg` to `patient.svg` so that patients added with the incorrect image are visible once again.

Oh, yes, did not think about this. Please add this.

lukasrad02 (changes requested):
Since #545 has been merged, this migration has to be number 12 now to avoid conflicts

backend/src/database/state-migrations/migrations.ts
```diff
@@ -59,6 +60,7 @@ export const migrations: {
     8: treatmentSystemImprovements8,
     9: removeIsBeingTreated9,
     10: renameDeleteTransferAction10,
+    11: renameIncorrectImages11,
```
anonym-HPI, line 63:
```suggestion
    12: renameIncorrectPatientImages12,
```
(having 12 and maybe adding patient into the name or else make the name more clear, do we have a guideline for naming migrations?)

ClFeSc:
> > > > Should it also change the images of placed patients once again in order to show images for placed patients with the incorrect image?
> > > 
> > > 
> > > What do you mean? Do you want to change `patient.svg` back to `male-patient.svg` or do you want to apply the changes from migration 8 in your new migration again?
> > 
> > 
> > I want to update all `male-patient.svg` to `patient.svg` so that patients added with the incorrect image are visible once again.
> 
> Oh, yes, did not think about this. Please add this.

Done.

backend/src/database/state-migrations/12-rename-incorrect-images.ts
```diff
@@ -0,0 +1,51 @@
...
+            state as {
+                patientCategories: {
+                    patientTemplates: { image: { url: string } }[];
+                }[];
+            }
+        ).patientCategories.forEach((patientCategory) =>
+            patientCategory.patientTemplates.forEach((patientTemplate) => {
+                if (patientTemplate.image.url === '/assets/male-patient.svg') {
+                    patientTemplate.image.url = '/assets/patient.svg';
+                }
```
lukasrad02, line 28: This code does essentially the same as `migratePatient()`/`migrateImageProperties()`.

Do you think it would make sense to use `migratePatient()` here, too? The most clean way might be to introduce a function `migratePatientTemplate()` that also calls `migrateImageProperties()` instead of using `migratePatient()` for patient templates, but since the typings in the migrations aren't that strict, one function for both should be fine IMO.

anonym-HPI: This `migratePatient` function just seems to be not really necessary, couldn't you just call `migrateImapeProperties` directly (and doing the `patient.image` part in line 12)?

anonym-HPI (changes requested):

backend/src/database/state-migrations/migrations.ts
Dassderdie: Has been fixed (now also with the file-name)

backend/src/database/state-migrations/12-rename-incorrect-images.ts
Dassderdie: > This `migratePatient` function just seems to be not really necessary, couldn't you just call `migrateImapeProperties` directly (and doing the `patient.image` part in line 12)?

Done

PR #560, opened 2022-10-12
Do not recreate validators on input change

lukasrad02:


frontend/src/app/shared/validation/exact-match-validator.directive.ts
```diff
@@ -13,14 +14,21 @@ import { CustomValidators } from './custom-validators';
...
-export class ExactMatchValidatorDirective implements Validator {
+export class ExactMatchValidatorDirective implements Validator, OnChanges {
+    private validator!: ValidatorFn;
+
     /**
      * The string to match
      */
     @Input() appExactMatchValidator!: string;
 
+    ngOnChanges(changes: SimpleChanges): void {
```
Dassderdie, line 25: Dust

```suggestion
    ngOnChanges(changes: SimpleChangesGeneric<this>): void {
```

`SimpleChangesGeneric<this>` provides better (not perfect) typing support than `SimpleChanges`. For discoverability reasons, I would always use it, even when it isn't actually needed.

Dassderdie (approved):
I haven't tried it out, but the code lgtm.

PR #562, opened 2022-10-12
fix issue 561 + add ssl comment

anonym-HPI:


ClFeSc (approved):
I can't verify if this fixes any problems, but the changes in themselves look not too bad.

PR #549, opened 2022-10-07
Refactor exercise selectors

Dassderdie:

PR #564, opened 2022-10-13
Refactor ApiService and Store

### LINKED ISSUE #470: Do not create WebSocket connection on landing page
Currently, a WebSocket connection to the server is always created when a client is on the website.
Instead, it should only be initiated when joining an exercise and stopped when leaving it.

This should probably be done by treating the `presentExerciseHelper` like the `TimeTravelHelper` in  #447 (only create it when joining an exercise and destroy it when leaving).

Dassderdie:
This (very big :/) PR mostly removes technical debt that I created when I added the time-travel functionality during the span of a little more than a week.

## What does this PR do?

* Splits the "god class" `ApiService` up into different smaller services
* Makes it clear that the state in the store can either be the state of a live exercise or of the exercise in timeTravel
* Moves all "important" values (`ownClientId`, `ownClient`, `isTimeTraveling`, `isJoined`, `timeConstraints`, ...) from the `ApiService` to the store. 
    * This removes some race conditions (`isTimeTraveling$` emitted true, but the state in the store is still the live exercise), 
    * removes logic to provide all of these values as snapshots **and** as observables, and
    * enables their use in other selectors (= less code, improved API, probably also better performance).
    * this also incorporates https://github.com/hpi-sam/digital-fuesim-manv/pull/549#issue-1401718526
* Fixes #470
* The WebSocket connection is now disconnected during the timeTravel and reconnects when rejoining the exercise.
* Disables notifications for joined clients or arrived vehicles during timeTravel.

## Which code was changed?

It only changes code in the frontend. The biggest changes were made in:
* `ApiService`
* The whole Store 
    * especially state, actions, and reducers
    * The implementation of most selectors changed too. But not the values they ultimately return.
* `NotificationService` (got integrated in the new `ExerciseService`)
* `getStateSnapshot` (got replaced with `selectStateSnapshot`)

These changes propagated through most of the frontend code base. That's the reason this PR got so big. Don't get intimidated by it, most changes are just renames or different imports.

The commit history doesn't have any value.


frontend/src/app/core/application.service.ts
```diff
@@ -0,0 +1,81 @@
...
+            default:
+                assertExhaustiveness(currentExerciseStateMode);
+        }
+    }
+
+    public async joinExercise(
+        ...args: Parameters<typeof this.exerciseService.joinExercise>
+    ) {
+        this.stopCurrentMode();
+        return this.exerciseService.joinExercise(...args);
```
Dassderdie, line 53: During joining (async) and if `joinExercise` fails, the frontend is in an invalid state.
The same is true for `startTimeTravel`.

frontend/src/app/core/application.service.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public async leaveExercise() {
+        this.stopCurrentMode();
+    }
+}
+
+function assertExhaustiveness(variable: never) {
+    throw Error(`Unhandled case: ${variable}`);
+}
```
Dassderdie, line 81: Move this utility function to shared (with a comment on how to use it https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking)

lukasrad02:
@Dassderdie I've just noticed, that renaming the ngrx actions in `frontend/src/app/state/application/application.actions.ts` doesn't makes as much sense as we thought, since they're not functions wrapping the call to `createAction()` but rather variables saving the respective result.

Especially in the reducers, `on(createXYZAction, [...])` does not express the right thing. On the other hand, the old names were identically to the names of the function in `TimeTravelService` and `ExerciseService` where the actions are used, which might cause confusion.

Shall I undo the rename, keep it as it is or convert the variables to functions (so that the `create...` name makes sense, but we'll need to find some solution for referencing the actions in the `on()`-statements in the reducers)?

Dassderdie:
> @Dassderdie I've just noticed, that renaming the ngrx actions in `frontend/src/app/state/application/application.actions.ts` doesn't makes as much sense as we thought, since they're not functions wrapping the call to `createAction()` but rather variables saving the respective result.
> 
> Especially in the reducers, `on(createXYZAction, [...])` does not express the right thing. On the other hand, the old names were identically to the names of the function in `TimeTravelService` and `ExerciseService` where the actions are used, which might cause confusion.
> 
> Shall I undo the rename, keep it as it is or convert the variables to functions (so that the `create...` name makes sense, but we'll need to find some solution for referencing the actions in the `on()`-statements in the reducers)?

I see. Wouldn't the correct thing be just to remove the `create`-prefix? The `Action` suffix would distinguish them from the functions in the services.

Dassderdie:
> > @Dassderdie I've just noticed, that renaming the ngrx actions in `frontend/src/app/state/application/application.actions.ts` doesn't makes as much sense as we thought, since they're not functions wrapping the call to `createAction()` but rather variables saving the respective result.
> > Especially in the reducers, `on(createXYZAction, [...])` does not express the right thing. On the other hand, the old names were identically to the names of the function in `TimeTravelService` and `ExerciseService` where the actions are used, which might cause confusion.
> > Shall I undo the rename, keep it as it is or convert the variables to functions (so that the `create...` name makes sense, but we'll need to find some solution for referencing the actions in the `on()`-statements in the reducers)?

Actually, I think the `createAction()` method from ngrx is just badly named. It doesn't return an action, but
> [...] returns a function, that when called, returns an object in the shape of the [Action](https://ngrx.io/api/store/Action) interface.
https://ngrx.io/guide/store/actions#writing-actions

A more accurate name for `createAction()` would probably be `createActionFactory()` or something like this. I would therefore stay with the current naming. While it is a bit confusing in `application.actions.ts`, we correctly describe what is saved in the variable. Ngrx is at fault here IMAO.

PR #565, opened 2022-10-13
Add z-Index to mapImages

Nils1729:
TODO
* [x] Add rudimentary support for setting the zIndex of mapImages
* [ ] improve usability, e.g. by 'go back' and 'come forward' buttons 
* [ ] add migration

### DIFF
--- a/shared/src/models/map-image.ts
+++ b/shared/src/models/map-image.ts
@@ -1,5 +1,5 @@
 import { Type } from 'class-transformer';
-import { IsBoolean, IsUUID, ValidateNested } from 'class-validator';
+import { IsBoolean, IsInt, IsUUID, ValidateNested } from 'class-validator';
 import { uuid, UUID, uuidValidationOptions } from '../utils';
 import { Position, getCreate, ImageProperties } from './utils';
 
@@ -15,6 +15,9 @@ export class MapImage {
     @Type(() => ImageProperties)
     public readonly image: ImageProperties;
 
+    @IsInt()
+    public readonly zIndex: number;
+
     /**
      * Whether the UI should prevent position changes of the map image by drag&drop
      */
@@ -24,10 +27,16 @@ export class MapImage {
     /**
      * @deprecated Use {@link create} instead
      */
-    constructor(topLeft: Position, image: ImageProperties, isLocked: boolean) {
+    constructor(
+        topLeft: Position,
+        image: ImageProperties,
+        isLocked: boolean,
+        zIndex: number
+    ) {
         this.position = topLeft;
         this.image = image;
         this.isLocked = isLocked;
+        this.zIndex = zIndex;
     }
 
     static readonly create = getCreate(this);
--- a/shared/src/store/action-reducers/map-images.ts
+++ b/shared/src/store/action-reducers/map-images.ts
@@ -1,6 +1,7 @@
 import { Type } from 'class-transformer';
 import {
     IsBoolean,
+    IsInt,
     IsNumber,
     IsOptional,
     IsPositive,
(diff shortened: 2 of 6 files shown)


Nils1729:
I have encountered two difficulties in improving the usability of the zIndex. 
* I would like to assign unique and monotone `zIndex` values to new `MapImages`. For that I need to know the highest `zIndex` that is currently in use, but I do not see how to get that from the `DragElementService`.
* Similar tools offer buttons that raise/lower features just enough to have a visible effect, e.g. raise just above the highest overlapping feature. I have not found a simple way to find features overlapping a given feature yet.

For now, I will continue with simply setting the `zIndex` in the UI.

Dassderdie:
> I have encountered two difficulties in improving the usability of the zIndex. 
> * I would like to assign unique and monotone `zIndex` values to new `MapImages`. For that I need to know the highest `zIndex` that is currently in use, but I do not see how to get that from the `DragElementService`.
> * Similar tools offer buttons that raise/lower features just enough to have a visible effect, e.g. raise just above the highest overlapping feature. I have not found a simple way to find features overlapping a given feature yet.
> 
> For now, I will continue with simply setting the `zIndex` in the UI.

1. In the "Add MapImage" reducer you have access to all mapImages that are currently in the exercise. E.g. if the z-index from the create action is undefined, you could replace it with the an z-index that's higher than all other ones in the exercise (or something similar - not sure whether this feature would be a good UX though).
2. I believe this is non trivial. One could either do this in the reducer (best practice for the synchronization) or openLayers (you  could calculate it based on the actual rendered image and not just the position and outer dimensions of it). For both options the cost/benefit is probably very low.

Nils1729:
>  1. In the "Add MapImage" reducer you have access to all mapImages that are currently in the exercise. E.g. if the z-index from the create action is undefined, you could replace it with the an z-index that's higher than all other ones in the exercise (or something similar - not sure whether this feature would be a good UX though).

I already tried this and it does the thing. To me however it seems odd to save the action with a placeholder value  for the `zIndex` as thus not all info about the action is contained in the action object.

>  2. I believe this is non trivial. One could either do this in the reducer (best practice for the synchronization) or openLayers (you  could calculate it based on the actual rendered image and not just the position and outer dimensions of it). For both options the cost/benefit is probably very low.

My idea for that was to create some actions that only know the UUID of the MapImage to be moved and the one that it should be on top or below of. This way we can calculate in the frontend which MapImages overlap. The reducer would then shift all MapImages with a zIndex inbetween by one.
I am not entirely sure yet if it is cleanly feasible to find overlapping MapImages.

Dassderdie:
> I already tried this and it does the thing. To me however it seems odd to save the action with a placeholder value  for the `zIndex` as thus not all info about the action is contained in the action object.

An action doesn't have (or even strive) to contain all of the information. The combination of action and current state is of the essence.
In this regard we (try to) follow the Redux Style guide:
https://redux.js.org/style-guide/#put-as-much-logic-as-possible-in-reducers

( Even though we don't use redux directly)

> My idea for that was to create some actions that only know the UUID of the MapImage to be moved and the one that it should be on top or below of. This way we can calculate in the frontend which MapImages overlap. The reducer would then shift all MapImages with a zIndex inbetween by one.

This would be a (valid) mix between saving the information in the action an calculating it in the reducer.

> I am not entirely sure yet if it is cleanly feasible to find overlapping MapImages.

I doubt it :/

As I wrote, the cost/benefit for this feature is probably not very good. Maybe just do a "Send to back" and "Bring to front"...

Dassderdie (changes requested):

shared/src/models/map-image.ts
```diff
@@ -15,6 +15,15 @@ export class MapImage {
     @Type(() => ImageProperties)
     public readonly image: ImageProperties;
 
+    /**
+     * Determines the rendering order among other mapImages.
+     * A smaller number means the mapImage is behind another one.
+     * The index can also be negative.
+     * MapImages with the same zIndex don't have a defined order.
```
Dassderdie, line 22: But OpenLayers seems to order them deterministically. At least I couldn't produce a state where one client displayed something different than another.

shared/src/store/action-reducers/map-images.ts
```diff
@@ -86,11 +93,44 @@ export class ReconfigureMapImageUrlAction implements Action {
...
+}
+
+export class SendToBackMapImageAction implements Action {
+    @IsString()
+    public readonly type = '[MapImage] Send to back';
+
+    @IsUUID(4, uuidValidationOptions)
+    public readonly mapImageId!: UUID;
+}
+
```
Dassderdie, line 126:
```suggestion
```

ClFeSc (approved):
I looked at the code and tested it, and all seems good.

Dassderdie (approved):

PR #566, opened 2022-10-13
Improve OpenLayers performance

Dassderdie:
### Test setup

Chrome@Version 106.0.5249.103 (Official Build) (64-bit)

* Load [exercise-state-2022-05-31-workshop-babz-grßes-szenario.json](https://github.com/hpi-sam/digital-fuesim-manv_test-scenarios/blob/main/exercise-state-2022-05-31-workshop-babz-gro%C3%9Fes-szenario.json)
* Pan a bit around and move the mouse a lot

### Before any changes

image
image

* Every `pointermove` event takes ca. 50ms
* Most of this is spent by `getImageData` calls, triggered by 
     * our own callback to set the correct pointer style and
     * `Translate.handleMoveEvent`
* The `pointermove` event is fired without breaks in between

### After c141f7a1c17da3ccbbdcb2a4b439dfb57be5aa76

image
image

* There are now fewer expensive `pointermove` events because they are now throttled to be called at most every 200ms (`Timer Fired`). 
* The `Translate.handleMoveEvent` has been eliminated. Therefore all `pointermove` are a bit cheaper.

### After cdc34b721123f5cee3e5b426bb1a90146d5d8bfc
image

* One can see that there are no blue `Zone` function calls in `pointermove`. Because `throttle` uses `setTimeout` (which is still patched by zone.js) these still run in the zone.
* While the impact of the zone.js calls is not measureable, they can also trigger angular change-detection, which currently takes around 0.4 ms. I cannot quantify whether or how often this change prevents the change-detection from running.

## What should be merged?

Not all of this PR must be merged. It would also suffice to just merge some of the changes:

1. Overwrite the `Translate.handleMoveEvent`
2. Eliminate `pointermove` events when using a mouse (on touch screens we don't set the cursor anymore) by
    1. throttling them
    2. not changing the cursor whatsoever (ignore all `pointermove` events)
3. Do not patch `wheel`, `pointermove`, and `requestAnimationFrame`


Dassderdie:
I would be for 1., 2. ii., and 3..

Dassderdie:
Waiting for a response in https://github.com/openlayers/openlayers/issues/14175 could also be interesting.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/translate-interaction.ts
```diff
@@ -0,0 +1,15 @@
+import type { MapBrowserEvent } from 'ol';
+import { Translate } from 'ol/interaction';
+
+/**
+ * A more performant implementation for OpenLayers Translate interaction
+ */
+export class TranslateInteraction extends Translate {
```
Dassderdie, line 7: Maybe the TranslateHelper could get directly integrated into this class.

Dassderdie:
Closed in favor of #575 

The zone.js stuff is not so relevant, I think.

PR #569, opened 2022-11-09
Fix security vulnerabilities

ClFeSc:
* Bumping `socket.io-parser` using `npm audit fix`
* Bumping `loader-utils` using `npm audit fix`

Supersedes #568


Dassderdie (approved):
I haven't tried it out, but the code changes lgtm

PR #586, opened 2022-12-21
Feature/improve popup trigger performance

Dassderdie:
* Updates all packages
    * Only version bumps on minor and patch level
    * OpenLayers now sets `willReadFrequently` on the canvas rendering contexts (https://github.com/openlayers/openlayers/releases/tag/v7.2.0). A quick test showed this improves the performance from ca. 150ms per click to ca. 120ms. There was no difference when the map was not moved in between clicks.

* Improvements in `hasFeatureAtPixel` in `PopupManager`
    * Early return in `hasFeatureAtPixel`
    * Skip not relevant layers
    * This improved the performance from ca. 120 ms to ca. 35ms

While the used workaround is still very ugly, it is now at least less expensive and reduces the stutter when opening a popup.

The creation and rendering of the HTML popup were not included in the performance measurements above. They are only the time it took the JS from the registration of the click to the trigger of the rendering in the DOM.


frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/popup-helper.ts
```diff
@@ -102,15 +106,22 @@ export class ImagePopupHelper {
...
      */
     private hasFeatureAtPixel(feature: Feature<Point>, pixel: Pixel): boolean {
-        return this.olMap
-            .getFeaturesAtPixel(pixel, {
-                // We assume the feature has no holes and is on the two axis the highest
-                hitTolerance: 0,
-            })
-            .some((_feature) => _feature.getId() === feature.getId());
+        return (
+            this.olMap.forEachFeatureAtPixel(
```
Dassderdie, line 114: From the documentation of this function:
````
@return
Callback result, i.e. the return value of last callback execution, or the first truthy callback return value.
````

File: frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/popup-helper.ts [hunk 2]
```diff
@@ -102,15 +106,22 @@ export class ImagePopupHelper {
     }
 
     /**
+     * @param layer The layer the feature is in
```
Dassderdie, line 109:
```suggestion
```

ClFeSc (approved):
I haven't checked the performance, but everything works without stuttering. Apart from the one comment, you made yourself, I'm happy with it.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/popup-helper.ts
ClFeSc: Maybe add this as a short comment?

Dassderdie: I thought about it, but it should be enough to have this documented on the `forEachFeatureAtPixel` function in ol (as a jsDoc comment -> hovering over it is enough).
If we duplicate this documentation, this could get out of sync.

I added the comment in GitHub only because you can't see the documentation by hovering over the function in GitHub.

Dassderdie:
@ClFeSc FYI, I updated OpenLayers once again from `7.2.0` to `7.2.2`. The only things that changed were in some JSdoc custom scripts. I believe to fix the commonjs build. I couldn't notice any difference but wanted to make sure that we don't run into some weird problem later.

PR #591, opened 2023-01-07
Feature/improve eslint rules

Dassderdie:
* You can review this PR by looking at the commits.
* Only small changes. I mostly made some incremental improvements to the eslint rules and removed outdated/unnecessary rules.
* I also removed outdated and incorrect documentation as well as some smaller additions (TODO-comments and test-scenario repository)


README.md
```diff
@@ -109,24 +108,25 @@ There are already the following [debug configurations](https://code.visualstudio
 -   `Launch Frontend [Firefox]` (You have to install an extra extension)
 -   `Debug Jest Tests`
 
-In addition you can make use of the following browser extensions:
+In addition, you can make use of the following browser extensions:
 
 -   [Angular DevTools](https://chrome.google.com/webstore/detail/angular-devtools/ienfalfjdbdpebioblfackkekamfmbnh)
--   [Redux DevTools Extension](https://github.com/zalmoxisus/redux-devtools-extension/) for [NgRx](https://ngrx.io/guide/store-devtools)
```
Dassderdie, line 115: This would need additional code in the frontend to register it. I also never needed it.

README.md
```diff
@@ -257,16 +262,14 @@ A consequence of the synchronization strategy described before is that it takes
...
-In our case the [optimisticActionHandler](./frontend/src/app/core/optimistic-action-handler.ts) encapsulates this functionality.
+If the server rejects the proposal or a race condition occurs, the client corrects its state again. In our case, the [optimisticActionHandler](./frontend/src/app/core/optimistic-action-handler.ts) encapsulates this functionality.
 
 The state in the frontend is not guaranteed to be correct. It is only guaranteed to automatically correct itself.
 
 If you need to read from the state to change it, you should do this inside the action reducer because the `currentState` passed into a reducer is always guaranteed to be correct.
 
 ### Performance considerations
 
--   Do _not_ save a very large JS primitve (a large string like a base64 encoded image) in a part of the state that is often modified (like the root). This primitive would be copied on each change. Instead, the primitive should be saved as part of a separate object. This makes use of the performance benefits of shallow copies.
```
ClFeSc, line 269: Is this no longer a problem?

frontend/src/app/core/time-travel-helper.ts
```diff
@@ -38,9 +38,6 @@ export class TimeTravelHelper {
      * @param exerciseTime The time to travel to, if it isn't in the timeConstraints, it will be clamped appropriately
      */
     public async jumpToTime(exerciseTime: number): Promise<void> {
-        if (!this.timeConstraints || !this.timeJumpHelper) {
-            throw new Error('Start the time travel before jumping to a time!');
-        }
```
ClFeSc, line 43: Why have you removed this check? Is it always `false`?

ClFeSc (commented):
Just some small things. Looks good!

frontend/src/app/core/time-travel-helper.ts
Dassderdie: Yes it is always false.

README.md
Dassderdie: Actually, I don't know. It could be that there are (or will be) performance optimizations in V8 regarding the copying of very big primitives in the same thread.

But this is a very unique usecase and now, with the benchmarks, a performance Regression from something like this would be easy to spot. In addition it is easy to fix without big architectural changes (just wrap it in an object).

ClFeSc (approved):

Issue #374, opened 2022-05-12
Refactor image heights to meters


Dassderdie:
Currently, it is the number of pixels at zoom 23.

Check whether this is independent of the choosen map.

Greenscreen23:
When implementing, one should take into account that this magic number now appears twice in the code. Once for viewports and once for simulated regions

Dassderdie:
> When implementing, one should take into account that this magic number now appears twice in the code. Once for viewports and once for simulated regions

There are even more than just these two magic numbers in other feature managers. They are not all the same though.

PR #601, opened 2023-01-17
Add simulated regions that can be placed on the map

### LINKED ISSUE #596: Regions can be placed on the map
Similar to current viewports, new regions can be placed on the map. These will later contain autonomous functionality.

Greenscreen23:
Closes #596


frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/simulated-region-modify.ts
```diff
@@ -0,0 +1,69 @@
...
+                modifyGeometry.geometry.scale(
+                    (mouseCoordinate[0]! - origin[0]!) /
+                        (corner[0]! - origin[0]!),
+                    (origin[1]! - mouseCoordinate[1]!) /
+                        (origin[1]! - corner[1]!),
+                    origin
+                );
+            });
+            // This renders the default LineString style. The function ignores all arguments.
+            return defaultStyle!(feature, 123);
```
benn02, line 66: If the function ignores all arguments, why do we provide any?

lukasrad02:
What do you think about changing the SVG to have an slightly opaque background? This would make is easier to recognize simulated regions and show that the map visible behind is not that relevant, since the patients, materials etc. in the simulated regions are not visible at all.

Greenscreen23:
> What do you think about changing the SVG to have an slightly opaque background? This would make is easier to recognize simulated regions and show that the map visible behind is not that relevant, since the patients, materials etc. in the simulated regions are not visible at all.

The problem with an opaque background is that we would have to implement it in ol. I don't know how to do that.

frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
```diff
@@ -228,6 +229,29 @@ export class DragElementService {
                     true
                 );
                 break;
+            case 'simulatedRegion': {
+                // This ratio has been determined by trial and error
+                const height = SimulatedRegion.image.height / 23.5;
```
benn02, line 234: In SWT we would say magic number and anyway a single magic number with a comment is a big code smell

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.html
```diff
@@ -18,6 +18,31 @@ <h5>Ansicht/Transferpunkt</h5>
...
+                            </div>
+                            <div class="card-body">
+                                <h5
+                                    class="card-title text-truncate"
+                                    [title]="'Simulierter Bereich'"
+                                >
+                                    Simulierter Bereich
+                                </h5>
+                            </div>
+                        </div>
```
lukasrad02, line 45: Not a big thing, but do we want to place the simulated regions in front of all existing features (viewport and transfer point)? Especially since the simulated regions are not ready to use yet.

Instead of

1. placing it behind the viewport and transfer point

we could also

2. introduce a new headline for this (would make sense especially if we'd like to have the different types (Patientenablage, B-Raum) as separate elements in the editor)
3. add tabs to the editor so that there is one "page" with all features needed for the existing concept and one "page" with all features needed for the new concept (there would be some overlap, for example transfer points and patients) but there are a lot of features which are not needed in one of the different modes.

If we choose the third option, I'd suggest to do this in another issue.

What do you think about this?

shared/src/models/simulated-region.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    static isInSimulatedRegion(
+        region: SimulatedRegion,
+        position: Position
+    ): boolean {
+        return (
+            region.position.x <= position.x &&
+            position.x <= region.position.x + region.size.width &&
+            region.position.y - region.size.height <= position.y &&
+            position.y <= region.position.y
```
Nils1729, line 51: Please stub this logic to throw an error or something so we are not confused when we implement the new semantics later.

frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
Dassderdie: #374 would hopefully encapsulate this number. At the moment this is some sort of conversion factor from the placeholder image height to the image height on the canvas. I could imagine that this conversion is actually not linear...

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
```diff
@@ -161,19 +166,33 @@ export class OlMapManager {
...
         const customInteractions =
             selectStateSnapshot(selectCurrentRole, this.store) === 'trainer'
-                ? [...alwaysInteractions, viewportTranslate, viewportModify]
+                ? [
+                      ...alwaysInteractions,
+                      viewportTranslate,
+                      viewportModify,
+                      simulatedRegionTranslate,
+                      simulatedRegionModify,
+                  ]
```
Dassderdie, line 195: You should be able to reuse the `viewportTranslate` interaction by just adding `simulatedRegionLayer` to the layers array of the translate interaction. (not tested).

Dassderdie: If this works, you could rename `createViewportModify()` to `createRectangularModify()`

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/simulated-region-modify.ts
Dassderdie: Because the typings of OpenLayers are incorrect here and require an unused second argument. I agree that the comment should make this clearer. Maybe a `// @ts-expect-error` could be used?

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/simulated-region-modify.ts
```diff
@@ -0,0 +1,69 @@
+import type { Feature } from 'ol';
```
Dassderdie, line 1: This whole file should be unnecessary as you could just use `viewport-modify.ts`

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/simulated-region-modify.ts
lukasrad02: I'd also though about using `// @ts-expect-error`, but what I don't like about this solution is that (as far as I know) we cannot specify *what* error to expect, so possible future errors or a change in the type definition which might resolve this issue would not be detected/reported by TypeScript.

Therefore, I'd rather just make the comment more clear than using `// @ts-expect-error`.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/simulated-region-modify.ts
lukasrad02: For now it would be possible to use `ViewportModify`, but this is only made for `LineString` features. We already ran into the issue that we cannot specify a background color, so we want to switch to a `Polygon` feature later (see #603).

At the moment, we're not sure what consequences this change might have. So I'd either include the switch to `Polygon` in this PR (which would take some time and there are some features waiting for this PR to get finished) or keep the two implementations and do the refactoring when we're adding support for `Polygon`s, which requires changes to multiple existing implementations anyway.

Dassderdie: You reuse the `ViewportModify` with `LineStrings` in this PR and refactor this Modify later to a `Polygon` (the viewports would also use a Polygon  - but without a `fill`-style).

Also YAGNI.

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.html
Greenscreen23: I've implemented 2.

shared/src/models/simulated-region.ts
Greenscreen23: Done!

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
Greenscreen23: I like your idea, I was also thinking about that  when I saw your PR :)

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/simulated-region-modify.ts
Greenscreen23: I've updated the comment

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/simulated-region-modify.ts
Greenscreen23: I've merged the two files into one

lukasrad02 (approved):
This looks good to me now

benn02 (approved):
I think this is fine

Nils1729 (approved):

Issue #603, opened 2023-01-17
Use opaque fill style for simulated regions


Nils1729:
This requires the Geometry to be changed from `LineString` to `Polygon`.

Consequently, we need to touch a lot of the OpenLayers wrappers (`ElementManager`, `FeatureManager`, ...)

Dassderdie:
The current implementation for the viewports uses `LineString`s. This does not only mean that there is no fill-color available, but you are also only able to drop elements on the border of your `simulatedRegions`, and clicks are only registered on the borders too.

The best solution for this would be to refactor the `viewports` and `simulatedRegions` from `LineString` to `Polygon`.
`viewports` would just be `Polygons` without the `fill`-attribute.

You can see a rough prototypical sketch of the necessary refactoring in e59939cb623799cf90dc1417f4b88a7f74e61f78

* Most utilities must support `Point`, `LineString` and `Polygon`
* In the `ModifyHelper`. `viewport-modify` and `translate-interaction` the `Polygon` should just be able to replace the `LineString`

Greenscreen23:
> The current implementation for the viewports uses LineStrings. This does not only mean that there is no fill-color available, but you are also only able to drop elements on the border of your simulatedRegions, and clicks are only registered on the borders too.

I think this is a huge deal for us, so we will probably have to do this at some point. I will take a look at this, but we'll merge this after #604

Greenscreen23:
> You can see a rough prototypical sketch of the necessary refactoring in https://github.com/hpi-sam/digital-fuesim-manv/commit/e59939cb623799cf90dc1417f4b88a7f74e61f78

What was your reasoning for removing catering and transfer lines? Just for the ease of it, or is there a problem that some layers can't be line layers while others are vector layers that I'm missing?

Dassderdie:
> > You can see a rough prototypical sketch of the necessary refactoring in https://github.com/hpi-sam/digital-fuesim-manv/commit/e59939cb623799cf90dc1417f4b88a7f74e61f78
> 
> What was your reasoning for removing catering and transfer lines? Just for the ease of it, or is there a problem that some layers can't be line layers while others are vector layers that I'm missing?

Just for the ease of it. I wanted to see whether the Polygon without a fill would have the correct hitbox.

PR #604, opened 2023-01-17
Refactor viewport-modify to resize-rectangle-interaction

Dassderdie:
This PR replaces the very hacky implementation to resize the viewport with a new one.

* This can be merged after #601 
* Instead of using some parts of the `Modify` interaction (side-effect in the style function) a new custom Interaction has been created.
* Previously, one could switch between resizing and translating a viewport via pressing `Shift`. Instead, now one ***resizes*** if one drags an edge of the viewport (+ a bit around it), and ***translates*** if one drags any other part. This should enable touch users to resize viewports (I haven't tested it yet).
* Regression: Currently, there is no indicator that one can resize or is currently resizing a viewport. Previously this was done via the blue circle of the Modify interaction. If wanted, this could later be added either by modifying the style of the viewport or adding an overlay to the `ResizeRectangleInteraction`.
* The code should now be more readable and extendable.
* The names have been modified so that this interaction is now not specific for a viewport and easier to reuse.
* The unnecessary `viewportTranslateInteraction` has been removed
* `feature.dispose()` now gets called when it gets deleted - this could have been a memory leak?


frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/viewport-feature-manager.ts
```diff
@@ -61,15 +61,9 @@ export class ViewportFeatureManager
...
-    private readonly modifyHelper = new ModifyHelper();
 
     private readonly style = new Style({
-        geometry(thisFeature) {
-            const modifyGeometry = thisFeature.get('modifyGeometry');
-            return modifyGeometry
-                ? modifyGeometry.geometry
-                : thisFeature.getGeometry();
-        },
+        fill: undefined,
```
Dassderdie, line 66:
```suggestion
```

I worked on a branche where #603 was already partially implemented -> this should not be changed, yet

Greenscreen23:
We copied your code for `viewports` to `simulatedRegions` because of several reasons. Do you mind applying this to `simulatedRegions` as well when they are merged?

Dassderdie:
> We copied your code for `viewports` to `simulatedRegions` because of several reasons. 

Do you mean #601 (as of yesterday)? Or are you currently copying something from the viewports of this PR?

> Do you mind applying this to `simulatedRegions` as well when they are merged?

This was the plan.

Greenscreen23:
> Do you mean https://github.com/hpi-sam/digital-fuesim-manv/pull/601 (as of yesterday)? Or are you currently copying something from the viewports of this PR?
As of yesterday, I've deduplicated it today.

Greenscreen23 (approved):
It integrates properly with our setup and works as expected. I've taken a look over the code, but I'm trusting you that this code will work fine with ol, as I don't think I'm experienced enough to understand ol.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/resize-rectangle-interaction.ts
```diff
@@ -0,0 +1,156 @@
...
+    originCorner: Coordinate;
+    /**
+     * The feature that is currently being resized.
+     */
+    feature: Feature<LineString>;
+    currentScale: { x: number; y: number };
+}
+
+// TODO: This doesn't work as a static member of ResizeEvent, because it is undefined at runtime. Why?
+const resizeRectangleEventType = 'resizerectangle';
```
ClFeSc, line 140: No idea. My only guess would be that something with a conflicting name already exists in the type hierarchy.

ClFeSc (approved):
I haven't tested it, but the code looks good. Much cleaner than before.

ClFeSc:
@Dassderdie Regarding the question of what we used last year, it might have been this article: <https://openlayers.org/en/latest/examples/modify-scale-and-rotate.html>.

Greenscreen23:
:shipit:

PR #606, opened 2023-01-18
Ignore vs code config files

lukasrad02:
This PR makes it easier to adjust you local development setup by introducing the following changes:

- Untracking and ignoring the VS Code config files `launch.json`, `settings.json`, and `tasks.json` so that one does not accidentally commit their custom config
- Adding tracked versions of the files above, named `*.example` that can be used as a starting point/template for your custom config
- Adding a new task "Start all but database" for local development setup without a database
- Explaining the new files and task in the readme


benn02 (approved):
I think this looks good

Issue #605, opened 2023-01-18
Close popups with ESC


Nils1729:
For ease of development, it would be nice if all popups on the map and elsewhere could be closed by pressing ESCAPE.

lukasrad02:
To implement this feature, we could add an interaction like in [this SO post](https://stackoverflow.com/a/69392146) to the OL map manager (`frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts`). But this catches the keyboard events only if no textbox in the currently open popup has keyboard focus (although the `Escape` key has no effect on the textbox).

So if we'd like to be able to use the Escape key always, we have to add events handlers to all input elements in the popups, which would be a lot of work. We've got to decide what exactly is the desired behavior.

Dassderdie:
Try this:

in `ol-map-manager.ts` at the end of `registerPopupTriggers()`:

```ts
        this.openLayersContainer.addEventListener('keydown', (event) => {
            if ((event as KeyboardEvent).key === 'Escape') {
                this.changePopup$.next(undefined);
            }
        });
```

I don't think we need OpenLayers here and can just use the HTML events.
I'm not sure whether `event.preventDefault()` and/or `event.stopPropagation()` would be good here...

Dassderdie:
Be aware that this doesn't save the last input if you closed the modal too soon (600ms) after the last keypress in an input element.
This is due to the self-made `saveOnTyping`-directive. I will provide a fix for that.

lukasrad02:
> Try this:
> 
> in `ol-map-manager.ts` at the end of `registerPopupTriggers()`:

Thank you for the hint. I've only tried registering events on `this.olMap`, which does not provide keyboard events.

lukasrad02:
> I'm not sure whether `event.preventDefault()` and/or `event.stopPropagation()` would be good here...

I'm not aware of any default behavior of the escape key and didn't notice any differences with or without `preventDefault()`.
The same applies to `stopPropagation()`, there are no other custom features using the escape key.

So I would not include any of these. But if you generally include one of these methods in this project, I would follow your convention.

Dassderdie:
> I'm not aware of any default behavior of the escape key and didn't notice any differences with or without `preventDefault()`. The same applies to `stopPropagation()`, there are no other custom features using the escape key.
>
> So I would not include any of these. But if you generally include one of these methods in this project, I would follow your convention.

[...]

PR #610, opened 2023-01-19
Close popups with esc

lukasrad02:
Popups can now be closed by pressing Escape. This makes it easier to use the software in environments with keyboard-based instead of touch-based interaction.

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
@@ -454,6 +454,13 @@ export class OlMapManager {
         this.olMap.getView().on(['change:resolution', 'change:center'], () => {
             this.changePopup$.next(undefined);
         });
+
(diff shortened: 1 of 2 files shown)


benn02 (approved):
Looks good to me

Dassderdie (approved):
If this is the same as my code from #605 then LGTM.

I would be happy for a `Co-authored-by: Julian Schmidt <julian.schmidt@student.hpi.de>` in a squashed commit message.

lukasrad02:
> If this is the same as my code from #605 then LGTM.

Oh, actually not. I've changed something in the code while checking whether we need `stopPropagation()` or `preventDefault()`. Although this change did not affect the functionality, I planned but forgot to revert it. With https://github.com/hpi-sam/digital-fuesim-manv/pull/610/commits/5b4633a2e7e27aff1d9597aecc8e916680bcd998 it should be exactly what you suggested.


> I would be happy for a `Co-authored-by: Julian Schmidt <julian.schmidt@student.hpi.de>` in a squashed commit message.

:+1:

PR #614, opened 2023-01-23
Introduce state-based patient model

### LINKED ISSUE #335: Allow changing the treatment factor

### LINKED ISSUE #402: Change Patientsystem from life-point-based to state-based

ClFeSc:
Closes #335 
Closes #402 

Continuation of https://github.com/hpi-sam/digital-fuesim-manv/pull/490. Note that the remarks and comments made there were not incorporated here, so they're still relevant.

Design by @hpistudent72

@hpi-sam/bp2022hg1 You can assign reviewers at your own discretion.


shared/src/store/action-reducers/utils/calculate-treatments.spec.ts
```diff
@@ -68,12 +68,12 @@ function assertCatering(
 function addPatient(
     state: Mutable<ExerciseState>,
     pretriageStatus: PatientStatus,
+    // TODO: Use this parameter again. When it is ignored tests will fail.
```
ClFeSc, line 71: I think we have to investigate why there are no failing tests when this is not used. Because as far as I can tell this means that all the tests that should check whether some behavior occurs for some colors, are completely independent of the color and therefore (partly) useless.

Dassderdie: I thought the whole algorithm only relies on the pretriageStatus, to punish incorrect pretriage. As the pretriageStatus is set to the realStatus after 5 min or so, the treatments effectively use the realStatus then.

shared/src/utils/validators/is-id-map.ts
```diff
@@ -1,48 +1,73 @@
...
+//             ) {
+//                 return 'invalid';
+//             }
+//             return instanceChildrenWithKey.map(([, child]) => child);
+//         },
+//         { toClassOnly: true }
+//     );
+//     const validateNested = ValidateNested({ ...validationOptions, each: true });
+//     return combineDecorators(transform, validateNested);
+// }
```
Dassderdie, line 55: Is this correct?

ClFeSc: Ah, forgot to remove the old code. Will do later.

shared/src/store/action-reducers/utils/calculate-treatments.spec.ts
ClFeSc: Good point. I'll remove the parameter and the comment.

benn02:
A few general Things: It would be nice if there were more than 1 Test Patient available w/o an external CSV. Not only for testing, but also it would be nice if basic usability was demonstrable w/o said external file. I would suggest a build in CSV that then must also adhere to our new licence(keep in mind). I would also suggest that the Speed in which Patients switch between phases is communicated more clearly to the user. In the offline equivalent, the trainers do also control the time, not a transparent factor. I would suggest changing the UI in a way where the trainers can control the time between phase changes directly. Lastly, migrations are missing. Since there is no direct translation between old and new patients, maybe deleting all patients would be an option. but keep in mind to keep statistics so exercises done before this change can be evaluated with the software after this change.

benn02 (commented):
A few general Things: It would be nice if there were more than 1 Test Patient available w/o an external CSV. Not only for testing, but also it would be nice if basic usability was demonstrable w/o said external file. I would suggest a build in CSV that then must also adhere to our new licence(keep in mind). I would also suggest that the Speed in which Patients switch between phases is communicated more clearly to the user. In the offline equivalent, the trainers do also control the time, not a transparent factor. I would suggest changing the UI in a way where the trainers can control the time between phase changes directly. Lastly, migrations are missing. Since there is no direct translation between old and new patients, maybe deleting all patients would be an option. but keep in mind to keep statistics so exercises done before this change can be evaluated with the software after this change.

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,311 @@
...
+const defaultImage: ImageProperties = {
+    url: '/assets/patient.svg',
+    height: 80,
+    aspectRatio: 1,
+};
+
+export function parsePatientData(importString: string) {
+    const patientData: PatientData[] = [];
+
+    const splitString = importString.split('|');
```
benn02, line 30: This applies to the one CSV File that we got I would like to discuss whether we should support more generic CSVs but most likely not in this PR

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
```diff
@@ -0,0 +1,311 @@
...
+        public startingPhase: number,
+        public finished: boolean
+    ) {}
+}
+
+function generateHealthStates(
+    healthStateInformation: string[],
+    pretriageInformation: PretriageInformation
+) {
+    const phaseTime = 12 * 60 * 1000;
```
benn02, line 130: I have seen this or similar constants (x * 60 *1000)as throughout all the Changes, and I don't really understand them. Maybe I would put them in a combined place.

shared/src/store/action-reducers/exercise.ts
```diff
@@ -98,11 +97,26 @@ export namespace ExerciseActionReducers {
...
 
-                currentPatient.currentHealthStateId = patientUpdate.nextStateId;
-                currentPatient.health = patientUpdate.nextHealthPoints;
+                currentPatient.currentHealthStateName =
+                    patientUpdate.nextStateName;
                 currentPatient.stateTime = patientUpdate.nextStateTime;
                 currentPatient.treatmentTime = patientUpdate.treatmentTime;
-                currentPatient.realStatus = getStatus(currentPatient.health);
+                if (currentPatient.treatmentHistory.length === 0) {
+                    for (let i = 0; i <= (60 * 1000) / tickInterval; i++) {
```
benn02, line 105: Magic Number

shared/src/models/patient-template.ts
```diff
@@ -73,40 +73,53 @@ export class PatientTemplate {
...
+                            newLatestTime,
+                            condition.isBeingTreated,
+                            condition.requiredMaterialAmount,
+                            condition.requiredNotArztAmount,
+                            condition.requiredNotSanAmount,
+                            condition.requiredRettSanAmount,
+                            condition.requiredSanAmount,
+                            condition.matchingHealthStateName
+                        );
+                    });
```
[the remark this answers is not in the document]
ClFeSc: The idea was that not every patient should be identical, meaning they would be easier to predict. With a little randomness, it's possible to generate a more natural behavior of patients without blowing up the CSV file.

christianzoellner:
Is this PR still ongoing?

ClFeSc:
> Is this PR still ongoing?

Theoretically, yes. Practically, I haven't had enough time over the course of the semester and time with the exams to work on this. Also, I have had some problems understanding the code as I have not written it. But I'll try to look into this again in the next few weeks.

Dassderdie:
> Theoretically, yes. Practically, I haven't had enough time over the course of the semester and time with the exams to work on this. Also, I have had some problems understanding the code as I have not written it. But I'll try to look into this again in the next few weeks.

Was there already a consensus about how to handle migrations of patients here? While I believe it is in theory possible to migrate them, this would be a lot of work.

In addition, I believe the code as @hpistudent72 wrote it would be significantly less performant than now (I estimate ca. one order of magnitude) . I proposed a solution to not only fix this, but also **improve** the performance by ca. one order of magnitude and reduce the size of history exports. Implementing this would also require even more work.

TLDR: this PR could be more work than just a simple merge.

christianzoellner:
Thanks @ClFeSc and @Dassderdie for your responses and your continued activities as a volunteer developers on this project!

I really do not want to push you by raising this issue again, but the topic of the patient model came up during a meeting I had this weekend at BABZ. 

Sadly I am not really familiar with the code, but I am available to answer any questions you might have about the underlying model and concept.

Dassderdie:
Maybe it would also help if this PR would be split up into multiple smaller ones. For example, the refactored patient system could be separated from the CSV import and the smaller UI improvements.

frontend/src/app/pages/exercises/exercise/shared/utility/parse-csv.ts
Dassderdie: Just because I saw this now: that's 12 minutes (min * s * ms).

lukasrad02:
Keeping open as reference until we decide how to continue with the patient state model (and whether to reuse the technical challenge implementation)

PR #616, opened 2023-01-24
Add type property to elements in the state

lukasrad02:
This PR adds a `type` property to elements stored in the state.

This allows us to dispatch based on the value of this property instead of guessing the type based on presence or absence of other attributes.


shared/src/utils/type-state-selector-map.ts
```diff
@@ -0,0 +1,15 @@
...
+    hospital: 'hospitals',
+    mapImage: 'mapImages',
+    material: 'materials',
+    patient: 'patients',
+    personnel: 'personnel',
+    simulatedRegion: 'simulatedRegions',
+    transferPoint: 'transferPoints',
+    vehicle: 'vehicles',
+    viewport: 'viewports',
+} as const;
```
Dassderdie, line 13: You could use `satisfies { [key: string]: keyof ExerciseState}` to validate the plurals.

shared/src/utils/type-state-selector-map.ts
```diff
@@ -0,0 +1,15 @@
+export const typeSelectorMap = {
```
Dassderdie, line 1: I don't really like this name as one could easily confuse this with the whole selector-concept from Redux that we use too.

What about `ElementTypePluralMap`?

shared/src/utils/type-state-selector-map.ts
Dassderdie, line 13: In general, I'm not fully convinced that making a distinction between singular and plural is good. It adds more code and doesn't (as far as I can see it) more information or clarification to the programmer.

But if you think this is worth it, then go for it 👍

shared/src/utils/type-state-selector-map.ts
```diff
@@ -0,0 +1,15 @@
...
+    alarmGroup: 'alarmGroups',
+    client: 'clients',
+    hospital: 'hospitals',
+    mapImage: 'mapImages',
+    material: 'materials',
+    patient: 'patients',
+    personnel: 'personnel',
```
Dassderdie, line 8: FYI: I regret this naming decision. Initially we had `personnel` and `personnels`, but changed it, after realising that `personnels` isn't a valid plural in the English language.
In hindsight, I believe this should have been a conscious neologism to still have the destination in variable names.

```ts
const personnel: Personnel;
const personnels: Personnel[];
```

Or using a synonym to personnel...

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/element-manager.ts
```diff
@@ -140,6 +130,4 @@ export abstract class ElementManager<
  */
 const featureKeys = {
     value: 'elementValue',
-    // TODO: In the future the type should be saved in the element itself
-    type: 'elementType',
 };
```
Dassderdie, line 133: I think you can now change this back to just a string (YAGNI).

shared/src/utils/type-state-selector-map.ts
lukasrad02: I've added a validation that also checks that the key is a valid element type.

shared/src/utils/type-state-selector-map.ts
lukasrad02: Yes, the name isn't that nice. It was simply the first name that came to my mind when I first implemented this to test the typings.

I've changed it according to your suggestion here and renamed the derived types (in `shared/src/store/action-reducers/exercise.ts` and `shared/src/store/action-reducers/utils/spatial-elements.ts`), too

shared/src/utils/type-state-selector-map.ts
lukasrad02: We've discussed this in our team and prefer using the singular.

shared/src/utils/type-state-selector-map.ts
lukasrad02: Since we have to use the mapping between singular and plural anyway, I don't think that `personnel` as plural is a big problem.

If we *would* use `personnels`, we could just append an `s`, but this would break as soon as we've got something like `category`/`categories`.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/element-manager.ts
lukasrad02: :+1:

shared/src/utils/type-state-selector-map.ts
```diff
@@ -0,0 +1,24 @@
...
+    simulatedRegion: 'simulatedRegions',
+    transferPoint: 'transferPoints',
+    vehicle: 'vehicles',
+    viewport: 'viewports',
+
+    // Typescript does not allow literal types for indexes
+    // eslint-disable-next-line @typescript-eslint/consistent-indexed-object-style
+} as const satisfies Record<ElementType, keyof ExerciseState>;
```
Dassderdie, line 22:
```suggestion
} as const satisfies { [Key in ElementType]: keyof ExerciseState};
```

Does this work?

lukasrad02: Yes, thank you!

Greenscreen23 (changes requested):

shared/src/store/action-reducers/utils/calculate-treatments.ts
```diff
@@ -92,38 +93,18 @@ function tryToCaterFor(
...
-    return (element as Personnel).personnelType !== undefined;
-}
-
-function isMaterial(
-    element: Mutable<Material | Patient | Personnel>
-): element is Mutable<Material> {
-    // as Material does not include any distinguishable properties, we will check if it is not of type Personnel or Patient
-    return !isPersonnel(element) && !isPatient(element);
-}
-
```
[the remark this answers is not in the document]
lukasrad02: In what way is using the exact same code as the one from the type guard, just inline, not "at runtime"?

shared/src/state-migrations/17-add-type-property.ts
```diff
@@ -0,0 +1,329 @@
...
+        Object.values(typedState.alarmGroups).forEach((alarmGroup) => {
+            alarmGroup.type = 'alarmGroup';
+        });
+
+        Object.values(typedState.clients).forEach((client) => {
+            client.type = 'client';
+        });
+
+        Object.values(typedState.eocLog).forEach((logEntry) => {
+            logEntry.type = 'logEntry';
```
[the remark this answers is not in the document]
lukasrad02: Nice catch! Fixed in [84b8ed7](https://github.com/hpi-sam/digital-fuesim-manv/pull/616/commits/84b8ed727e274f8e4b26651a64f7c501395bca2e)

shared/src/state-migrations/17-add-type-property.ts
```diff
@@ -0,0 +1,329 @@
...
+                [key: UUID]: {
+                    type: 'alarmGroup';
+                };
+            };
+            clients: {
+                [key: UUID]: {
+                    type: 'client';
+                };
+            };
+            eocLog: { type: 'logEntry' }[];
```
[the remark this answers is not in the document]
lukasrad02: Fixed in [84b8ed7](https://github.com/hpi-sam/digital-fuesim-manv/pull/616/commits/84b8ed727e274f8e4b26651a64f7c501395bca2e), too

shared/src/models/element.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { MapImage } from './map-image';
+import type {
+    Material,
+    Patient,
+    Personnel,
+    SimulatedRegion,
+    TransferPoint,
+    Vehicle,
+    Viewport,
+} from '.';
```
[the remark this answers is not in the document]
lukasrad02: Seems like the auto-import of VS Code made some weird decisions here. Fixed it.

shared/src/store/action-reducers/exercise.ts
```diff
@@ -122,20 +124,25 @@ export namespace ExerciseActionReducers {
...
             return draftState;
         },
         rights: 'server',
     };
 }
 
+type TransferTypePluralMap = Pick<
+    ElementTypePluralMap,
+    'personnel' | 'vehicle'
+>;
```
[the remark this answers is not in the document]
lukasrad02: Yes, you're right. I've just went trough the code and fixed all places where it was broken, without thinking about the overall structure.

I'm now exporting the literal union in `shared/src/store/action-reducers/transfer.ts` to use it for the `Pick<>` in `shared/src/store/action-reducers/exercise.ts`.

shared/src/store/action-reducers/utils/calculate-treatments.ts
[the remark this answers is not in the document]
Greenscreen23: Inline guards are also runtime, you are right. However, does typescript infer the type in this scenario?
```ts
if (element.type === 'material') {
     // does typescript know that element is a Material in here?
}
```

If it does, then you can ignore my comment. I just thought this might be something that typescript doesn't do and that is fixed by using type guards with the return value set to `element is Material`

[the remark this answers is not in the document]
lukasrad02: > However, does typescript infer the type in this scenario?

Yes.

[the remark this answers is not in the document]
Dassderdie: This should be the relevant section in the ts documentation: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions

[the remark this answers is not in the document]
Greenscreen23: I've just attempted to reproduce it, but I've not managed to reproduce this. Maybe you can tell me what is wrong with my poc? 

```ts
const poc = (element: Readonly<{ type: keyof ElementTypePluralMap }>): void => {
    if (element.type === 'material') {
        console.log(element.id)
    }
    if (isMaterial(element)) {
        console.log(element.id)
    }
}

const isMaterial = (element: Readonly<{ type: keyof ElementTypePluralMap }>): element is Material =>
    element.type === 'material'
```

I currently get this error
```
error TS2339: Property 'id' does not exist on type 'Readonly<{ type: "simulatedRegion" | "transferPoint" | "alarmGroup" | "vehicle" | "client" | "hospital" | "mapImage" | "material" | "patient" | "personnel" | "viewport"; }>'.

34         console.log(element.id)
```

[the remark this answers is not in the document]
Dassderdie: Your element needs to have the type of the actual Elements (`Material`, `Personnel`, ...).

So something like this: `ExerciseState[keyof ElementTypePluralMap]`(or `ExerciseState[ElementTypePluralMap[keyof ElementTypePluralMap]]`) instead of the `Readonly<{ type: ... }>`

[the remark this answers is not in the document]
Greenscreen23: You're correct, I understand it now. Thank you?

Greenscreen23 (approved):
lgtm

PR #622, opened 2023-01-24
Fix #621 Features disappear from the map in imported exercises

### LINKED ISSUE #621: Features disappear from the map in imported exercises
After importing the exercise https://github.com/hpi-sam/digital-fuesim-manv_test-scenarios/blob/main/exercise-state-2022-05-31-small-demo.json elements disappear from the map after performing an action.

For example, when you drop a patient onto a vehicle an then move the map, all other patients, material, and the personnel disappears:

https://user-images.githubusercontent.com/49586507/214319673-d3a39fd5-315b-41b7-8b8f-4ec594fde8e4.mp4

This issue has first been detected during the review of #607, but as it also appears at the current state of the `dev` as well as `main` branch.

The problem seems to be in the frontend only, as there is no websocket communication despite the action for the patient dropped onto the vehicle. Additionally, the browser console shows several type errors:

```
[...]
ERROR TypeError: can't access property "vehicleTemplates", exercise is undefined
ERROR TypeError: can't access property "mapImageTemplates", exercise is undefined
ERROR TypeError: can't access property "clients", exercise is undefined
ERROR TypeError: can't access property "currentStatus", exercise is undefined
ERROR TypeError: can't access property "configuration", exercise is undefined
ERROR TypeError: can't access property "transferPoints", exercise is undefined
[...]
```

Dassderdie:
### Bug description

This bug can only be observed if (at least) two actions are simultaneously optimistically proposed **and** the client didn't receive any other action from the server yet since he joined the exercise.

In the provided reproduction (https://github.com/hpi-sam/digital-fuesim-manv/issues/621#issuecomment-1402155917) this was the case as dropping a patient on a vehicle doesn't just propose the `[Vehicle] Load vehicle`-action, but also sends a `[Patient] Move patient`-action right before.

The bug only occurred in these specific circumstances as we need to trigger the `general and "safe" way` to perform an action from the server (= reset the state in the client to the `serverState` and reapply all optimistic actions that are still pending) and the `serverState` must be the initial one.

The actual bug was that when initializing the `OptimisticUpdateHandler`, `getState` would still return `undefined` as the initial exercise state hasn't been retrieved yet, and the store would therefore be still empty for this value. Therefore the initial `serverState` in the `OptimisticUpdateHandler` would be `undefined` instead of the correct value.

### Fix

Initialize the `optimisticUpdateHandler` only after the exercise state is saved in the store.

### DIFF
--- a/frontend/src/app/core/exercise.service.ts
+++ b/frontend/src/app/core/exercise.service.ts
@@ -54,28 +54,11 @@ export class ExerciseService {
         ...socketIoTransports,
     });
 
-    private readonly optimisticActionHandler = new OptimisticActionHandler<
+    private optimisticActionHandler?: OptimisticActionHandler<
         ExerciseAction,
         ExerciseState,
         SocketResponse
-    >(
-        (exercise) =>
-            this.store.dispatch(createSetExerciseStateAction(exercise)),
-        () => selectStateSnapshot(selectExerciseState, this.store),
-        (action) => this.store.dispatch(createApplyServerActionAction(action)),
-        async (action) => {
-            const response = await new Promise<SocketResponse>((resolve) => {
-                this.socket.emit('proposeAction', action, resolve);
-            });
-            if (!response.success) {
-                this.messageService.postError({
-                    title: 'Fehler beim Senden der Aktion',
-                    error: response.message,
-                });
-            }
-            return response;
-        }
-    );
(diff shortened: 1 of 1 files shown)


lukasrad02 (approved):
Looks good to me, thanks for the quick fix!

PR #627, opened 2023-01-26
Update pipeline actions to node 16

lukasrad02:
The old actions were based on Node 12, which has reached EOL: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


lukasrad02:
I've incremented all action versions to the newest available version so there are no Node 12 based actions any longer. This works for all actions except `cypress-io/github-action`.

This action was previously `v2` which was working fine. For the newest version (`v5`) as well as the next higher version (`v3`) I did not managed to get it running. The pipeline always fails with the following error.

```
Error: Cannot find module 'cypress'
Require stack:
- /home/runner/work/_actions/cypress-io/github-action/v***/dist/index.js
```

I've already tried disabling caching of the Cypress binary by changing the name of the cache entry (see https://github.com/hpi-sam/digital-fuesim-manv/pull/627/commits/6402002905c41d7569773628eb1d203e6db30a57), but this did not help.

@ClFeSc I was told that you have the best knowledge concerning the pipeline and Cypress. Do you have an idea how this could be fixed?

ClFeSc:
> @ClFeSc I was told that you have the best knowledge concerning the pipeline and Cypress. Do you have an idea how this could be fixed?

I'll look into it on Thursday or Friday.

ClFeSc:
I've re-run the cypress job with debug logs enabled (<https://github.com/hpi-sam/digital-fuesim-manv/actions/runs/4017160960/jobs/7029491207>). This showed that the `cypress-io/github-action@v3` action actually also caches something. I'm not sure whether this is relevant, but it was not obvious. We also have two Cypress versions (`12.3.0` and `9.7.0`) available, although this didn't change from before.

I then tried running a Cypress command that doesn't have anything to do with Cypress (`echo 'Command should be running'`, see https://github.com/hpi-sam/digital-fuesim-manv/pull/627/commits/69ed40298fafcbd71638907311b8a954eb389d3c). However, this still results in the same error. I'll try to look further into this tomorrow, but my guess would be that `node@16`, anything newer than `cypress-io/github-action@v2`, and `cypress@9.7.0` don't work together, like `cypress-io/github-action@v3` requiring `cypress@10.x.y` or something like this.

ClFeSc:
I was able to prevent this problem by adding `cypress^9.5.3` as a dev dependency to the root package. I'm not sure this is a good solution, but it does work.

lukasrad02:
> I was able to prevent this problem by adding `cypress^9.5.3` as a dev dependency to the root package. I'm not sure this is a good solution, but it does work.

Thank you for your investigation and fix. It isn't a perfectly fine solution, but I think its okay

Nils1729 (approved):

PR #630, opened 2023-01-27
Remove unecessary unsupportedChangeProperties

Dassderdie:
This PR removes a relict from the time when we didn't have any clue about OpenLayers.

The original idea was:
* If the element changes, we have to imperatively tell OpenLayers what exact properties of the feature should be changed (e.g., via `feature.setPosition()`)
* The same would apply to all styles as well as elements that are represented through multiple features
* This would be a lot of code
* Instead of implementing these changes for every feature, we could just delete it and create it new. We would, therefore, only have to specify the style during the creation of the element and not change the features in between.

These predictions didn't work out:
* We always just used a single feature to represent an element
* Features can have a style function that can easily be reevaluated every time the element changes.

Therefore, we didn't use `unsupportedChangeProperties` (or made incorrect use of it).

The only properties we put in it were:
* `id` - the `id` never changes as elements are identified by it
* `internalName` (for TransferPoint) - the style is reevaluated when it changes  -> no problems
* `image` - the style is reevaluated when it changes -> no problems

You can test this out by adding the following code to `shared\src\store\action-reducers\exercise.ts` in the `exerciseTick` function:

```ts
            const rtwImage = {
                url: '/assets/rtw-vehicle.png',
                height: 100,
                aspectRatio: 3693 / 1670,
            };

            const ktwImage = {
                url: '/assets/ktw-vehicle.png',
                height: 100,
                aspectRatio: 5046 / 2465,
            };
            for (const vehicle of Object.values(draftState.vehicles)) {
                vehicle.image =
                    draftState.currentTime % 2000 === 0 ? ktwImage : rtwImage;
            }
```

This PR removes this feature. Should there ever be a FeatureManager that wants to use it, it would probably be sufficient to do this on a higher abstraction level instead of directly in the `ElementManager`.

Btw. I believe this is the first refactoring that actually reduces the Lines of code 😉

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/catering-lines-feature-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/catering-lines-feature-manager.ts
@@ -11,15 +11,9 @@ import { LineStyleHelper } from '../utility/style-helper/line-style-helper';
 import { ElementManager } from './element-manager';
 
 export class CateringLinesFeatureManager
-    extends ElementManager<
-        CateringLine,
-        LineString,
-        ReadonlySet<keyof CateringLine>
-    >
+    extends ElementManager<CateringLine, LineString>
     implements FeatureManager<LineString>
 {
-    readonly unsupportedChangeProperties = new Set(['id'] as const);
-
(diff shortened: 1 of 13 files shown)


Greenscreen23 (approved):
Nice! :+1:

PR #634, opened 2023-01-31
Remove generic type LineString | Point in favor of the default Geometry

Greenscreen23:


Greenscreen23:
I've looked at all occurences of `Feature<` and `VectorSource` and these are all the occurences I found that I missed during my refactoring

benn02 (approved):
If those are all occurrences, the code Looks Good To me

PR #635, opened 2023-01-31
Fix Cannot Transfer patients to Hospital

### LINKED ISSUE #632: Cannot transfer to hospitals
Currently, it is not possible to transfer vehicles to hospitals.

Steps to reproduce (as a trainer or trainee):
- Start an exercise
- Add a hospital, transfer point, patient and RTW to the map
- Add the hospital as a transfer option to the transfer point
- Add the patient to the RTW
- Start the exercise
- Try to transfer the RTW to the hospital

You should see a popup like this:
Screenshot_20230131_140846

Expected behaviour:
- There should be an option to transfer the RTW to the hospital in the popup.

benn02:
closes #632


Greenscreen23 (approved):
I came looking far and wide, and I found no other occurence where you might have missed something. I sadly also did not find any rock hyrax as well :/

Issue #574, opened 2022-12-09
Basic "Impressum" (imprint) and "Datenschutzerklärung" (privacy page)


anonym-HPI:
There should be in the footer on the start page to webpage/impressum and webpage/datenschutz.
Webpage should be dynamic, so the links should be something like `href="/impressum"`.

When clicking these links there should be two template pages telling how the person can overwrite them.
The templates should have a way (at the top, I would say) to go back to the startpage and probably include the footer as well (to switch between the to pages).

I would put the pages in a folder under docker or a new folder in the root directory called `custom_webpages` or something like that.
These pages should then be able to be overwritten with the help of a mounting of docker volume (example should be included in `docker-compose.yml` and just needs to be uncommented).

PR #636, opened 2023-01-31
add AGPL v3.0 (or later) license

anonym-HPI:
This is the proposal for the AGPL v3.0 (GNU Affero General Public License Version 3.0 https://www.gnu.org/licenses/agpl-3.0.de.html).

Everyone that approves this Pull Request will agree to the license to have all his/her attributions to the software be published under this license.

In the future, when this PR gets through I would enable forced signed commits, to make clear that people commiting agree to publish there code under this license have the right to do so.

Especially the LICENSE-README and the text should be read. I tried to make it possible to have the images included with another license, without making it possible to exploit this, e.g. using images as proprietary binary files and someone publishing the source code to use the binary image file, without publishing the binary source code included in an image (or something similiar exploitable to undermine the license).
Also I spoke with Lukas to not include a header with the license info in every file, this seems to be also not be necessary (https://www.gnu.org/licenses/gpl-howto.html.en "Why license notices?")


Greenscreen23 (changes requested):

LICENSE
```diff
@@ -0,0 +1,661 @@
...
+    License will therefore apply, along with any applicable section 7
+    additional terms, to the whole of the work, and all its parts,
+    regardless of how they are packaged.  This License gives no
+    permission to license the work in any other way, but it does not
+    invalidate such permission if you have separately received it.
+
+    d) If the work has interactive user interfaces, each must display
+    Appropriate Legal Notices; however, if the Program has interactive
+    interfaces that do not display Appropriate Legal Notices, your
+    work need not make them do so.
```
lukasrad02, line 221: At the moment, we're not showing any legal notices in our software.

As far as I understand the AGPL, we're free to omit those notices since our work is not based on any (A)GPL work. But omitting this would allow third parties to omit it, too. So we should integrate those notices when working on #574 and maybe even work on #574 before merging this.

Greenscreen23:
The license mentions the contributers as the ones listed in the readme. We did not yet add ourselves to that list, so #639 should probably be resolved before we merge this pr.

LICENSE-README
```diff
@@ -0,0 +1,21 @@
+Copyright (c) 2023 See the README.md at the bottom for authors/contributors.
+
+This is software is a digital implementation of the "FüSim MANV"/"FueSim MANV" (Führungssimulation Massenanfall von Verletzen), a German simulation system for training emergency medical services leadership personnel on how to manage Mass Casualty Incidents.
+
+The software is licensed with the GNU Affero General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+Some files are explicitly excluded and have their own license. These files are images that are used in the software. This exlusion is not allowed to be exploited, e.g. to enhance the features of this software and every image must be easily replacable by a person wanting to use this software, by just replacing the URL link for each image in an export created via the user interface of this software. The individual licenses for the images can be looked up in frontend/src/assets/image-sources.md. This path to the images can be changed in this file (LICENSE-README), if the structure of this software needs it, but must be included, as long as they are images.
```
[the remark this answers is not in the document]
christianzoellner: For me, the initial text was also not fully comprehensible, I like Lukas' proposal better.

However, I think that "purpose of being displayed" might be to narrow, as the BABZ patient data files might also need to be covered. 
I would define the permitted scope for the exception as "assets copied from external sources (i.e. that are not specifically designed for this project) and that are not sourcecode" or some such. That way, anyone creating content especially for this project might not make use of the exceptions.

christianzoellner (commented):
Thanks for the proposal for `LICENSE` and the related `LICENSE-README`.

One specific addition that I would propose for the `LICENSE-README` is to add an exception for the FüSim Patient Files besides the exception for the images. I have written permission from the BABZ that we are allowed to publish the patient files as `.csv`, but we are not authors and thus cannot allow redistribution.

anonym-HPI:
> Thanks for the proposal for `LICENSE` and the related `LICENSE-README`.
> 
> One specific addition that I would propose for the `LICENSE-README` is to add an exception for the FüSim Patient Files besides the exception for the images. I have written permission from the BABZ that we are allowed to publish the patient files as `.csv`, but we are not authors and thus cannot allow redistribution.

Thanks, we should make an exception for that. Patient templates or templates in general that can be part of an export should be excluded to be forced to be under this license.
As you said as we have no author rights to them, they should be excluded.
If someone has a good way of saying this, a change request would be good.
Question is, should we wait for the bachelor thesis of @hpistudent72 to be merged, as this changes maybe smth? Maybe we can find a wording that includes such changes.

LICENSE
anonym-HPI: Yes, that should probably happen, question is, is un the homepage enligh or would there be a footer needed?

lukasrad02: My idea was to add a separate page with license notices that could state the license of this project and include notices on third-party-libraries, too. A link to this page can then be added to the footer. I'm currently working on this in PR #643.

But in general, it should suffice to have this notice somewhere in the frontend, I don't think that every page has to include a full copyright/license notice:

> If the interface presents a list of user commands or options, such as a menu, a prominent item in the list meets this criterion.

lukasrad02:
> In the future, when this PR gets through I would enable forced signed commits, to make clear that people commiting agree to publish there code under this license have the right to do so.

Just for clarification, since the term "signed commit" can be misunderstood: If I remember correctly, you're talking about signing off the commits (`Signed-off-by`, `git commit -s`), not about cryptographically signing the commits (GPG keys, `git commit -S`)?

LICENSE-README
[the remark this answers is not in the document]
anonym-HPI: > For me, the initial text was also not fully comprehensible, I like Lukas' proposal better.
> 
> However, I think that "purpose of being displayed" might be to narrow, as the BABZ patient data files might also need to be covered. I would define the permitted scope for the exception as "assets copied from external sources (i.e. that are not specifically designed for this project) and that are not sourcecode" or some such. That way, anyone creating content especially for this project might not make use of the exceptions.




> Thanks for the proposal for `LICENSE` and the related `LICENSE-README`.
> 
> One specific addition that I would propose for the `LICENSE-README` is to add an exception for the FüSim Patient Files besides the exception for the images. I have written permission from the BABZ that we are allowed to publish the patient files as `.csv`, but we are not authors and thus cannot allow redistribution.

Do you mean the data in https://github.com/hpi-sam/digital-fuesim-manv/blob/licensing/add-license-594/shared/src/data/default-state/patient-templates.ts#L811
Line 811 and below?

christianzoellner:
> Question is, should we wait for the bachelor thesis of @hpistudent72 to be merged, as this changes maybe smth? Maybe we can find a wording that includes such changes.

Not necessary to wait.

As far as I know, the current patients are hand-written based on the FüSim datasets, so currently there is no need to have a licencse exception at all. However, once the code from @hpistudent72 is merged, we have a csv proper import for this and other datasets, and it will be best to just deploy the datasets in the repo.

Either way, having a suitable generic exception in the licence that covers both images and datasets is useful.

anonym-HPI:
I have created a new commit, I am not sure if the text about "datasets" needs be that way, looking at https://github.com/hpi-sam/digital-fuesim-manv/pull/614, I can't see any csv included, are these always then imported at runtime or is there a way to host them, to be accessible for every exercise? If only at runtime I think this exclusion wouldn't be necessary or could be rewritten to only include datasets that can also be importet at runtime.
@christianzoellner @ClFeSc

LICENSE-README
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2023 See the README.md at the bottom for authors/contributors.
```
ClFeSc, line 1: I'd make this whole file a `.md` file with all the appropriate formatting (headings, links, etc.) to make it easier to read.

LICENSE-README
```diff
@@ -0,0 +1,27 @@
...
+(at your option) any later version.
+All exceptions can be part of another license chosen by the author, but using this software with such exceptions allows everyone for testing purposes to inspect these exceptions and test if they conform to the given restrictions (rules) given to them, for example can datasets be exported by a user and imported for testing purpose only or read with a program, to check they comply to the restrictions. For this the user that uses such datasets in a software must have the permission of the author to give the users of this software these rights.
+
+We have decided to not have the license header in every file, as this LICENSE-README file should be enough (https://www.gnu.org/licenses/gpl-howto.html.en) to make it clear that the whole software/program is under the chosen license. But if it will become necessary this option is possible. Still every file in this repository/folder and subfolders are under the mentioned license, except the mentioned exluded images.
+
+Fuesim MANV is free software: you can redistribute it and/or modify
```
ClFeSc, line 17: Careful! I don't think the Fuesim MANV (developed by the BABZ) is free software, only the digital-fuesim-manv (this rep) is.

README.md
```diff
@@ -278,6 +278,8 @@ If you need to read from the state to change it, you should do this inside the a
 
 ## Licenses and Attributions
 
+-   Read [LICENSE-README](LICENSE-README) for information about the used license
+    -   Update the date at the top accordingly (see https://www.gnu.org/licenses/gpl-howto.html.en under copyright notice)
```
ClFeSc, line 282:
```suggestion
    -   Update the date at the top accordingly (see <https://www.gnu.org/licenses/gpl-howto.html.en#copyright-notice>)
```

ClFeSc, line 282: Maybe this hint is better suited for the `LICENSE-README` file?

ClFeSc (commented):
Some notes on the form of the presentation, for now.

LICENSE-README
anonym-HPI: We could do that, the problem is the longetivity of the .md format. Sounds unrealistic, but though having just a normal text file with no ending would be the best compatible format.
As long as the .md file can be understand correctly and works when read without a Markdown viewer and still works, we could do that.
The LICENSE file could also be an .md file (https://www.gnu.org/licenses/agpl-3.0.md).

LICENSE-README
anonym-HPI: It is an analog game, no software.
We should probably rewrite it to "digital Fuesim MANV"

README.md
ClFeSc: Yes.

LICENSE-README
anonym-HPI: Files are now `.md`, I made the links, do you have a proposal for heading and so on?

benn02 (approved):

LICENSE-README.md
```diff
@@ -0,0 +1,29 @@
...
+(at your option) any later version.
+All exceptions can be part of another license chosen by the author, but using this software with such exceptions allows everyone for testing purposes to inspect these exceptions and test if they conform to the given restrictions (rules) given to them, for example can datasets be exported by a user and imported for testing purpose only or read with a program, to check they comply to the restrictions. For this the user that uses such datasets in a software must have the permission of the author to give the users of this software these rights.
```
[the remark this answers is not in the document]
anonym-HPI: This paragraph should make it possible for people to use custom data sets, e.g. own types of patients or vehicles. Author is the one creating the data sets, could be more clear.
The exception with testing is meant so that other people can check if they just added datasets or if they changed smth which falls under AGPL. So that we or others can't be sued for copying and inspecting the code and datasets from a server or so, for testing purposes to check if a violation accured.
Don't know if this is necessary.
Do you get the idea?
Exceptions should always be able to be checkable, even by a third person (e.g. you or me, given the data from a person taking part in an online training course of X), if it violates this license.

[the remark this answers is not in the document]
Nils1729: I can see why we would want to inspect the code, although permission and free access is granted by AGPL itself. 

However, I can't imagine anything we could learn from inspecting the *datasets* as all "added" functionality is by definition reflected in the code we and anyone can access anyways. However, this 'checking' clause reads like any dataset must be effectively publicly accessible, possibly with restricted permission use by others, but still accessible.

Few authors of datasets might want to use the software under this condition and again, I do not see the benefit of asserting anything about the datasets.

I suggest we define datasets more explicitly (e.g. What is `... allowed through the source code given template class` supposed to refer to?) and include them in the paragraph about images. They are just custom assets after all.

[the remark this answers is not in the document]
Dassderdie: KISS - Keep It Simple Stupid.
I don't see an advantage in adding a lot of legal text to the license.
1. It will be nearly impossible to enforce.
2. If we formulate it ourselves, it is probably not foolproof and would probably not hold up in a court.
3. If something is too complicated, people either ignore it or are put off by it and don't use this software.

Greenscreen23 (changes requested):
We should format the LICENSE-README.md to limit the line length to 70 (like AGPL), so it is easier to read. We should probably also add some markdown formatting (f.e. headings) to it :)

LICENSE.md
```diff
@@ -0,0 +1,660 @@
...
+get its source. For example, if your program is a web application, its
+interface could display a "Source" link that leads users to an archive
+of the code. There are many ways you could offer source, and different
+solutions will be better for different programs; see section 13 for
+the specific requirements.
+
+You should also get your employer (if you work as a programmer) or
+school, if any, to sign a "copyright disclaimer" for the program, if
+necessary. For more information on this, and how to apply and follow
+the GNU AGPL, see <https://www.gnu.org/licenses/>.
```
Greenscreen23, line 660: Is this something thats supposed to be in the license? I think we can remove it.

They suggest adding a copyright line with a pointer to the license to each source code file. Is this something that we want to do? It could probably be done with a simple shell script, but we need to be wary of merge conflicts.

README.md
```diff
@@ -278,7 +278,8 @@ If you need to read from the state to change it, you should do this inside the a
 
 ## Licenses and Attributions
 
--   License information about used images can be found [here](frontend/src/assets/image-sources.md). All images are licensed under their original license.
+-   Read [LICENSE-README.md](LICENSE-README.md) for information about the used license
+-   License information about used images can be found [here](frontend/src/assets/image-sources.md). All images are licensed under their original license, for further information what is meant with an image, look at [LICENSE-README.md](LICENSE-README.md).
```
Greenscreen23, line 282:
```suggestion
Digital Fuesim MANV
Copyright (C) 2023 HPI-SAM

This program is free software: you can redistribute it and/or modify
it under the terms of the GNU Affero General Public License as
published by the Free Software Foundation, either version 3 of the
License, or (at your option) any later version.

This program is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU Affero General Public License for more details.

You should have received a copy of the GNU Affero General Public License
along with this program.  If not, see <https://www.gnu.org/licenses/>.

License information about used images can be found in [frontend/src/assets/image-sources.md](frontend/src/assets/image-sources.md). All images are licensed under their original license, for further about licensing of images and datasets refer to [LICENSE-README.md](LICENSE-README.md).
```

LICENSE-README.md
```diff
@@ -0,0 +1,29 @@
+Copyright (c) 2023 See the [README.md](README.md) at the bottom for authors/contributors.
+
+Update the date at the top accordingly (see <https://www.gnu.org/licenses/gpl-howto.html.en#copyright-notice>).
```
Greenscreen23, line 3: I think we can remove this line, because it is just a reminder

LICENSE-README.md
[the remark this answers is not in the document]
Greenscreen23: I'd suggest deleting line 15 and moving the `Especially, you must not use these exception for source code files.` below the dataset definition.

Greenscreen23 (changes requested):

LICENSE.md
anonym-HPI: We can't change anything in LICENSE.md, if you meant that. The license text is copyright protected (at the top).

We could make such notice in every file, but we would have to keep the years uptodate, this could be done via a plugin.
But I have written somewhere (either in License-readme.md or in a text at the top of the PR), that I would say we don't do it.
It seems not legally necessary, I posted somwhere a link explaining why they recommend it.

anonym-HPI:
Have changed a few things, also added license.html to maybe include notices to the third parties (this should probably be done also in LICENSE-README.md).
@lukasrad02 You mentioned noticing on third-libraries.

LICENSE-README.md
```diff
@@ -0,0 +1,24 @@
...
+The individual license for each image can be found in [frontend/src/assets/LICENSES-IMAGES.md](frontend/src/assets/LICENSES-IMAGES.md). Files added to this project may only use this exception if these files serve the only purpose of being displayed and will only be parsed for the purpose of displaying them.
+Another exception exists for datasets. Datasets are files that include no new features and just different compositions allowed through the source code given template class. Datasets must always be in human readable formats (for example csv, JSON, txt). Datasets license must be mentioned in a LICENSE-DATASET.md (or LICENSES-DATASETS.md) file in the folder where the dataset is placed, stating which file the license applies to and give information about the license and author. These datasets are not allowed to create any new features or enhance the software in any way, as to create elements that can be part of an export and are importable into the distributed source code (this software) under the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
```
christianzoellner, line 10: Add blank line (paragraph separator) before the dataset part.

Also, I propose some minor changes for readability.
```suggestion
Another exception to the default licence exists for datasets. Datasets are files that include no new features and just contain different compositions allowed through template classes given in the source code. Datasets must always be in human readable formats (for example csv, JSON, txt). Datasets license must be mentioned in a LICENSE-DATASET.md (or LICENSES-DATASETS.md) file in the respective folder where the dataset is placed. The licence files must state explicitly which files they refer to, who the authors are and which license applies. Datasets are not allowed to create any new features or enhance the software in any way, as to create elements that can be part of an export and are importable into the distributed source code (this software) under the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
```

frontend/src/assets/about/license.html
```diff
@@ -0,0 +1,11 @@
+This software is a digital implementation of the "FüSim MANV"/"FueSim MANV" (Führungssimulation Massenanfall von Verletzen), a German simulation system for training emergency medical services leadership personnel on how to manage Mass Casualty Incidents.
```
christianzoellner, line 1:
```suggestion
This software is a digital implementation of the "FüSim MANV"/"FueSim MANV" (Führungssimulation Massenanfall von Verletzen), a German paper-based simulation game for training emergency medical services leadership personnel in managing Mass Casualty Incidents.
```

christianzoellner (approved):
Looks good to me (after you shortened the ambiguous parts).

I have some minor proposals below.

LICENSE-README.md
```diff
@@ -0,0 +1,29 @@
...
+Especially, you must not use this exception for source code files.
+Another exception exists for datasets. Datasets are files that include no new features and just different compositions allowed through the source code given template class. Datasets must always be in human readable formats (for example csv, JSON, txt). Datasets license must be mentioned in a CUSTOM-LICENSE file in the folder stating which file is meant and which gives information about the license and author. These datasets are not allowed to create any new features or enhance the software in any way, as to create elements that can be part of an export and are importable into the distributed source code (this software) under the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. Therefore any dataset must be able to be part of an export and must be able to be imported into this software that must be made available under the terms of GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or
```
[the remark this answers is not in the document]
anonym-HPI: @Nils1729 Now better? If yes, please mark this as resolved.

LICENSE-README.md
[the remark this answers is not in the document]
anonym-HPI: Is this resolved?

README.md
anonym-HPI: @Greenscreen23 Are you ok with the text added? I am not sure about the text with images and datasets.

lukasrad02:
> @lukasrad02 You mentioned noticing on third-libraries.

I'm not an expert for licenses. Especially, I'm not sure whether we have to watch out for the licenses of indirect dependencies, but to be safe, I've checked them with [nlf](https://www.npmjs.com/package/nlf). The licenses found (for the root directory, `frontend`, `shared`, and `backend`) are

> AFL-2.1 OR BSD-3-Clause, Apache-2.0 AND MIT, BSD-3-Clause OR GPL-2.0, MIT AND BSD-3-Clause, MIT AND Zlib, MIT OR CC0-1.0, 0BSD, Apache, Apache-2.0, BSD, BSD-2-Clause, BSD-3-Clause, CC-BY-3.0, CC-BY-4.0, CC0-1.0, GPL, ISC, LGPL, MIT, Public Domain, Python-2.0, Unlicense

From what I know, most of this should not be critical as the licenses have no copyleft.

GPL and LGPL have a copyleft, I'm not sure whether they allow usage in an AGPL project, since on the one hand, AGPL is more strict, on the other hand, AGPL does not restrict the freedom of the users.

However, LGPL only occurs as part of `spdx-expression-parse` (licensed under MIT according to npmjs.com). GPL was found for `argparse` (licensed under Python-2.0 according to npmjs.com), `fraction.js`(licensed under MIT according to npmjs.com), and `node-forge` (which allows BSD-3-Clause as well as GPL), so those two licenses do not seem to be a problem.

What we should keep in mind is that (if I'm not wrong) some licenses require mentioning the usage of the licensed library in the software.

anonym-HPI:
> > @lukasrad02 You mentioned noticing on third-libraries.
> 
> I'm not an expert for licenses. Especially, I'm not sure whether we have to watch out for the licenses of indirect dependencies, but to be safe, I've checked them with [nlf](https://www.npmjs.com/package/nlf). The licenses found (for the root directory, `frontend`, `shared`, and `backend`) are
> 
> > AFL-2.1 OR BSD-3-Clause, Apache-2.0 AND MIT, BSD-3-Clause OR GPL-2.0, MIT AND BSD-3-Clause, MIT AND Zlib, MIT OR CC0-1.0, 0BSD, Apache, Apache-2.0, BSD, BSD-2-Clause, BSD-3-Clause, CC-BY-3.0, CC-BY-4.0, CC0-1.0, GPL, ISC, LGPL, MIT, Public Domain, Python-2.0, Unlicense
> 
> From what I know, most of this should not be critical as the licenses have no copyleft.
> 
> GPL and LGPL have a copyleft, I'm not sure whether they allow usage in an AGPL project, since on the one hand, AGPL is more strict, on the other hand, AGPL does not restrict the freedom of the users.
> 
> However, LGPL only occurs as part of `spdx-expression-parse` (licensed under MIT according to npmjs.com). GPL was found for `argparse` (licensed under Python-2.0 according to npmjs.com), `fraction.js`(licensed under MIT according to npmjs.com), and `node-forge` (which allows BSD-3-Clause as well as GPL), so those two licenses do not seem to be a problem.
> 
> What we should keep in mind is that (if I'm not wrong) some licenses require mentioning the usage of the licensed library in the software.

We have probably no problem using libraries, agree. What does GPL or LGPL restrict users, that AGPL doesn't?

The thing with noticing is, do we notice all or only the ones required and where do we notice them? Via the LICENSE-README.md or just README.md or license.html, probably?

lukasrad02:
> What does GPL or LGPL restrict users, that AGPL doesn't?

I was referring to the addition for network software which grants users more freedom as the may retrieve the source code even for network application.

> The thing with noticing is, do we notice all or only the ones required and where do we notice them? Via the LICENSE-README.md or just README.md or license.html, probably?

I think the `license.html` in the assets folder is a good place since its already embedded in the frontend so we do not need to add any buttons or links. I would suggest to simply list all libraries and their licenses to be safe.

lukasrad02:
We should also add the SPDX identifier `AGPL-3.0-or-later` to our `package.json`s

README.md
Greenscreen23: Its probably fine

LICENSE-README.md
```diff
@@ -0,0 +1,22 @@
...
+Another exception to the default licence exists for datasets. Datasets are files that include no new features and just contain different compositions allowed through template classes given in the source code. Datasets must always be in human readable formats (for example csv, JSON, txt). Datasets license must be mentioned in a LICENSE-DATASET.md (or LICENSES-DATASETS.md) file in the respective folder where the dataset is placed. All licence files must state explicitly which file (or files) they refer to, who the authors are and which license applies (or licenses apply). Datasets are not allowed to create any new features or enhance the software in any way, as to create elements that can be part of an export and are importable into the distributed source code (this software) under the [GNU Affero General Public License](LICENSE.md) as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
```
Nils1729, line 8: Is there any particular reason datatsets must be human readable? This would cause trouble if someone wants to support compressed exports. It is quite likely this may be desired considering the current size of exercise exports.

anonym-HPI: > Is there any particular reason datatsets must be human readable? This would cause trouble if someone wants to support compressed exports. It is quite likely this may be desired considering the current size of exercise exports.

The text doesn't state that exports must be human readable, but that datasets themselves need to be.
What could be a problem is that `importable into the distributed source code (this software)`, this could be made more clear that with this the running software resulting of the source code is meant (=this program or so).
But we could also add that exports can be compressed via this software and decompressed by import or so.

Do you think about compression standards as zip/tar or so or do you mean the export being a binary with our own datastructure for more efficient file sizes, not the current `json`?

anonym-HPI:
> AGPL-3.0-or-later

Not sure, if we can do it like that, true our software is this license, but we have exceptions.

Greenscreen23 (approved):
I'm fine with how it is now

LICENSE-README.md
anonym-HPI: Actually the dataset definition could include exports / imports, I commited a change that should make compressioned and decompressioned exports and imports possible.
@Nils1729 are you okay with this change (https://github.com/hpi-sam/digital-fuesim-manv/pull/636/commits/08916c6cd309cce308c6706f62868feb24c11f4e)?

anonym-HPI:
Somebody an idea how to automatically create `frontend/src/assets/source-code/THIRD-PARTY-LICENSE-ACKNOWLEDGEMENTS.json` and pushing it to a PR, if it changed without making the tests broke, as they seem to be (at least you can't see their current status). Thinking about doing it as the last workflow job, but it should be done before docker containers are built, so that they include this file?!

Nils1729:
> Somebody an idea how to automatically create `frontend/src/assets/source-code/THIRD-PARTY-LICENSE-ACKNOWLEDGEMENTS.json` and pushing it to a PR, if it changed without making the tests broke, as they seem to be (at least you can't see their current status). Thinking about doing it as the last workflow job, but it should be done before docker containers are built, so that they include this file?!

I'm not sure I fully understand your question, but maybe [this](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) is what you're observing?

> For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.

At least we had this issue with jobs not running on `release` PRs. We worked around it by creating them as draft and enabling the `pipeline.yml` workflow to run on `mark as ready`.

anonym-HPI:
> 

What is happening can now be seen. The one workflow pushed a commit and now the only workflow that can be seen is `tests-done` is waiting. All other are hidden (as the run for the old commit). What I want to do at best, is to let this workflow run before any other workflow and let the other workflows run on this newly created one. The reason I create this file in a workflow is that otherwise it could be different, depending on machine used (at least one package has linux or windows or smth. representing a windows version in its name). So not easy to check if a person updated this file, otherwise. So I thought just automatically create it.
I could also only create it when `npm run deployment` or `npm run build:deployment` is run, but that would result in the file being in the docker container, but not in the repository itself. Maybe a legally different, as we are not packing it in the repository.

The packing the source code into the docker container results that everyone can easily download the exact branch or the docker container was built from, without any non commited changes. I may write in the `README.md` if this feature would be used for locally changed files, they must be first commited to comply with the license, otherwise they need to link to a public repo or host it otherwise.

lukasrad02:
> What is happening can now be seen. The one workflow pushed a commit and now the only workflow that can be seen is `tests-done` is waiting. All other are hidden (as the run for the old commit).

As Nils mentioned in his comment, this is expected behavior of GitHub and cannot be changed. Actions triggered by workflows (here: the commit that changes the license acknowledgements) can not trigger other workflows (here: the tests).

> What I want to do at best, is to let this workflow run before any other workflow and let the other workflows run on this newly created one. The reason I create this file in a workflow is that otherwise it could be different, depending on machine used (at least one package has linux or windows or smth. representing a windows version in its name). So not easy to check if a person updated this file, otherwise. So I thought just automatically create it.

I don't think that it is a good idea to create this file automatically on each push or pull request, since at least I frequently push commits and continue to work on the same branch locally. So branches would diverge frequently.

I would suggest to include your new job in the "Create Release PR" workflow, since this is already modifying files and creating commits, so the additional commits and files won't matter and this workflow creates its own branch so it does not interfere with commits of other people. 

> The packing the source code into the docker container results that everyone can easily download the exact branch or the docker container was built from, without any non commited changes. I may write in the `README.md` if this feature would be used for locally changed files, they must be first commited to comply with the license, otherwise they need to link to a public repo or host it otherwise.

With my suggestion to update the acknowledgements in the release workflow, the this requirement would be fulfilled for the `main` branch and the docker images except `dfm:dev`.
I think its okay if the acknowledgements might not be up-to-date for the `dev` branch. For the `dfm:dev` docker image, the `deploy-dev` job could include a step that updates the file without committing it, so the acknowledgements on <https://dev.fuesim-manv.de> are up-to-date.

anonym-HPI:
> 

Changed it so it will be done in PR releases and in every docker build.

One thing that could be done is having the third party acknowledgements in some kind of html table, instead of `json`.
`npm-license-crawler` can output `json` or `csv`.
If somebody wants to look at it?

anonym-HPI:
This PR (except the `json` thing, but we can also keep it that way, for the beginning at least) should be ready for final review.

LICENSE-README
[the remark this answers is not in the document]
anonym-HPI: @christianzoellner can this be marked as `resolved`?

LICENSE-README.md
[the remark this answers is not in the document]
anonym-HPI: @Nils1729 Can this be marked as `resolved`?

LICENSE-README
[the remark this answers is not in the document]
anonym-HPI: Or @lukasrad02 can this be marked as `resolved` (you opened it)?

lukasrad02 (commented):
When I run `npm run install:all` on your branch (to make sure I have all packages, like the new `npm-license-crawler` installed), I get a lot of changes in the `package-lock.json` files.

Did you make sure that your lockfiles are up-to-date by running `npm run install:all` using the latest npm version?

anonym-HPI:
> When I run `npm run install:all` on your branch (to make sure I have all packages, like the new `npm-license-crawler` installed), I get a lot of changes in the `package-lock.json` files.
> 
> Did you make sure that your lockfiles are up-to-date by running `npm run install:all` using the latest npm version?

Haven't done `npm run install:all` in this PR, have done it now locally, seems to also update many packages.
I am not sure this should be part of this PR or more part of a release, where packages are automatically updated.
But I can also see that they are many new packages, when I run it, but the only thing I installed should be the crawler.

lukasrad02:
> > When I run `npm run install:all` on your branch (to make sure I have all packages, like the new `npm-license-crawler` installed), I get a lot of changes in the `package-lock.json` files.
> > Did you make sure that your lockfiles are up-to-date by running `npm run install:all` using the latest npm version?
> 
> Haven't done `npm run install:all` in this PR, have done it now locally, seems to also update many packages. I am not sure this should be part of this PR or more part of a release, where packages are automatically updated. But I can also see that they are many new packages, when I run it, but the only thing I installed should be the crawler.

At least on `dev`, the lock files are in sync with what should be installed (i.e., running `npm run install:all` causes no changes to any file tracked by git). I would like to keep it like this, because out-of-date lockfiles could cause trouble during development, so updating the lock files only in the release process won't suffice.

It also seems like your PR introduces some undesired changes to the lockfiles. For example, all changes to the `package-lock.json` in `shared/` get reverted when I run `npm run install:all` on your branch. Since you've only installed a package to the root `package.json`, your PR should not introduce any changes to the lock files of other branches. Therefore, please clean this up by running `npm run install:all`.

anonym-HPI:
> > > When I run `npm run install:all` on your branch (to make sure I have all packages, like the new `npm-license-crawler` installed), I get a lot of changes in the `package-lock.json` files.
> > > Did you make sure that your lockfiles are up-to-date by running `npm run install:all` using the latest npm version?
> > 
> > 
> > Haven't done `npm run install:all` in this PR, have done it now locally, seems to also update many packages. I am not sure this should be part of this PR or more part of a release, where packages are automatically updated. But I can also see that they are many new packages, when I run it, but the only thing I installed should be the crawler.
> 
> At least on `dev`, the lock files are in sync with what should be installed (i.e., running `npm run install:all` causes no changes to any file tracked by git). I would like to keep it like this, because out-of-date lockfiles could cause trouble during development, so updating the lock files only in the release process won't suffice.
> 
> It also seems like your PR introduces some undesired changes to the lockfiles. For example, all changes to the `package-lock.json` in `shared/` get reverted when I run `npm run install:all` on your branch. Since you've only installed a package to the root `package.json`, your PR should not introduce any changes to the lock files of other branches. Therefore, please clean this up by running `npm run install:all`.

@lukasrad02 
I get many new packages, when I run `npm run install:all` on the newest `dev` branch, this doesn't seem right. It also tries to install a lot of packages. I don't know why, maybe because I am using npm 7.24.2, instead of at least version 8.
Thing is I shouldn't have npm 7.24.2 installed, same happens via WSL. But it results in the lock files jumping back to `lockfileVersion` 2 instead of 3. Which I then change manually back to 3. 

The only new package I want to install in the root is `npm-license-crawler`.
Could you try to push a fix to this seemingly mess?

anonym-HPI:
> > > When I run `npm run install:all` on your branch (to make sure I have all packages, like the new `npm-license-crawler` installed), I get a lot of changes in the `package-lock.json` files.
> > > Did you make sure that your lockfiles are up-to-date by running `npm run install:all` using the latest npm version?
> > 
> > 
> > Haven't done `npm run install:all` in this PR, have done it now locally, seems to also update many packages. I am not sure this should be part of this PR or more part of a release, where packages are automatically updated. But I can also see that they are many new packages, when I run it, but the only thing I installed should be the crawler.
> 
> At least on `dev`, the lock files are in sync with what should be installed (i.e., running `npm run install:all` causes no changes to any file tracked by git). I would like to keep it like this, because out-of-date lockfiles could cause trouble during development, so updating the lock files only in the release process won't suffice.
> 
> It also seems like your PR introduces some undesired changes to the lockfiles. For example, all changes to the `package-lock.json` in `shared/` get reverted when I run `npm run install:all` on your branch. Since you've only installed a package to the root `package.json`, your PR should not introduce any changes to the lock files of other branches. Therefore, please clean this up by running `npm run install:all`.

@lukasrad02 Do you still encounter problems or is all fine now with the lockfiles?

lukasrad02:
> @lukasrad02 Do you still encounter problems or is all fine now with the lockfiles?

The lockfiles seem to be fine now.

However, I've noticed that the `npm-license-crawler` hasn't been updated for about 4 years and introduces a couple of deprecated dependencies:

```
npm WARN deprecated readdir-scoped-modules@1.1.0: This functionality has been moved to @npmcli/fs
npm WARN deprecated natives@1.1.6: This module relies on Node.js's internals and will break at some point. Do not use it, and update to graceful-fs@4.x.
npm WARN deprecated jquery-extend@2.0.3: this package has been deprecated
npm WARN deprecated mkdirp@0.3.5: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
```

Could you please check whether we can achieve the same result with other packages so we don't have lots of deprecated dependencies in our project?

anonym-HPI:
> > @lukasrad02 Do you still encounter problems or is all fine now with the lockfiles?
> 
> The lockfiles seem to be fine now.
> 
> However, I've noticed that the `npm-license-crawler` hasn't been updated for about 4 years and introduces a couple of deprecated dependencies:
> 
> ```
> npm WARN deprecated readdir-scoped-modules@1.1.0: This functionality has been moved to @npmcli/fs
> npm WARN deprecated natives@1.1.6: This module relies on Node.js's internals and will break at some point. Do not use it, and update to graceful-fs@4.x.
> npm WARN deprecated jquery-extend@2.0.3: this package has been deprecated
> npm WARN deprecated mkdirp@0.3.5: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
> ```
> 
> Could you please check whether we can achieve the same result with other packages so we don't have lots of deprecated dependencies in our project?

I updated to a newer package, problem is this packages seems to be not widely used (https://www.npmjs.com/package/@weichwarenprojekt/license-reporter), has only three stars on github and no forks. Not sure about trusting this package or how long it will last.

The problem is, that we have not only packages installed in the root folder, but also in subfolders. It would be easy to create with a more known package a file for each folder, but combining them is also possible, but something I tried to avoid.
I could try something like:
https://www.npmjs.com/package/license-report
and join the jsons together.

What are your thoughts?

lukasrad02:
> What are your thoughts?

I do not have a strong opinion on that. Just use the tool that best meets your requirements.

anonym-HPI:
> > What are your thoughts?
> 
> I do not have a strong opinion on that. Just use the tool that best meets your requirements.

I looked into the creators of the plugin and it seems to be a german company, which is a few years in business, so less of chance to be some kind of scam/malware as a plugin of some anonymous person.

I added some commits, e.g. pr template, added the third party licensing as a json and html file (with a back button, which has no styling, could add it, but would leave it as is, if no complaints are made in this PR) and linked them in the license.html file.

I am not sure about putting the json, html file of third party acknowledgements into the repo aswell as the zip file of the repo. As in the README.md is already written that you need to run `npm run licensing:all` (which actually is run, when using `npm run build:deployment`, so maybe mention this?!).
These three files take up around 6MB of space and if they are already there, maybe they don't get updated correctly?
Docker includes them automatically (see Dockerfile).

lukasrad02:
> I am not sure about putting the json, html file of third party acknowledgements into the repo aswell as the zip file of the repo.

I am strictly against putting files in the repo that can be built from the code that is already part of the repo.

package.json
```diff
@@ -25,15 +25,22 @@
...
+        "merge-coverage": "mkdir coverage && cp backend/coverage/coverage-final.json coverage/backend.json && cp frontend/coverage/coverage-final.json coverage/frontend.json && cp shared/coverage/coverage-final.json coverage/shared.json && nyc merge coverage coverage/merged/coverage.json && nyc report -t coverage/merged --report-dir coverage/reports --reporter=html --reporter=cobertura --reporter=text --reporter=text-summary",
+        "licensing:third-party-licenses": "npm-run-all -s -n -c licensing:third-party-licenses:json licensing:third-party-licenses:html",
+        "licensing:third-party-licenses:json": "license-reporter --output frontend/src/assets/source-code/THIRD-PARTY-LICENSE-ACKNOWLEDGEMENTS.json",
+        "licensing:third-party-licenses:html": "node third-party-license-json-to-html.cjs",
+        "licensing:source-code-zip-file": "git archive --output=./frontend/src/assets/source-code/digital-fuesim-manv.zip --format=zip HEAD",
+        "licensing:all": "npm-run-all -s -n -c licensing:third-party-licenses licensing:source-code-zip-file"
```
lukasrad02, line 33: `npm-run-all` does not seem to be part of a normal Node.js+npm installation. Could you please add it to the devDependencies?

By the way: The `package-lock.json` of this branch does not match the installed packages. On the first glance, it seemed like this way only the `npm-run-all` package wich was part of your lockfile, but not the `package.json` (maybe because you've used `npx` once?), but please make sure the lockfile is up-to-date with the packages from the `package.json` after adding the missing package.

.github/pull_request_template.md
```diff
@@ -4,3 +4,4 @@ Please make sure to fulfil the following conditions before marking this PR ready
 
 -   [ ] If this PR adds or changes features or fixes bugs, this has been added to the changelog
 -   [ ] If this PR adds new actions or other ways to alter the state, [test scenarios](https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios) have been added
+-   [ ] I have the right to the code I commited in this Pull Request (this includes that I don't use any AI tools that violate or could violate the license [LICENSE-README.md](LICENSE-README.md) of this project or copied code from e.g. a forum that falls under an incompatible license and if compatible have given credit in the correct way).
```
lukasrad02, line 7:
```suggestion
-   [ ] I have the right to license the code submitted with this Pull Request (i.e., this is my own code or code licensed under a license compatible to AGPL v3.0). In case of using third party code, I have given appropriate credit.
```

I think this a bit more concise without being as oddly specific (regarding AI tools or forums) as the current wording.

CHANGELOG.md
```diff
@@ -123,6 +127,8 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 -   Loading or unloading a vehicle from its popup does not close the popup automatically.
 -   Patient treatment sorts the patients by their UUID before triaging or assigning treatments in simulated regions. This way, the order of triage does not depend on the order the patients have been added.
 -   The load all into vehicle button in the vehicle popup is now visible to trainers only.
+-   the software includes a file with third party acknoledgments, a license.html and the docker container will include a git archive of the source code downloadable in zip format
+-   new npm run commands for licensing (third party acknowledgements) and creating source code archive
```
lukasrad02, line 131: Could you please move this up to the *unreleased* section

LICENSE-README.md
```diff
@@ -0,0 +1,22 @@
+Copyright (c) 2023 See the [README.md](README.md#contributors) for authors/contributors (bottom of the document).
+
+This software is a digital implementation of the "FüSim MANV"/"FueSim MANV" (Führungssimulation Massenanfall von Verletzen), a German paper-based simulation game for training emergency medical services leadership personnel in managing Mass Casualty Incidents.
+
+The software is licensed with the [GNU Affero General Public License](LICENSE.md) as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
```
lukasrad02, line 5:
```suggestion
The software is licensed under the [GNU Affero General Public License](LICENSE.md) as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
```

third-party-license-json-to-html.cjs
```diff
@@ -0,0 +1,70 @@
...
+let table_header = json2html.transform(data[0], template_table_header);
+let table_body = json2html.transform(data, template_table_body);
+
+let header =
+    '<!DOCTYPE html>' +
+    '<html lang="en">\n' +
+    '<head><title>Third Party Licenses</title></head>';
+let body =
+    '<input type="button" value="Zur&uuml;ck" onclick="history.back()">' +
+    '<h1>Third Party Licenses</h1><br><table id="my_table" style="width:100%">\n<thead>' +
```
lukasrad02, line 39:
```suggestion
    '<h1>Third Party Licenses</h1><br><table>\n<thead>' +
```

The ID is not used and the rules in the `style` attribute are redundant to the ones in your `<style>` element.

third-party-license-json-to-html.cjs
```diff
@@ -0,0 +1,70 @@
...
+                background: #dddddd;
+            }
+            tr:nth-child(even) td {
+                border: 1px solid white;
+            }
+            tr:nth-child(odd) td {
+                border: 1px solid #dddddd;
+            }
+        </style>`;
+let html = header + style + body + '</html>\n';
```
lukasrad02, line 63: Elements are not meant to be direct children of the `html` element. Could you please include the `style` in the `head`?

lukasrad02 (changes requested):

anonym-HPI:
> 

Have removed them and also the workflow, but created a `README.md` file there, so the folder stays in the repo. Dockerfile already creates these files and `npm run deployment`, too.

.github/pull_request_template.md
anonym-HPI: I put AI tools into it, as many people probably don't know the problems that can occur when they for example use Githubs copilot (which may violate licenses). Of course our license is compatible with many licenses, but with such tools you probably won't know the source and therefore the attributions are missing.
Same with forums like stackoverflow, which also have corresponding licenses (which should be compatible mostly with ours, but still need attribution).

Also if I am thinking about it, I am not sure if we need to also include attributions in the html file (like scraping through every code file and looking where something like stackoverflow is mentioned). We probably don't "need" to, as e.g. GPL code doesn't need to attributed when it is sent over a network only, which is common for a website/webapp like ours. I am not a lawyer, so the whole third party acknowledgements could be just a nice gesture.
We may put some hint that some part of the code is inspired or copied from forums, such as stackoverflow on the top of the html or so?!

third-party-license-json-to-html.cjs
anonym-HPI: Removed

CHANGELOG.md
anonym-HPI: Oh, thanks, I have missed this, I have moved not all to the _unreleased_ section.

package.json
anonym-HPI: @lukasrad02 I have installed it, is something else missing?

lukasrad02: Looks good now. There are no missing packages and the lockfiles are up-to-date

lukasrad02:
> Have removed them and also the workflow

Thank you! Could you please also add these files to the gitignore so that they won't be committed by accident?

> but created a `README.md` file there, so the folder stays in the repo

You could use a `.gitkeep` file instead. This would convey the message from the current README just by its name.

lukasrad02:
I've just taken a look into the generated HTML file and wondered whether we might want to preserve whitespaces for the license text column. For example, this is how the first row currently looks like (cropped):

image

While the license should look like this (cropped):

image

The only thing I've changed for the second screenshot was to add the following rule to the stylesheets:

```css
td:last-child {white-space:pre;}
```

I think this makes many licenses much more readable. However, it has a huge downside: Some license texts do not contain line breaks on their own, so the license text column gets extremely large (multiple screen widths on my laptop), causing it to overflow.

One could, of course, restrict the with of this column and add scroll bars only to those cells that are to large, but before tinkering around with CSS rules for hours, I'd like to hear your opinion on this.

.github/pull_request_template.md
lukasrad02: > I put AI tools into it, as many people probably don't know the problems that can occur when they for example use Githubs copilot (which may violate licenses). [...] Same with forums like stackoverflow

I understand your motivation, but on the other hand, if we mention some sources explicitly, it could seem that other sources that are not mentioned are okay. And problems with AI tools, forums etc. are not specific to our project but apply to all sorts of (open source) software with any license, so I don't think this is something that we have to mention explicitly.

> Also if I am thinking about it, I am not sure if we need to also include attributions in the html file (like scraping through every code file and looking where something like stackoverflow is mentioned).

I think we should do this. Searching for URLs (e.g. with `https?://` regex) in our code files could be a good way to find such third party snippets might be an easy way to find them, as all the snippets I've encountered so far had their source in a comment above.

> We may put some hint that some part of the code is inspired or copied from forums, such as stackoverflow on the top of the html or so?!

I think for the typical extent of code from forums it would be fine to do it like this and keep the copyright notices in the comments of the code as a detailed reference.

anonym-HPI: You are right (I would also add it applies probably even more to non open source code).
I can try to find a way to extract these and add them to the table.

anonym-HPI:
> I've just taken a look into the generated HTML file and wondered whether we might want to preserve whitespaces for the license text column. For example, this is how the first row currently looks like (cropped):
> 
> image
> 
> While the license should look like this (cropped):
> 
> image
> 
> The only thing I've changed for the second screenshot was to add the following rule to the stylesheets:
> 
> ```css
> td:last-child {white-space:pre;}
> ```
> 
> I think this makes many licenses much more readable. However, it has a huge downside: Some license texts do not contain line breaks on their own, so the license text column gets extremely large (multiple screen widths on my laptop), causing it to overflow.
> 
> One could, of course, restrict the with of this column and add scroll bars only to those cells that are to large, but before tinkering around with CSS rules for hours, I'd like to hear your opinion on this.

It would be better to have licenses be in a readable format, but did I understand correctly that when making a max width it would create scroll bars inside rows? I would say we should avoid that.
One scroll bar for the whole page is enough.
If there is a way (and I think CSS allows for it) to disable these and just let these license text be like they are right now, this would probably an improvement to licenses that have it and don't touch the others.
What do you think?

anonym-HPI:
> > Have removed them and also the workflow
> 
> Thank you! Could you please also add these files to the gitignore so that they won't be committed by accident?
> 
> > but created a `README.md` file there, so the folder stays in the repo
> 
> You could use a `.gitkeep` file instead. This would convey the message from the current README just by its name.

changed it

lukasrad02:
> But did I understand correctly that when making a max width it would create scroll bars inside rows?

Yes, if we would specify a max width for the column, there would be horizontal scroll bars in every table cell that is larger than this max width. (Depending on the `overflow` rule, it could also just be cropped or grow out of the table and cross the border, but this is definitely something we don't want either.)

> I would say we should avoid that. One scroll bar for the whole page is enough.

If you prefer this behavior, just go with the CSS rule from my first comment.

> If there is a way (and I think CSS allows for it) to disable these and just let these license text be like they are right now, this would probably an improvement to licenses that have it and don't touch the others. What do you think?

I'm sorry, but I do not understand what you mean. Do you want to differentiate between different license texts (since you're saying "and don't touch the others")? If yes, how would this differentiation look like?

anonym-HPI:
> > But did I understand correctly that when making a max width it would create scroll bars inside rows?
> 
> Yes, if we would specify a max width for the column, there would be horizontal scroll bars in every table cell that is larger than this max width. (Depending on the `overflow` rule, it could also just be cropped or grow out of the table and cross the border, but this is definitely something we don't want either.)
> 
> > I would say we should avoid that. One scroll bar for the whole page is enough.
> 
> If you prefer this behavior, just go with the CSS rule from my first comment.
> 
> > If there is a way (and I think CSS allows for it) to disable these and just let these license text be like they are right now, this would probably an improvement to licenses that have it and don't touch the others. What do you think?
> 
> I'm sorry, but I do not understand what you mean. Do you want to differentiate between different license texts (since you're saying "and don't touch the others")? If yes, how would this differentiation look like?

Your improvement is good, with " not touching the others" I mean having your improvement for licenses that line breaks plus not having "the other" have scroll bars. Best of both, your improvement and no worsening.
It may not be possible if there is no CSS way of solving this, as we can't differantiate between the licenses with proper lines and without easily.

.github/pull_request_template.md
anonym-HPI: @lukasrad02 I looked at trying to use https:// finding, but it resulted in a lot of files (even when only searchin in the `src` folders). They would be many false positives, as there are many "this is a quick fix, see the following issue" or so.
Sometimes `inspired by` is used, but not always, sometimes `see` is used, but also not in a way that would be copyright related.
There are multiple ways to solve this.
1. We don't include them (which shouldn't be a problem legally, as long as it isn't a license similar to AGPL that also counts sending code over the network as distribution). But the same applies probably to all third-party notices right now. They are just like a nice gesture or some kind of thank you page.
2. Going through every `https://` and looking manually if they fall under copyright and rewrite it to something like `inspired by` or add some other wording that can be found.
3. Just include a list of all lines that include an `https://` in it

The second solution would need a bit of work and would mean that in the future everyone needs to follow it, so some guideline also, maybe.

I am using the following filtering right now in VSCode:
search: `https://`
replace: empty
files to include: `src/**`
files to exclude: `node_modules, dist, package-lock.json, package.json, THIRD-PARTY-LICENSE-ACKNOWLEDGEMENTS.html, THIRD-PARTY-LICENSE-ACKNOWLEDGEMENTS.json, *.md`

resulting in 112 results in 52 files.

anonym-HPI:
@lukasrad02 I used now `white-sprace:pre-wrap` which fixes these very long lines and have added a list with third party inspired by or copied from list. This list probably includes more links than necessary, e.g. documentation links (but not all, if I remember, at least one I kept out). This is not the prettiest way to solve this, as you now need to manually update it. 
If you have an idea for a tool that finds urls in the source code and automatically puts them in a list (e.g. json or html), this would be great. Also the file name is clunky, if anyone has a better naming scheme.

christianzoellner:
Dear team, 
@hpi-sam/bp2021hg1 @hpi-sam/bp2022hg1 

In the last months I got some messages by different parties interested in the digital FüSim: The BABZ already uses it for their transport scenario, the Berliner Feuerwehr wants to use it for internal trainings, and at least one company is looking into offering emergency management trainer courses based on this software.

So I propose that we publish a license soon(ish). 

Are there any important debates left before we merge the license file via this PR?

Best regards,
Christian

anonym-HPI:
> Dear team, @hpi-sam/bp2021hg1 @hpi-sam/bp2022hg1
> 
> In the last months I got some messages by different parties interested in the digital FüSim: The BABZ already uses it for their transport scenario, the Berliner Feuerwehr wants to use it for internal trainings, and at least one company is looking into offering emergency management trainer courses based on this software.
> 
> So I propose that we publish a license soon(ish).
> 
> Are there any important debates left before we merge the license file via this PR?
> 
> Best regards, Christian

If there are no complaints anymore I would suggest pushing it. The dependency problem I suggest gets fixed in another PR, all PRs seem to have dependency check(s) failing.

lukasrad02:
> Are there any important debates left before we merge the license file via this PR?

From my side, there are no concerns regarding the license itself. The only discussions left were about processes and tools.

Unfortunately, I do not have time to take a closer look at these things due to the holidays. However, they are rather easy to change later on, so I'll approve this PR.

lukasrad02 (approved):
Approval according to my last comment

Nils1729 (approved):

anonym-HPI:
@Dassderdie @ClFeSc Do you have any comments or do you approve this?

Dassderdie (approved):

anonym-HPI:
Everyone has approved, that contributed code, as far as I can see over [here](https://github.com/hpi-sam/digital-fuesim-manv/graphs/contributors?from=2021-11-07&to=2023-11-06&type=c) (@hpistudent72 gave permission, also for his bachelor thesis code)
I have changed the template for Pull Requests a bit (see last commit) in short:
Added DCO description and make it more clear that code published in the Pull Request is/needs to be under the license of this project to be merged.
We are using DCO robo check from now on and requiring signing off from now on.
What is DCO, see [here](https://github.com/dcoapp/app#how-it-works).

The DCO robo check should enforce sign offs from now on, but I also enabled it in the settings of dev and main (we have to see, if this should be disabled, maybe they conflict each other? DCO robo should allow bots without signoffs, not sure about the other setting).

Signoff check will be mandatory after this Pull Request is merged.

PR #642, opened 2023-02-01
Refactor Start Point Literals

### LINKED ISSUE #637: Make type properties distinctable
#616 added a `type` property to elements that are part of the state. The string literals used in this PR overlap with the string literals for the types in `shared/src/models/utils/start-points.ts`.

Although there should not be a situation in which we have to differentiate between an element from #616 and a `StartPoint` from the file mentioned above, the `type` properties should be unique to avoid confusion.

benn02:
closes #637

### DIFF
--- a/shared/src/models/utils/start-points.ts
+++ b/shared/src/models/utils/start-points.ts
@@ -7,8 +7,8 @@ import { getCreate } from './get-create';
 export type StartPoint = AlarmGroupStartPoint | TransferStartPoint;
 
 export class TransferStartPoint {
-    @IsValue('transferPoint' as const)
-    public readonly type = 'transferPoint';
+    @IsValue('transferStartPoint' as const)
+    public readonly type = 'transferStartPoint';
 
     @IsUUID(4, uuidValidationOptions)
     public readonly transferPointId: UUID;
@@ -24,8 +24,8 @@ export class TransferStartPoint {
 }
 
 export class AlarmGroupStartPoint {
-    @IsValue('alarmGroup' as const)
-    public readonly type = 'alarmGroup';
+    @IsValue('alarmGroupStartPoint' as const)
+    public readonly type = 'alarmGroupStartPoint';
 
     @IsString()
     public readonly alarmGroupName: string;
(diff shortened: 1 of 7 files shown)


shared/src/state-migrations/19-rename-start-point-types.ts
```diff
@@ -0,0 +1,143 @@
...
+                                          | 'alarmGroup'
+                                          | 'alarmGroupStartPoint'
+                                          | 'transferPoint'
+                                          | 'transferStartPoint';
+                                  };
+                              };
+                          }
+                        | { type: 'coordinates' }
+                        | { type: 'simulatedRegion' }
+                        | { type: 'vehicle' };
```
lukasrad02, line 47: I think you can omit these other possible position types on materials, vehicles and personnel as they're not relevant for the migration.

shared/src/models/utils/start-points.ts
```diff
@@ -7,8 +7,8 @@ import { getCreate } from './get-create';
 export type StartPoint = AlarmGroupStartPoint | TransferStartPoint;
```
lukasrad02, line 7: The input property `startPoint` in `frontend/src/app/pages/exercises/exercise/shared/transfer-overview/start-point-name/start-point-name.component.ts` could use this union type as well.

lukasrad02 (changes requested):

shared/src/state-migrations/19-rename-start-point-types.ts
benn02: I didn't want to put position: any as a possibility and if I put {type: Exclude<any,'transfer'}> as a possibility for position to put the 3 cases in one it doesn't work either, so this is the only way I could think of w/o putting any as an option for the transfer Position as well.

benn02: If you know a better way, I am open though

lukasrad02: You don't have to specify the other possible types at all since you don't care about them. Just use

```ts
[materialId: UUID]: {
    position: {
        type: 'transfer',
        transfer: {
            startPoint: {
                type:
                    | 'alarmGroup'
                    | 'alarmGroupStartPoint'
                    | 'transferPoint'
                    |  'transferStartPoint';
            };
        };
    }
}
```

Or am I missing something here?

benn02: Hmm, It works which I think is weird because now I am saying that the type is always transfer AND I'm checking it in the if statement later. In my mind, this should lead to a Condition is always true warning, but it doesn't.

shared/src/models/utils/start-points.ts
benn02: Done

lukasrad02 (approved):

PR #649, opened 2023-02-02
Make Map movable on Drag with Pop Up

### LINKED ISSUE #648: Make Map movable when Pop-up is open
Currently, any Click on the Map closes any Pop-up. This is not optimal because sometimes Pop-ups aren't fully readable when the Screen is too small, or they are opened in obscure locations. To fix this in a simple way, It will be allowed to move the map (drag not single click) while keeping pop-ups open.

benn02:
closes #648

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
@@ -444,10 +444,6 @@ export class OlMapManager {
             ) {
                 return;
             }
-            this.changePopup$.next(undefined);
-        });
-        this.olMap.getView().on(['change:resolution', 'change:center'], () => {
-            this.changePopup$.next(undefined);
         });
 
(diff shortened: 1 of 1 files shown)


frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
```diff
@@ -436,18 +436,12 @@ export class OlMapManager {
...
                 event.coordinate[0] === event.startCoordinate[0] &&
                 event.coordinate[1] === event.startCoordinate[1]
             ) {
-                return;
             }
-            this.changePopup$.next(undefined);
-        });
-        this.olMap.getView().on(['change:resolution', 'change:center'], () => {
-            this.changePopup$.next(undefined);
         });
```
lukasrad02, line 445: You can remove the whole event handler, since it doesn't do anything after your change. Keep in mind that you can remove the `translateInteraction` parameter from the `registerPopupTriggers` function.

Nils1729 (approved):

PR #654, opened 2023-02-07
Cypress tests

Greenscreen23:
Todo for testing:
- [x] Transfer vehicles to hospitals (requires #671)
- [x] Modify Patients on the map
- [x] Rename Vierports and simulated regions
- [x] Connect and manage transfer points
- [x] Pause transfer
- [x] Transfer vehicles between transfer points
- [x] Load and unload vehicles (requires #671)
- [ ] ~~Test removal~~ (requires #671)


Greenscreen23:
This setup currently also works fine on node 18, maybe its because we updated cypress.

frontend/cypress/support/commands.ts
```diff
@@ -41,8 +41,98 @@
...
+export function socket() {
+    return cy.window().its('socket');
+}
+
+export function registerSocketListener() {
+    performedActionsStack.length = 0;
+    cy.socket().then((s) =>
+        s.on('performAction', (action: any) =>
+            performedActionsStack.push(action)
+        )
```
Dassderdie, line 84: I think this would break in the future.

At the moment, a performed action is sent to all clients in an exercise, including the client that initially proposed it. The proposing client gets additionally a positive response to its proposal-request. This complicates the `OptimisticActionHandler` as we have to identify the `performAction`, that corresponds to our successful proposal (which means we have to compare actions by value). This also leads to a special case where the [`optimisticUpdateHandler` behaves unitntuivively](https://github.dev/hpi-sam/digital-fuesim-manv/blob/b458be325d600ea5f5456b5c743695b972c0a553/frontend/src/app/core/optimistic-action-handler.spec.ts#L290-L291).

This could be fixed by excluding the proposing client from the corresponding `performAction` already on the server. (I initially didn't want to do this, as I thought we would use Rooms from `socket.io`, which have a `broadcast`-method, which I thought could be more performant - but we don't use these rooms, and the performance benefit would be negligible/non-existent).


Alternatives:
* Add these actions directly to the window-object in the `proposeAction` method in the `ExerciseService`
* Maybe use the Ngrx-Store and its reducers somehow?

frontend/cypress/support/commands.ts
```diff
@@ -41,8 +41,98 @@
...
+            .trigger('mousemove', { force: true, ...offset })
+            .click({ force: true })
+    );
+}
+
+export function storeState() {
+    return cy
+        .window()
+        .its('store')
+        .then((store: any) => store.source._value.application);
```
Dassderdie, line 72:
```suggestion
        .then((store: any) => getStateSnapshot(store, selectApplication));
```

Or can't/shouldn't cypress import functions from the application?

We have  or something like this.

frontend/cypress/support/commands.ts
```diff
@@ -41,8 +41,98 @@
...
+            /**
+             * We have to force the click here, as the img
+             * of the element being dragged is in front of the map
+             **/
+            .trigger('mousemove', { force: true, ...offset })
+            .click({ force: true })
+    );
+}
+
+export function storeState() {
```
Dassderdie, line 68:
```suggestion
export function getStoreState() {
```

frontend/cypress/e2e/alarm-group.exercise.cy.ts
```diff
@@ -0,0 +1,128 @@
...
+            .then((a) => a.at(-1))
+            .its('type')
+            .should('eq', '[AlarmGroup] Remove AlarmGroupVehicle');
+
+        cy.storeState()
+            .its('exerciseState')
+            .its('alarmGroups')
+            .then((alarmGroups) => alarmGroups[Object.keys(alarmGroups)[0]!])
+            .its('alarmGroupVehicles')
+            .then(alarmGroupVehicles => Object.keys(alarmGroupVehicles))
```
Dassderdie:
```suggestion
            .then(alarmGroupVehicles => isEmpty(alarmGroupVehicles))
```

Dassderdie: Or directly via cypress, if it has such an comparator...

frontend/cypress/support/commands.ts
Dassderdie: Or something similar. The problem is that `store` is also a verb.

Dassderdie (commented):
Some thoughts on it.
Take it with a grain of salt as I don't have any experience with cypress, yet.

And don't forget to update the documentation in the root- and the frontend `Readme`s.

frontend/cypress/e2e/alarm-group.exercise.cy.ts
```diff
@@ -0,0 +1,128 @@
+describe('The alarm group overview on the exercise page', () => {
+    beforeEach(() => {
+        cy.createExercise().joinExerciseAsTrainer();
+        cy.get('[data-cy=trainerToolbarCreationButton]').click();
+        cy.get('[data-cy=trainerToolbarAlarmGroupsButton]').click();
+    });
+
+    afterEach(() => cy.deleteExercise())
```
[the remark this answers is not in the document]
Greenscreen23: Thanks for the hint! Th best practice site had some interesting remarks :)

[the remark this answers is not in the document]
Greenscreen23: I'll just remove them, we'll just have these exercises dangling while development. If this becomes an issue later on, we might have to reset the db and restart the backend while testing.

frontend/cypress/e2e/alarm-group.exercise.cy.ts
```diff
@@ -0,0 +1,135 @@
...
+
+        cy.performedActions()
+            .then((a) => a.at(-1))
+            .its('type')
+            .should('eq', '[AlarmGroup] Remove AlarmGroupVehicle');
+
+        cy.storeState()
+            .its('exerciseState')
+            .its('alarmGroups')
+            .then((alarmGroups) => alarmGroups[Object.keys(alarmGroups)[0]!])
```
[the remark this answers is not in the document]
Greenscreen23: :D Thanks

Greenscreen23:
Another thing we should investigate at some point: Tests seem to be more flaky when being used with a database.

Greenscreen23:
@Dassderdie What do you think of my changes to `frontend/src/app/core/optimistic-action-handler.ts`?
I was looking at the type definitions, and just could not figure out why there were nested Immutables and why everything always needed to extend instead of just being that type.

frontend/src/app/app.component.ts
```diff
@@ -1,8 +1,20 @@
...
     selector: 'app-root',
     templateUrl: './app.component.html',
     styleUrls: ['./app.component.scss'],
 })
-export class AppComponent {}
+export class AppComponent {
+    constructor(private readonly store: Store<AppState>) {
+        const anyWindow = window as any
+        if (anyWindow.Cypress && !environment.production) {
+            anyWindow.store = this.store
```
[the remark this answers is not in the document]
Greenscreen23: I've implemented some helpers in d1d74e066d79cc962218497431dbb7bbe60b1591

frontend/src/app/core/exercise.service.ts
```diff
@@ -64,6 +65,11 @@ export class ExerciseService {
         private readonly store: Store<AppState>,
         private readonly messageService: MessageService
     ) {
+        const anyWindow = window as any
+        if (anyWindow.Cypress && !environment.production) {
```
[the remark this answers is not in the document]
Greenscreen23: Done in d1d74e066d79cc962218497431dbb7bbe60b1591

frontend/cypress/support/commands.ts
Greenscreen23: Thanks for the hint. Listening to both proposed and performed actions is probably a good choice, as we are going to need them for later testing.

I'll also take a look how one could integrate the ngrx store properly into cypress.

frontend/cypress/support/commands.ts
Greenscreen23: I've tried this. The result is that ngrx errors because it can't find angular. But I've replaced it with a version that works and doesn't select these hacky values from the store.

Dassderdie:
> @Dassderdie What do you think of my changes to `frontend/src/app/core/optimistic-action-handler.ts`? I was looking at the type definitions, and just could not figure out why there were nested Immutables and why everything always needed to extend instead of just being that type.
> 
> Is there something that I missed? Everything seems to be working fine (so far)

Without trying it out, I believe your change allows for mutable Action and State types to be passed to the `OptimisticActionHandler`, resulting in its methods returning mutable types.

Though, it could arguably be better not to enforce this immutability via typings and just rely on the comments stating that they must be immutable.

I think I'm fine with it.

Greenscreen23:
> > @Dassderdie What do you think of my changes to `frontend/src/app/core/optimistic-action-handler.ts`? I was looking at the type definitions, and just could not figure out why there were nested Immutables and why everything always needed to extend instead of just being that type.
> > Is there something that I missed? Everything seems to be working fine (so far)
> 
> Without trying it out, I believe your change allows for mutable Action and State types to be passed to the `OptimisticActionHandler`, resulting in its methods returning mutable types.
> 
> Though, it could arguably be better not to enforce this immutability via typings and just rely on the comments stating that they must be immutable.
> 
> I think I'm fine with it.

But `Action` extends `Immutable<JsonObject>`, so it's immutable, right?

Dassderdie:
> > > @Dassderdie What do you think of my changes to `frontend/src/app/core/optimistic-action-handler.ts`? I was looking at the type definitions, and just could not figure out why there were nested Immutables and why everything always needed to extend instead of just being that type.
> > > Is there something that I missed? Everything seems to be working fine (so far)
> > 
> > 
> > Without trying it out, I believe your change allows for mutable Action and State types to be passed to the `OptimisticActionHandler`, resulting in its methods returning mutable types.
> > Though, it could arguably be better not to enforce this immutability via typings and just rely on the comments stating that they must be immutable.
> > I think I'm fine with it.
> 
> But `Action` extends `Immutable<JsonObject>`, so it's immutable, right? I was just confused that `ImmutableAction` was basically `Immutable<Immutable<JsonObject>>`.
> 
> ~I would have liked it best, if we could just remove those generics all together, and just write `Immutable<ExerciseAction>` instead of `Action`, which extends `Immutable<JsonObject>`. But that would completely break your test.~ Nvm, I see where you're going. Lemme rewrite that real quick

This is valid code:
```ts
function a<
    T extends {
        readonly [key: string]: any;
    }
>(x: T): T {
    return T;
}

const x = { a: 3 };
a(x).a = 4;
```

Greenscreen23:
> > But `Action` extends `Immutable<JsonObject>`, so it's immutable, right? I was just confused that `ImmutableAction` was basically `Immutable<Immutable<JsonObject>>`.
> > ~I would have liked it best, if we could just remove those generics all together, and just write `Immutable<ExerciseAction>` instead of `Action`, which extends `Immutable<JsonObject>`. But that would completely break your test.~ Nvm, I see where you're going. Lemme rewrite that real quick
> 
> This is valid code:
> 
> ```ts
> function a<
>     T extends {
>         readonly [key: string]: any;
>     }
> >(x: T): T {
>     return T;
> }
> 
> const x = { a: 3 };
> a(x).a = 4;
> ```

But what if you declare `const x = { a: 3 } as { readonly a: number }`. Is it still valid?

Dassderdie:
> But what if you declare `const x = { a: 3 } as { readonly a: number }`. Is it still valid?

This correctly breaks it.

But my original intention was, that we are unable to pass a mutable `T` in there, as mutable objects would completely break the behavior of the class.

Greenscreen23:
I've also added another commit, does this one fit your needs more?

Dassderdie:
> I've also added another commit. Does this one fit your needs more?

This is now equivalent to the previous code 👍. I used the Generics as reusable type variables throughout the class, but I agree that it was a bit hacky...

Greenscreen23:
> 



> > I've also added another commit. Does this one fit your needs more?
> 
> This is now equivalent to the previous code +1. I used the Generics as reusable type variables throughout the class, but I agree that it was a bit hacky...

My main problem with those generics was the `extends` keyword. As I was exporting values to cypress, I noticed that I  could not place an action into my action array, because `ImmutableAction` could extend `Immutable<JsonObject>`. Is there a way to specify a type alias like `ImmutableAction = Immutable<JsonObject>`, so that is forced to be that object and not a subclass?

Dassderdie:
Sorry, I pressed "Edit" instead of "Quote"...


> My main problem with those generics was the `extends` keyword. As I was exporting values to cypress, I noticed that I  could not place an action into my action array, because `ImmutableAction` could extend `Immutable<JsonObject>` and was therefore not assignable to it. 

Why don't you replace `Immutable<JsonObject>` with `ExerciseAction` in cypress? This would also allow you to profit of the typings in the e2e tests.

> Is there a way to specify a type alias like `ImmutableAction = Immutable<JsonObject>`, so that is forced to be that object and not a subclass?

[Not really.](https://github.com/microsoft/TypeScript/issues/12936) There is a hack where you check that the intersection is empty, but this has other caveats I think.


Btw. the `OptimisticUpdateHandler` was designed to be relatively well encapsulated and separated from the rest of the codebase. Therefore, it might be even better to move the cypress stuff one level up to the ExerciseService.

Greenscreen23:
> Btw. the `OptimisticUpdateHandler` was designed to be relatively well encapsulated and separated from the rest of the codebase. Therefore, it might be even better to move the cypress stuff one level up to the ExerciseService.

I'll take a look at it tomorrow. I already thought about that, but I can't remember why I didn't like it.

frontend/cypress/e2e/alarm-group.exercise.cy.ts
```diff
@@ -0,0 +1,128 @@
+describe('The alarm group overview on the exercise page', () => {
```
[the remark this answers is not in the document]
Greenscreen23: I agree that I might have to reword some of these statements, but I think that _The_ is fine here. The way cypress is printing it to stdout is like this:
_describe string_
    _it string one_
    _it string two_

So it's currently _The alarm group overview on the exercise page ... can create alarm groups_, which is not perfect, but the first thing to update would be the _it string_

frontend/src/app/pages/exercises/exercise/shared/trainer-map-editor/trainer-map-editor.component.html
```diff
@@ -27,6 +27,7 @@ <h5>Ansicht/Transferpunkt</h5>
                                     template: viewportTemplate
                                 })
                             "
+                            data-cy="getViewportDiv"
```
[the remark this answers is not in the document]
Greenscreen23: Done in b469a9a506da5ec1e5f851ca01f25ec7f74873bd

frontend/cypress/e2e/alarm-group.exercise.cy.ts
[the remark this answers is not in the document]
Greenscreen23: Done in f5867a6ee1ea946ff1e887721e86839f9c89fe7b

frontend/cypress/e2e/alarm-group.exercise.cy.ts
```diff
@@ -0,0 +1,135 @@
...
+            .should('eq', '[AlarmGroup] Remove AlarmGroupVehicle');
+
+        cy.storeState()
+            .its('exerciseState')
+            .its('alarmGroups')
+            .then((alarmGroups) => alarmGroups[Object.keys(alarmGroups)[0]!])
+            .its('alarmGroupVehicles')
+            .then((alarmGroupVehicles) => Object.keys(alarmGroupVehicles))
+            .its('length')
+            .should('eq', 0);
```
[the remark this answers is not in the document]
Greenscreen23: I've done a complete overhaul in 067215703e42101eaf227589d1d3a001614d354e

frontend/cypress/e2e/alarm-group.exercise.cy.ts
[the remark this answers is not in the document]
Greenscreen23: I've done some renaming in 09a7deb602a48e42cf7661cf112e9fa6f02e6cda

frontend/src/app/shared/directives/app-save-on-typing.directive.ts
```diff
@@ -42,7 +43,7 @@ export class AppSaveOnTypingDirective implements OnDestroy {
                 }),
                 // Keeping a key (like backspace) pressed for a more than a certain threshold will result in many key presses
                 // The debounceTime should be above that threshold to not register the initial keypress as the first update
-                debounceTime(600),
+                mergeMap(value => isBeingTestedByCypress() ? of(value) : of(value).pipe(debounceTime(600))),
```
Greenscreen23, line 46: Is this the correct way of conditionally piping filters? I'm far from an expert of observables.

Dassderdie: Because `isBeingTestedByCypress()` should not change during the execution of the of the observable, this should suffice:

```ts
pipe(
...,
isBeingTestedByCypress() ? debounceTime(600) : tap(),
...
)
```

Though, I would reconsider whether this is a good idea, as it not just logs stuff to cypress but changes the behavior, making the e2e tests less useful.

Greenscreen23: Alternatively, I could add a `cy.wait(600)` to every assertion if an event from an input field was fired. I'd figured that faster tests are more important than the debounce time.

Greenscreen23: But thank you, that was exactly what I was looking for, I just could not find the _identity_ filter

frontend/src/app/shared/functions/cypress.ts
```diff
@@ -0,0 +1,24 @@
...
+export const setupCypressTestingValues = (
+    apply: (values: Partial<CypressTestingValues>) => void
+) => {
+    const anyWindow = window as any;
+    if (isBeingTestedByCypress()) {
+        if (!anyWindow.cypressTestingValues)
+            anyWindow.cypressTestingValues = {};
+        apply(anyWindow.cypressTestingValues);
+    }
+};
```
Dassderdie, line 24: Instead of an `apply` function, directly pass an `Partial<CypressTestingValues>.
Use `defaults()` from lodash to add it to the window property via `window.cypressTestingValues = defaults(values, window.cypressTestingValues)`.

If the function currently has side effects, one should do it separately from setting the values with an additional check of `isBeingTestedByCypress()`.

This should make this API easier to understand, as it is in this codebase usually discouraged to mutate function arguments.

Greenscreen23: I've changed it in 8c05375cce5b6456dc1c970b594b7c028fd5dd63, do you think this is better now?

frontend/cypress/support/commands.ts
```diff
@@ -41,8 +41,212 @@
...
+
+export function nthElement(subject: Array<unknown>, n: number) {
+    return cy.log(`${n}th element`).wrap(subject.at(n), { log: false });
+}
+
+export function nthLastElement(subject: Array<unknown>, n: number) {
+    return cy
+        .log(`${n}th last element`)
+        .wrap(subject.at(subject.length - (n + 1)), { log: false });
+}
```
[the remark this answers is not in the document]
Dassderdie: If you want to change the log message, you can also do it by checking whether the index is `0`, `-1`, `< 0`, `>= 0`.

frontend/cypress/support/commands.ts
```diff
@@ -41,8 +41,212 @@
...
+export function initializeParticipantSocket() {
+    cy.get('@websocketBaseUrl', { log: false }).then(
+        (websocketBaseUrl: any) => {
+            cy.wrap(io(websocketBaseUrl, { transports: ['websocket'] })).as(
+                'participantSocket'
+            );
+        }
+    );
+
+    cy.get('@participantId', { log: false }).then((participantId: any) => {
```
benn02, line 140: Do you need to use any here? same for Trainer

frontend/cypress/support/commands.ts
```diff
@@ -41,8 +41,212 @@
...
+        cy.get('@participantSocket', { log: false }).then(
+            (participantSocket: any) => {
```
benn02, line 142: Maybe there is a better Type than any

frontend/src/app/core/exercise.service.ts
```diff
@@ -60,6 +61,9 @@ export class ExerciseService {
         SocketResponse
     >;
 
+    private readonly proposedActions: Immutable<ExerciseAction>[] = [];
+    private readonly performedActions: Immutable<ExerciseAction>[] = [];
```
benn02, line 65: I do not konw why do you create those here... They are never read or otherwise used according to VSCode. And I don't feel like that we should have them in that spot just for testing

shared/src/store/action-reducers/simulated-region.ts
```diff
@@ -61,8 +61,8 @@ export class ResizeSimulatedRegionAction implements Action {
 }
 
 export class RenameSimulatedRegionAction implements Action {
-    @IsValue('[SimulatedRegion] Rename simulatedRegion' as const)
```
benn02, line 64: This needs a migration

frontend/cypress/e2e/exercise.cy.ts
```diff
@@ -0,0 +1,680 @@
+describe('A trainer on the exercise page', () => {
```
benn02, line 1: This File looks like WIP with all the comments, so I did not look at it

frontend/cypress/support/commands.ts
Greenscreen23: Yup, I actually need to, as cypress thinks that I'm storing Jquery elements and there is no real cast to these.

frontend/cypress/support/commands.ts
Greenscreen23: Same as above

frontend/src/app/core/exercise.service.ts
Greenscreen23: I forgot to remove them in a refactoring, thanks ^^

Greenscreen23: See 450f945dd0c502eaf53a51e87be8e27b05509508

shared/src/store/action-reducers/simulated-region.ts
Greenscreen23: Done in 721f72b0fa9ffccfbdcfc2af642b3f90476ce1c2
Did I miss anything?

frontend/cypress/support/commands.ts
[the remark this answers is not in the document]
benn02: I think the tests are now failing because the of the new function being different from the ones that it replaced

benn02 (commented):
It now all lgtm (apart from the one unresolved conversation) but since there are still things to come, I will not approve right now

frontend/cypress/support/commands.ts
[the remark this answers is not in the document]
Greenscreen23: The cypress run is currently being killed because its limited to 5 minutes, I'll increase it on the next commit.

Greenscreen23:
Screenshot_20230216_160317
Cypress is using the wrong node version in the pipeline. Maybe its failing because of that?

frontend/cypress/e2e/exercise.cy.ts
```diff
@@ -0,0 +1,851 @@
...
+                    .its('exerciseState')
+                    .its('clients')
+                    .atKey(participantSocketUUID)
+                    .should('have.property', 'viewRestrictedToViewportId');
+            }
+        );
+
+        cy.get('@participantSocket', { log: false }).invoke('disconnect');
+
+        cy.wait(0);
```
benn02, line 790: What does this do?

Greenscreen23: It waits for the socket to disconnect. It's there for a reason, but I don't know if I completely understand why.

frontend/cypress/support/commands.ts
```diff
@@ -41,8 +41,211 @@
...
+    return cy.log('last element').wrap(subject.at(-1), { log: false });
+}
+
+export function atPosition(subject: Array<unknown>, n: number) {
+    if (n < 0) {
+        return cy
+            .log(`${n}th element`)
+            .wrap(subject.at(subject.length + n), { log: false });
+    }
+    return cy.log(`${n}th element`).wrap(subject.at(n), { log: false });
```
Dassderdie, line 238:
```suggestion
    return cy.log(`${n}th element`).wrap(subject.at(n), { log: false });
```

This should be equivalent code.

frontend/cypress/e2e/exercise.cy.ts
```diff
@@ -0,0 +1,897 @@
...
+
+        cy.getState()
+            .its('exerciseState')
+            .its('hospitals')
+            .itsValues()
+            .firstElement()
+            .its('patientIds')
+            .should('not.be.empty');
+
+        cy.wait(150); // Do not trigger a double click
```
benn02, line 489: On my machine it worked to just double click explicitly... that would get rid of the wait

frontend/cypress/e2e/exercise.cy.ts
```diff
@@ -0,0 +1,897 @@
...
+            .its('exerciseState')
+            .its('transferPoints')
+            .itsValues()
+            .firstElement()
+            .should('have.property', 'internalName', 'ABC123');
+
+        cy.get('[data-cy=transferPointPopupCloseButton]').click();
+        cy.dragToMap('[data-cy=draggableTransferPointDiv]');
+
+        cy.wait(150);
```
benn02, line 548: Same here

frontend/cypress/e2e/exercise.cy.ts
Greenscreen23: My problem was that clicking again would trigger a double click. You are saying that if you trigger a double click there, then it works?

benn02: that would then be a double click to end the drop action and a single correct click

benn02: You can try it by trippleclicking on an object. it zoomes in and opens the popup. It zoomes in which is not wanted but not bad either and imo it looks cleaner to have another click than a wait

frontend/cypress/e2e/exercise.cy.ts
Greenscreen23: Fixed in b501980e368668c216da57909d2ea74c919bebe8

frontend/cypress/e2e/exercise.cy.ts
Greenscreen23: Fixed in b501980e368668c216da57909d2ea74c919bebe8

frontend/cypress/support/commands.ts
Greenscreen23: Fixed in b501980e368668c216da57909d2ea74c919bebe8

frontend/cypress/e2e/exercise.cy.ts
Greenscreen23: Edit: Fixed in 483896ce042712d2fa69f3896bca3e1e5fd2d7ec

frontend/cypress/e2e/exercise.cy.ts
Greenscreen23: Edit: Fixed in 483896ce042712d2fa69f3896bca3e1e5fd2d7ec

frontend/cypress/support/commands.ts
Greenscreen23: Edit: Fixed in 2e505aa4abb4a0ec3d9c92e0ddd60eff102818d3

frontend/cypress/e2e/exercise.cy.ts
```diff
@@ -0,0 +1,826 @@
...
+                'have.property',
+                'type',
+                '[Emergency Operation Center] Add Log Entry'
+            );
+
+        cy.getState()
+            .its('exerciseState')
+            .should('have.property', 'currentStatus', 'running');
+
+        cy.wait(1000);
```
benn02, line 679: Oh no a magic number maybe introduce a variable luke commonErrorTimeout such as tickDuration for this

frontend/cypress/e2e/exercise.cy.ts
```diff
@@ -0,0 +1,826 @@
...
+            .firstElement()
+            .its('position')
+            .its('transfer')
+            .should('have.property', 'isPaused', false);
+
+        cy.get('[data-cy=transferOverviewCloseButton]').click();
+        cy.get('[data-cy=trainerToolbarStartButton]').click();
+        cy.get('[data-cy=confirmationModalOkButton]').click();
+
+        cy.wait(1000);
```
benn02, line 645: Magic Number

frontend/cypress/e2e/exercise.cy.ts
Greenscreen23: See f828d327ae552d4dc91bbbcc5989e7b025c9aff4

PR #657, opened 2023-02-07
Adjust nginx configuration for about pages

lukasrad02:
The about pages request the content for the pages from the server and rely on the server to answer either 200 if the file exists or 404 if the file doesn't. If the file doesn't exist, the page should show a hint how to customize this page.

Since the application uses client-side routing, the server serves the `index.html` for all requests where a file with a matching name could not be found. This breaks the functionality of the about pages.

This PR adjusts the configuration of the nginx in the docker container to fix the bug for the about pages while preserving the ability of client-side-routing.

### DIFF
--- a/docker/nginx/locations
+++ b/docker/nginx/locations
@@ -1,12 +1,17 @@
     # Frontend
+    root /usr/local/app/frontend/dist/digital-fuesim-manv;
+
+    location /assets/about {
+        # Requests to this location should not fall back to index.html
+    }
+
     location / {
-        alias   /usr/local/app/frontend/dist/digital-fuesim-manv/;
         try_files $uri $uri/ /index.html;
     }
 
     # API
     location /api {
-		proxy_pass	http://localhost:3201;
+        proxy_pass	http://localhost:3201;
         proxy_http_version 1.1;
         proxy_set_header   Host $host;
         proxy_set_header   X-Forwarded-Host  $host;
@@ -14,10 +19,10 @@
         proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header   X-Forwarded-Proto $scheme;
         include conf.d/upload-limit;
-	}
+    }
 
     # Websocket
-	location /socket.io {
+    location /socket.io {
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "Upgrade";
         proxy_pass http://localhost:3200;
@@ -28,4 +33,4 @@
         proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header   X-Forwarded-Proto $scheme;
         include conf.d/upload-limit;
-	}
+    }
--- a/docker/nginx/locations
+++ b/docker/nginx/locations
@@ -1,10 +1,10 @@
     # Frontend
     root /usr/local/app/frontend/dist/digital-fuesim-manv;
+    error_page 404 /index.html;
+    error_page 403 /index.html;
 
     location /assets/about {
         # Requests to this location should not fall back to index.html
-        error_page 404 /non_existent_page.html;
-        error_page 403 /non_existent_page.html;
     }
 
     location / {
--- a/docker/nginx/locations
+++ b/docker/nginx/locations
@@ -3,6 +3,8 @@
 
     location /assets/about {
         # Requests to this location should not fall back to index.html
+        error_page 404 /non_existent_page.html;
+        error_page 403 /non_existent_page.html;
     }
 
     location / {
--- a/frontend/src/app/pages/about/about-placeholder/about-placeholder.component.html
+++ b/frontend/src/app/pages/about/about-placeholder/about-placeholder.component.html
@@ -3,7 +3,8 @@
     <h2 class="display-4">{{ pageTitle }}</h2>
 
     <p>
-        <a (click)="back($event)" href="#">Zurück zur vorherigen Seite</a>
+        <a (click)="back($event)" href="#">Zurück zur vorherigen Seite</a> |
+        <a routerLink="/">Zur Startseite</a>
     </p>
 
     <div


anonym-HPI:
Seems to work, mostly.
But using a path like: `/assets/about/non-existent-file` will result in an nginx 404.
This is not that important, as someone needs to enter such a path probably manually and for crawlers or other automatic requests it is still a 404.

One way is to just add the following:

`error_page 404 /404.html;`
As long as `/404.html` doesn't exist it will return `404`, but this makes it so that the themed 404 page is returned.
Could be also `/non-existent-page.html` or  so.
This could also be maybe be used, with documentation, so that people can create there custom 404 page by providing it at this location. But then we would have again the problem that the JS doesn't get a 404, but maybe people that create a custom 404 page, would already have license, imprint and privacy in place?

What I found is that it says "Go back to previous page" on the Imprint, Privacy and License page. Maybe it should be a link to the startpage?

anonym-HPI:
> > `error_page 404 /404.html;` As long as `/404.html` doesn't exist it will return `404`, but this makes it so that the themed 404 page is returned. Could be also `/non-existent-page.html` or so. This could also be maybe be used, with documentation, so that people can create there custom 404 page by providing it at this location. But then we would have again the problem that the JS doesn't get a 404, but maybe people that create a custom 404 page, would already have license, imprint and privacy in place?
> 
> As long as the status code is still 404, it should work no matter what HTML file will be delivered. As you've noticed yourself, nginx serves a default 404 page too and this does not affect the functionality of the imprint/privacy/license pages. But I'm nut sure whether this is worth the work. And having two kinds of 404 pages (the one in the frontend and the one for nginx) causes just some additional complexity.
> 

Without adding the `error_page 404 /non_existent_page.html;` we would have two different 404 pages.
With it (as long as the page doesn't exist) we would have only the Angular themed page, else we would have everywhere the Angular themed 404 page, but for files not existing in `/assets/about`, these would result in a 404 nginx page.
Probably nobody is getting to such a page accidently, so we could leave the nginx page, but we could also put the `error_page 404 /non_existent_page.html;` in the config to have everywhere the Angular themed error page.

> > What I found is that it says "Go back to previous page" on the Imprint, Privacy and License page. Maybe it should be a link to the startpage?
> 
> A link to the landing page might be better for situations in which someone navigated between the three about pages several times and then has to use the "back" link for the same number of times. On the other hand, there are also links to the about pages on the exercise pages. for a participant using this links (no matter whether by accident or not), a link to the landing page is not as useful as a "back" link.
> 
> I've built the navigation in a way that the links on the exercise page will open in a new tab so that the "back" link just closes the additional tab an no session information is lost (i.e., the participant do not has to be approved and assigned to a viewport again). In this scenario, the link might have no benefit over closing the tab in the browser's default way (clicking the "x" button of the tab), but these controls might be hidden if the exercise runs in full screen mode.
> 
> That's why I've built the link the way it is.

Geet it, but on phone or desktop, the link is not opened in a new tab (at least the links on pages itself or home page, which is fine, as they are not in an exercise).
What I would propose to have the title clickable (or if we had a logo, this one) or add a button "Going to Homepage" or so.

lukasrad02:
Here are two possible designs for the link to the landing page. Do you prefer one of them @anonym-HPI?

Screenshot 2023-02-09 at 15-34-21 DigitalFuesimManv

Screenshot 2023-02-09 at 15-35-22 DigitalFuesimManv

anonym-HPI:
> 

Would say the second one. Not 100% sure about the order of "back" and "home". In most browsers this is the chosen order and makes sense to have back on the left.

anonym-HPI:
`error_page 404` works in the way, that when nginx would serve its default `404` page, it instead would then server this custom page. But in our case as `/non_existent_page.html` is handled by the `location /` including the `try_files $uri $uri/ /index.html` it will try to find the file, as the file does not exist it will server the index.html, which will result in the Angular 404 page.
Same for other http status codes that are themed (like you mentioned - but if you want a themed 403 page, you would need to create one first, it seems, look for `error-404` in `frontend/src/app/pages/error-404`).
I thought maybe this page would land in the `dist` folder, so the way with `/non_existent_page.html` is not even needed, but seems to be, as `frontend/dist` is very empty, no `error-404` page in sight, probably in some `.js` file included.
Not sure how `error_page 403` would work, when trying to connect to a page that doesn't exist, if it would result in a `404`. 
Also for what do you think we would need `403` (forbidden)? No content on the server right now is containing any secrets that are authentication protected and should be not retrievable. Actually every file can be accessed without any authentication. Authentication is via the API and to the exercises, but these are in memory or in the database and not files.

docker/nginx/locations
```diff
@@ -3,6 +3,8 @@
 
     location /assets/about {
         # Requests to this location should not fall back to index.html
+        error_page 404 /non_existent_page.html;
+        error_page 403 /non_existent_page.html;
```
anonym-HPI, line 7:
```suggestion
        # Requesting existing files in this location should not fall back to index.html.
        # Files that do not exist should be handled by index.html
        error_page 404 /non_existent_page.html;
```

Right now I don't see a reason to add the `403`, proof me wrong.

anonym-HPI: I just found out, `error_page 404 /index.html;` also works, so would not need `non_existent_page.html`.

Also, found out when 403 seems to come, when accessing e.g. `/assets`.
I would propose to use:


```
error_page 403 /index.html;
error_page 404 /index.html;

```
Before all location blocks.

anonym-HPI: Yes, 

> When you're accessing `/assets/about/` without a file name specified, you'll receive a 403 since you're not allowed to see directory listings. At least this is the behavior I would expect and I get in my local docker setup (using the Dockerfile and docker-compose file from the repo).
> 
> However, I can't reproduce this with `fuesim-manv.de/assets/about/`, which shows me the Angular 404 page (although the HTTP status code is 403). Did you change some settings of nginx in deployment?

Yes, I changed it to this (removed uneccary commented out stuff):

```
    # Frontend
    root /usr/local/app/frontend/dist/digital-fuesim-manv;

     error_page 404 /index.html;
     error_page 403 /index.html;

    location /assets/about {
        # Requests to this location should not fall back to index.html
    }

    location / {
        try_files $uri $uri/ /index.html;
    }
...
```

lukasrad02: I like this approach since it it does not rely on a specific file (`non_existing_page.html`) to not exist. I've changed the config in https://github.com/hpi-sam/digital-fuesim-manv/pull/657/commits/3a51279e23ec41f501395dbcd277d1415c15b58a accordingly.

benn02 (approved):

anonym-HPI (approved):
Can be merged, but you have written about 403.  Right now when you reach a 403 site, it shows the 404 page. This is fine, but if you want a different one, I would suggest adding it in this PR.

lukasrad02:
> Can be merged, but you have written about 403. Right now when you reach a 403 site, it shows the 404 page. This is fine, but if you want a different one, I would suggest adding it in this PR.

It is on purpose, since the 403 from nginx should only occur if you're trying to access a directory without an `index.html` which is effectively a "Not found", since you're requesting a path which serves no content.

PR #664, opened 2023-02-08
Set up simulation framework

Nils1729:


shared/src/simulation/activities/unload-vehicle.ts
```diff
@@ -0,0 +1,39 @@
...
+import { uuid, UUID, uuidValidationOptions } from '../../utils';
+import { IsValue } from '../../utils/validators';
+import type {
+    SimulationActivity,
+    SimulationActivityState,
+} from './simulation-activity';
+
+// in ms (see state.currentTime)
+// TODO: make configurable
+const MAGIC_VALUE_UNLOAD = 5000;
```
lukasrad02, line 11: To make this value configurable, I would move it to the `BehaviorState`. Then the user could attach a behavior with the parameters of their choice to a simulated region and the behavior would supply its actions with the selected value.

lukasrad02:
I've noticed that you're separating behaviors and activities by placing them in two different folders. Wouldn't it be clearer if a behavior and its actions would be in the same file?

Or do you expect that we'll have many "generic" activities that would be spawned by multiple behaviors?

shared/src/simulation/activities/unload-vehicle.ts
```diff
@@ -0,0 +1,57 @@
...
+    {
+        activityState: UnloadVehicleActivityState,
+        tick(draftState, simulatedRegion, activityState) {
+            const vehicle = getElement(draftState, 'vehicle', activityState.vehicleId)
+            if (isNotInSimulatedRegion(vehicle)) {
+                // The vehicle has left the region for some reason. Cancel unloading.
+                removeActivity(simulatedRegion, activityState.id)
+            } else if (
+                draftState.currentTime >=
+                activityState.startTime + activityState.duration
```
lukasrad02, line 52: You're calling the parameter `duration` which is consistent to building the sum of `startTime` and `duration`. But when creating the action state in the behavior, you're already initializing `duration` with `arrivalTime + unloadDelay` which makes the `duration` become an `endTime`.

You should decide for one option.

shared/src/simulation/activities/index.ts
```diff
@@ -0,0 +1,19 @@
+import { unloadVehicleActivity } from './unload-vehicle';
+
+export const simulationActivities = {
+    unloadVehicleActivity,
+};
+
+export type ExerciseSimulationActivity = typeof unloadVehicleActivity;
```
lukasrad02, line 7: I guess this should be defined similar to the `ExerciseSimulationBehavior` (i.e. defining `ExerciseSimulationActivity` depending on `simulationActivities`).

shared/src/simulation/activities/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type ExerciseSimulationActivityState = InstanceType<
+    ExerciseSimulationActivity['activityState']
+>;
+
+// TODO: typing
+export const simulationActivityDirectory = Object.fromEntries(
```
lukasrad02, line 14: I would rather name this `[...]Dictionary` to be consistent to the `ExerciseActionTypeDictionary` of the actions and action reducers.

shared/src/simulation/activities/unload-vehicle.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export const unloadVehicleActivity: SimulationActivity<UnloadVehicleActivityState> =
+    {
+        activityState: UnloadVehicleActivityState,
+        tick(draftState, simulatedRegion, activityState) {
+            const vehicle = getElement(draftState, 'vehicle', activityState.vehicleId)
+            if (isNotInSimulatedRegion(vehicle)) {
+                // The vehicle has left the region for some reason. Cancel unloading.
+                removeActivity(simulatedRegion, activityState.id)
```
lukasrad02, line 49: The vehicle could also have been deleted for some reason which would lead to an error thrown from `getElement`.

In general, we should decide how to handle errors in activities and behaviors. I would propose to throw errors and catch them in the function invoking all the activities/behaviors. Then we could also do some logging which would make it easier to debug the simulation system.

shared/src/simulation/events/vehicle-arrived.ts
```diff
@@ -0,0 +1,33 @@
...
+    readonly type = 'vehicleArrivedEvent';
+
+    @IsUUID(4, uuidValidationOptions)
+    public readonly id: UUID = uuid();
+
+    @IsUUID(4, uuidValidationOptions)
+    readonly vehicleId!: UUID;
+
+    @IsUUID(4, uuidValidationOptions)
+    readonly SimulatedRegionId!: UUID;
```
lukasrad02, line 18: Use `camelCase` for consistency.

shared/src/simulation/utils/simulated-region.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export function removeActivity(
+    simulatedRegion: Mutable<SimulatedRegion>,
+    activityId: UUID
+) {
+    // TODO: Maybe add a proper teardown function to SimulatedActivity
+    delete simulatedRegion.activities[activityId];
+}
+
+export function unloadVehicle(
```
lukasrad02, line 90: I think that this should rather be in a `vehicle.ts` utils file.

shared/src/simulation/activities/simulation-activity.ts
```diff
@@ -0,0 +1,23 @@
...
+        draftState: Mutable<ExerciseState>,
+        simulatedRegion: Mutable<SimulatedRegion>,
+        activityState: Mutable<S>,
+        tickInterval: number
+    ) => void;
+    readonly terminate?: (
+        draftState: Mutable<ExerciseState>,
+        simulatedRegion: Mutable<SimulatedRegion>,
+        activityId: UUID
+    ) => void;
```
lukasrad02, line 22: I would rename this to `onTerminate` (or similar) to state that this is just an event handler and you have to use `terminateActivity()` (from the utils) if you want to terminate an activity.

shared/src/simulation/behaviors/unload-arrived-vehicles.ts
```diff
@@ -0,0 +1,53 @@
...
+    static readonly create = getCreate(this);
+}
+
+export const unloadArrivingVehiclesBehavior: SimulationBehavior<UnloadArrivedVehiclesBehaviorState> =
+    {
+        behaviorState: UnloadArrivedVehiclesBehaviorState,
+        handleEvent(draftState, simulatedRegion, behaviorState, event) {
+            if (
+                event.type === 'vehicleArrivedEvent' &&
+                event.simulatedRegionId === simulatedRegion.id
```
lukasrad02, line 41: I think this check (and therefore the `simulatedRegionId` property on the event at all) is not necessary since `handleEvents` will only be called for events related to the simulated region passed as argument.

Nils1729:
To the reviewer: I suggest leaving the debug elements in the popup for now as they will be required not only for review and demonstration purposes but also for debugging new features built onto this PR.

shared/src/state-migrations/20-add-simulation-properties.ts
```diff
@@ -0,0 +1,28 @@
+import { seededRandomState } from '../simulation';
+import type { UUID } from '../utils';
+import type { Migration } from './migration-functions';
+
+export const addSimulationProperties20: Migration = {
+    actions: null,
```
lukasrad02, line 6: You have to migrate the `[SimulatedRegion] Add simulated region` action too, since the state you're migrating may be the initial state with all actions re-applied on import.

shared/src/simulation/activities/unload-vehicle.ts
```diff
@@ -0,0 +1,76 @@
...
+    constructor(
+        vehicleId: UUID,
+        startTime: number,
+        duration: number,
+        id?: UUID
+    ) {
+        this.vehicleId = vehicleId;
+        this.startTime = startTime;
+        this.duration = duration;
+        if (id) this.id = id;
```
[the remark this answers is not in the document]
Nils1729: Currently not but for consistency with all other objects in the state I made it optional.

Nils1729:
> > To the reviewer: I suggest leaving the debug elements in the popup for now as they will be required not only for review and demonstration purposes but also for debugging new features built onto this PR.
> 
> As this is more or less the last PR of the current milestone (we still have to implement automatic treatments based on this PR but then we're done) and we'll merge the `dev` branch into `main` after the milestone, when do you plan to remove/change this? Since I think we should not have debugging stuff on the `main` branch.

Good point. I'll revert these changes after the review. Still we could take advantage of a similar debugging interface in the future for dev deployments.

shared/src/models/simulated-region.ts
```diff
@@ -47,6 +53,20 @@ export class SimulatedRegion {
         this.name = name;
     }
 
+    @Type(...simulationEventTypeOptions)
+    @ValidateNested()
+    public readonly inEvents: readonly ExerciseSimulationEvent[] = [];
```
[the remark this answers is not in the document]
Nils1729: Yes, the idea was that events/request could be send from regions. An alternative approach would be to have some sort of target but that would need to be globally unique.

shared/src/simulation/utils/randomness.ts
```diff
@@ -0,0 +1,93 @@
...
+    @IsLiteralUnion({ sfc32: true })
+    readonly algo = 'sfc32';
+
+    @ValidateIf((o) => o.algo === 'sfc32')
+    @IsArray()
+    @ArrayMinSize(4)
+    @ArrayMaxSize(4)
+    @IsInt({ each: true })
+    @Min(0, { each: true })
+    @Max(4294967295 /* UINT32_MAX */, { each: true })
```
lukasrad02, line 33: There can be cases in which elements of the sfc state are bigger than this value which causes subsequent validations (e.g. when ex- and importing an exercise) to fail.

shared/src/simulation/activities/unload-vehicle.ts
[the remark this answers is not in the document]
lukasrad02: In my point of view, `id` is just some property like `startTime` or `duration` that has to be set to create a valid instance. Therefore, I don't think its inconsistent to enforce this parameter to be set and enforcing it can prevent us from hard to find bugs with inconsistent ids.

shared/src/simulation/utils/randomness.ts
```diff
@@ -1,93 +1,67 @@
...
 }
 
+// Returns 0 to UINT32_MAX
 export function nextInt(
     draftState: Mutable<ExerciseState>,
     upperBound: number
 ): number {
-    return Math.trunc(advance(draftState) % upperBound);
+    const state = advance(draftState)
+        .slice(4)
```
lukasrad02, line 48: I think you want to have 4 Elements from the array here, but you're retrieving all but the first 4 elements here.

lukasrad02 (approved):
Looks good to me!

Keep in mind to remove the debugging elements from the UI

PR #665, opened 2023-02-09
Document meaning of position and size of rects

lukasrad02:
In #663 we've fixed a bug related to the position and size of viewports and simulated regions. To prevent similar bugs to occur again, this PR should improve the documentation of these values.

### DIFF
--- a/shared/src/models/simulated-region.ts
+++ b/shared/src/models/simulated-region.ts
@@ -22,7 +22,8 @@ export class SimulatedRegion {
     public readonly type = 'simulatedRegion';
 
     /**
-     * top-left position
+     * Position of the viewport.
+     * For positive {@link size}s, this refers to the upper left corner of the simulated region, but during resizing, corners may be swapped.
      *
      * @deprecated Do not access directly, use helper methods from models/utils/position/position-helpers(-mutable) instead.
      */
@@ -30,6 +31,11 @@ export class SimulatedRegion {
     @IsPosition()
     public readonly position: Position;
 
+    /**
+     * Size of the viewport.
+     * The width of a viewport grows to the left, the height grows downwards.
+     * Negative sizes, referring to the opposite direction, are possible.
+     */
     @ValidateNested()
     @Type(() => Size)
     public readonly size: Size;
--- a/shared/src/models/viewport.ts
+++ b/shared/src/models/viewport.ts
@@ -21,7 +21,8 @@ export class Viewport {
     public readonly type = 'viewport';
 
     /**
-     * top-left position
+     * Position of the viewport.
+     * For positive {@link size}s, this refers to the upper left corner of the viewport, but during resizing, corners may be swapped.
      *
      * @deprecated Do not access directly, use helper methods from models/utils/position/position-helpers(-mutable) instead.
      */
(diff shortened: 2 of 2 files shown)


shared/src/models/simulated-region.ts
```diff
@@ -22,14 +22,20 @@ export class SimulatedRegion {
     public readonly type = 'simulatedRegion';
 
     /**
-     * top-left position
+     * Position of the viewport.
```
ClFeSc, line 25:
```suggestion
     * Position of the simulated region.
```

shared/src/models/simulated-region.ts
```diff
@@ -22,14 +22,20 @@ export class SimulatedRegion {
...
+     * For positive {@link size}s, this refers to the upper left corner of the simulated region, but during resizing, corners may be swapped.
```
ClFeSc, line 26:
```suggestion
     * For positive {@link size}s, this refers to the upper left corner of the simulated region, but after resizing, corners may be swapped.
```

To make clear that this does not only happens *while* resizing (so in the act of it) but also persists thereafter.

shared/src/models/simulated-region.ts
```diff
@@ -22,14 +22,20 @@ export class SimulatedRegion {
...
      *
      * @deprecated Do not access directly, use helper methods from models/utils/position/position-helpers(-mutable) instead.
      */
     @ValidateNested()
     @IsPosition()
     public readonly position: Position;
 
+    /**
+     * Size of the viewport.
```
ClFeSc, line 35:
```suggestion
     * Size of the simulated region.
```

shared/src/models/simulated-region.ts
```diff
@@ -22,14 +22,20 @@ export class SimulatedRegion {
...
+     * The width of a viewport grows to the left, the height grows downwards.
```
ClFeSc, line 36:
```suggestion
     * The width of a simulated region grows to the left, the height grows downwards.
```

shared/src/models/viewport.ts
```diff
@@ -21,14 +21,20 @@ export class Viewport {
     public readonly type = 'viewport';
 
     /**
-     * top-left position
+     * Position of the viewport.
+     * For positive {@link size}s, this refers to the upper left corner of the viewport, but during resizing, corners may be swapped.
```
ClFeSc, line 25:
```suggestion
     * For positive {@link size}s, this refers to the upper left corner of the viewport, but after resizing, corners may be swapped.
```

Same argument as with the simulated regions.

shared/src/models/simulated-region.ts
Dassderdie, line 36:
```suggestion
     * The width of a viewport grows to the right, the height grows downwards.
```

Or am I wrong again?

ClFeSc: No, should be correct.

Dassderdie:
There are still multiple occurrences of `topLeftCoordinate` missing in this PR.

I also believe the `ResizeRectangleInteraction` is currently only working by chance.
It seems to rely on `PolygonGeometryHelper.getElementCoordinates` putting the `(x, y)`-coordinate first.

To fix this, I propose to change the position of a `Viewport` and `SimulatedRegion` to be the center of the rectangle instead and enforce the Size to be positive. This should also reduce confusion in the future.

I would offer to do this migration sometime in the next week.

lukasrad02:
> I would offer to do this migration sometime in the next week.

Sounds good, thank you!

So we can close this PR as it will be obsolete after your changes anyway.

Issue #660, opened 2023-02-08
Enforce connection of vehicle to material an personnel with regard to simulated regions


Nils1729:
When vehicles are being moved to simulated regions, the associated personnel and materials are currently left behind. Materials and Personnel should be moved to the simulated region too. This preferably only affects those elements that are not in the vehicle as the others should stay in the vehicle which itself is moved to the simulated region.

christianzoellner:
Alternatively, you could reject moves of emtpy vehicles or out-of-vehicel-personnel to simulated regions. Might be less complicated.

PR #669, opened 2023-02-14
Improve vehicle load and unload behavior

lukasrad02:
This PR adds the feature requested in #660 that personnel will stay near their vehicle, meaning that when moving it into a simulated region, the personnel will be added to this region, too.

We've decided that the personnel should be on the vehicle rather than being added to the simulated region separately. To avoid personnel being removed from the map suddenly, the action reducer refuses to add vehicles to a simulated region if they're not fully loaded. In this case, a `ReducerError` is thrown, resulting in a toast shown to the user (which was fixed in #676).

For consistency with the previous behavior, adding a patient to a vehicle still automatically adds the personnel and material to the vehicle, too.

To prevent making it harder to add a vehicle to a simulated region than it was before, a load vehicle button has been added to the vehicle, that makes all personnel enter the vehicle automatically. This can be used to ensure the vehicle can be added to the region.

This RP introduces a breaking change due to fixing the todo: *Block vehicles whose material and personnel are unloaded* in the `transportPatientToHospital` action reducer. Importing an exercise in which a vehicle was transferred to a hospital without it's personnel and material on board will fail. However, this case is not likely to occur since adding a patient to a vehicle forcefully adds the personnel and material, too. The only case in which the error can occur is if the patient was added while a personnel of the vehicle was in transfer and therefore could not be added to the vehicle.

### DIFF
--- a/shared/src/store/action-reducers/utils/completely-load-vehicle.ts
+++ b/shared/src/store/action-reducers/utils/completely-load-vehicle.ts
@@ -1,0 +1,66 @@
+import type { Vehicle } from '../../../models';
+import {
+    isInVehicle,
+    isNotInTransfer,
+    VehiclePosition,
+} from '../../../models/utils';
+import { changePosition } from '../../../models/utils/position/position-helpers-mutable';
+import type { ExerciseState } from '../../../state';
+import type { Mutable } from '../../../utils';
+import { getElement } from './get-element';
+
+/**
+ * Checks whether all materials and personnel belonging to a vehicle are in it.
+ *
+ * @param draftState The state to operate in
+ * @param vehicle The vehicle to check
+ * @returns `true` if everything is in the vehicle, `false` otherwise
+ */
+export function isCompletelyLoaded(
+    draftState: Mutable<ExerciseState>,
+    vehicle: Mutable<Vehicle>
+) {
+    const materialsLoaded = Object.keys(vehicle.materialIds).map((materialId) =>
+        isInVehicle(getElement(draftState, 'material', materialId))
+    );
+
+    const personnelLoaded = Object.keys(vehicle.personnelIds).map(
+        (personnelId) =>
+            isInVehicle(getElement(draftState, 'personnel', personnelId))
+    );
+
+    return (
+        materialsLoaded.every((loaded) => loaded) &&
+        personnelLoaded.every((loaded) => loaded)
+    );
+}
+
+/**
+ * Loads all materials and personnel belonging to a vehicle into it.
+ * Personnel that is currently in transfer won't be loaded.
+ *
+ * @param draftState The state to operate in
+ * @param vehicle The vehicle to load
+ */
+export function completelyLoadVehicle(
+    draftState: Mutable<ExerciseState>,
+    vehicle: Mutable<Vehicle>
+) {
+    const vehiclePosition = VehiclePosition.create(vehicle.id);
+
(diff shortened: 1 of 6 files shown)


shared/src/store/action-reducers/utils/completely-load-vehicle.ts
```diff
@@ -0,0 +1,66 @@
...
+    const personnelLoaded = Object.keys(vehicle.personnelIds).map(
+        (personnelId) =>
+            isInVehicle(getElement(draftState, 'personnel', personnelId))
+    );
+
+    return (
+        materialsLoaded.every((loaded) => loaded) &&
+        personnelLoaded.every((loaded) => loaded)
+    );
+}
```
benn02, line 36: This assumes that if a thing is in A vehicle, the vehicle is loaded. This is currently correct, since material and personnel are strictly bound to their vehicle. But I think the usage of Vehicle.isInVehicle would be cleaner here since it additionally checks if the things are inside the correct vehicle

shared/src/store/action-reducers/utils/completely-load-vehicle.ts
```diff
@@ -0,0 +1,66 @@
...
+) {
+    const vehiclePosition = VehiclePosition.create(vehicle.id);
+
+    Object.keys(vehicle.materialIds).forEach((materialId) => {
+        changePosition(
+            getElement(draftState, 'material', materialId),
+            vehiclePosition,
+            draftState
+        );
+    });
```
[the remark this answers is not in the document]
benn02: I also don't really like the error message "Fehler beim Senden der Aktion". I feel like something more akkurate would be "Diese Aktion ist nicht möglich". The former implies a technical error while the error is on the user side and the software is working fine, what the letter makes more obvious. I also think we should always keep reducer errors that are of technical concern and expected errors on unwanted user behaviour apart with different messages and errors.

benn02:
I also don't really like the error message "Fehler beim Senden der Aktion". I feel like something more akkurate would be "Diese Aktion ist nicht möglich". The former implies a technical error while the error is on the user side and the software is working fine, what the letter makes more obvious. I also think we should always keep reducer errors that are of technical concern and expected errors on unwanted user behaviour apart with different messages and errors.

benn02:
> I also don't really like the error message "Fehler beim Senden der Aktion". I feel like something more akkurate would be "Diese Aktion ist nicht möglich". The former implies a technical error while the error is on the user side and the software is working fine, what the letter makes more obvious. I also think we should always keep reducer errors that are of technical concern and expected errors on unwanted user behaviour apart with different messages and errors.

I will open a new issue for this

shared/src/store/action-reducers/utils/completely-load-vehicle.ts
lukasrad02: Arguing this way, we should mark `isInVehicle` as deprecated, since it should never be used for any purpose than as perquisite for `currentVehicleIdOf` (as `Vehicle.isInVehicle`) does.

Changes the code in a way that personnel may enter multiple or any vehicle(-s) would lead to a big bunch of changes anyway while sticking to the generic `isInVehicle` here would at least keep the changes introduced by this PR small.

lukasrad02: Okay, you've got a point.

I've also changed the occurrences in `vehicle-popup.component.ts` to be consistent across this PR.

benn02 (approved):

PR #671, opened 2023-02-15
Dragging onto the Map now triggers Drop Events

benn02:
The Drag Service now executes the Side Effects of the Drop. To make this possible, a refactoring of the onDrop Functions was necessary.

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core';
 import { Store } from '@ngrx/store';
 import type {
+    Element as Elem,
     ImageProperties,
     MapImageTemplate,
     PatientCategory,
@@ -16,7 +17,11 @@ import {
     SimulatedRegion,
     MapPosition,
 } from 'digital-fuesim-manv-shared';
+import type { Feature } from 'ol';
+import type VectorLayer from 'ol/layer/Vector';
 import type OlMap from 'ol/Map';
+import type { Pixel } from 'ol/pixel';
+import type VectorSource from 'ol/source/Vector';
 import { ExerciseService } from 'src/app/core/exercise.service';
 import type { AppState } from 'src/app/state/app.state';
 import {
@@ -24,6 +29,7 @@ import {
     selectPersonnelTemplates,
 } from 'src/app/state/application/selectors/exercise.selectors';
 import { selectStateSnapshot } from 'src/app/state/get-state-snapshot';
+import type { FeatureManager } from '../exercise-map/utility/feature-manager';
 
 @Injectable({
     providedIn: 'root',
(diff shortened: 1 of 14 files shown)


frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
```diff
@@ -179,83 +209,128 @@ export class DragElementService {
...
+                return;
+            }
+            // We stop propagating the event as soon as the onFeatureDropped function returns true
+            return this.layerFeatureManagerDictionary
+                .get(layer as VectorLayer<VectorSource>)!
+                .onFeatureDrop(
+                    createdElement as Elem,
+                    droppedOnFeature as Feature
+                );
+        });
```
lukasrad02, line 331: Maybe you can move this code to a function out of this class to share it with `OlMapInteractionsManager.handleTranslateEnd`

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/catering-lines-feature-manager.ts
```diff
@@ -12,6 +12,7 @@ import type { CateringLine } from 'src/app/shared/types/catering-line';
 import type { AppState } from 'src/app/state/app.state';
 import { selectVisibleCateringLines } from 'src/app/state/application/selectors/shared.selectors';
 import type OlMap from 'ol/Map';
+import type { Element as Elem } from 'digital-fuesim-manv-shared';
```
lukasrad02, line 15: In my opinion, it would be okay to shadow the global `Element` with this import, since we're not interacting with the DOM from the feature managers and not obvious what `Elem` should be when you see it in the code.

If you change it, please do so in the other feature managers, too.

frontend/src/app/pages/exercises/exercise/shared/core/drag-element.service.ts
benn02: I thought about that, but while they are similar in Structure, they are different in break conditions and the Way the onFeatureDrop Function is called, so I suggest leaving it as is.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/catering-lines-feature-manager.ts
benn02: This could lead to Problems in the Feature Managers since they know the Type element as the specific Element that they manage

lukasrad02: I would suggest renaming the type parameter `Element` to something like `ManagedElement` to differentiate between types and type parameters.

benn02: I renamed both

lukasrad02 (approved):
Looks good to me

PR #674, opened 2023-02-16
Use @noble/hashes instead of hash.js

Nils1729:


benn02:
Did you execute npm run build because if i run it it addes noble to the package lock of benchmark. I think that should be committed.

Nils1729:
> Did you execute npm run build because if i run it it addes noble to the package lock of benchmark. I think that should be committed.

I did run `npm run install:all`. I just found an error in that script so dependencies for `benchmark` were not installed. Fixed it.

benn02 (approved):

PR #626, opened 2023-01-26
Allow elements to be added to simulated regions

Nils1729:
This is a sketch as we will merge this functionality after #608, #625 and #616.

PR #675, opened 2023-02-16
Refactor Viewport and SimulatedRegion position to be the center instead of a corner

### LINKED ISSUE #666: Refactor Viewport and SimulatedRegion position to be the center instead of a corner
See #665 for the previous discussions.

Dassderdie:
* Refactor Viewport and SimulatedRegion position to be the center instead of a corner
* Simplified `GeometryHelper` API by removing `Positions<>`
* The `ResizeRectangleInteraction` now has a minimum size and prevents flipping of the rectangle


frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/simulated-region-feature-manager.ts
```diff
@@ -142,19 +142,16 @@ export class SimulatedRegionFeatureManager
...
             this.exerciseService.proposeAction(
                 {
                     type: '[SimulatedRegion] Add Element',
                     simulatedRegionId: droppedOnSimulatedRegion.id,
-                    elementToBeAddedType: droppedElement.type as
-                        | 'material'
-                        | 'patient'
-                        | 'personnel'
-                        | 'vehicle',
+                    elementToBeAddedType: droppedElement.type,
```
Dassderdie, line 154: [I already brought this up previously.](https://github.com/hpi-sam/digital-fuesim-manv/pull/626/files#r1093237887)

If someone would have mistyped one of the keys in the array that was previously present, you wouldn't have noticed this. Therefore I changed it back to the old condition.
The type checks now work correctly as `droppedElement` is typed since your refactoring.

benn02 (commented):
LGTM but it still has unresolved conflicts, please resolve them for approval

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/resize-rectangle-interaction.ts
```diff
@@ -66,47 +75,54 @@ export class ResizeRectangleInteraction extends PointerInteraction {
...
         return true;
     }
 
+    private calculateNewScale(
+        draggedCorner: number,
+        originCorner: number,
+        mouseCoordinate: number
+    ) {
+        const oldXLength = draggedCorner - originCorner;
+        const newXLength = mouseCoordinate - originCorner;
```
Greenscreen23, line 107: Please remove the X from those variable names, as it is also used for X coordinates.

shared/src/state-migrations/21-refactor-rectangular-element-positions-to-center.ts
```diff
@@ -0,0 +1,67 @@
...
+    },
+    state: (state: any) => {
+        for (const viewport of Object.values(state.viewports)) {
+            migrateRectangularElement(viewport);
+        }
+        for (const simulatedRegion of Object.values(state.simulatedRegions)) {
+            migrateRectangularElement(simulatedRegion);
+        }
+    },
+};
```
Greenscreen23, line 48: What about simulated regions?

Greenscreen23 (changes requested):

lukasrad02 (commented):
Besides the aspects mentioned by Benildur and Lukas, I'm fine with this.

shared/src/state-migrations/25-refactor-rectangular-element-positions-to-center.ts
```diff
@@ -0,0 +1,95 @@
...
+ * Migrates the position of a rectangular element to the center of the element.
+ *
+ * This migration is not accurate for previously "flipped" viewports (height and/or width was negative).
+ * To catch this case, access to the size of the viewport before this migration would be required.
+ *
+ * @param migratedSize the size of the element after it had been migrated to the center
+ */
+function migratePositionWithMigratedSize(coordinates: any, migratedSize: any) {
+    migratePosition(coordinates, migratedSize);
+}
```
Dassderdie, line 90: The `move`-actions only include the `targetPosition`. However, the "real" position of the element was defined by its position and size (because of negative height/width). This information gets lost during this migration. Therefore, the current migration system doesn't allow for an accurate migration in these cases.

I believe this to be acceptable, as this is an edge case that would only result in some regions in imported exercise histories being at slightly different positions.

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/resize-rectangle-interaction.ts
Dassderdie: Fixed

shared/src/state-migrations/21-refactor-rectangular-element-positions-to-center.ts
```diff
@@ -0,0 +1,67 @@
...
+                    }
+                    break;
+                }
+                case '[Viewport] Resize viewport': {
+                    const oldViewportSize = cloneDeep(action.newSize);
+                    oldViewportSizes[action.viewportId] = oldViewportSize;
+                    migratePosition(action.targetPosition, oldViewportSize);
+                    migrateSize(action.newSize);
+                    break;
+                }
```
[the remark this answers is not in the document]
Dassderdie: Added them.

shared/src/state-migrations/25-refactor-rectangular-element-positions-to-center.ts
lukasrad02: Depending on the position and size, the location can differ a lot, even with values that are realistic to occur in an exercise. Since resizing of viewports and simulated regions is an operation of exercise creation, this is a breaking change according to our definition (see <https://github.com/hpi-sam/digital-fuesim-manv#versions>).

We though about a solution for this problem in our team. Our idea was to keep a log which "optical" corner of the rectangle the current location refers to. This log could be updated from the `newSize` of each resize action before migrating it (so one can deduct the corner from the sign of the width and height). With the information about the corner, one could restore the exact desired position and extent on the map after a move action.

What do you think about this?

Dassderdie: > Depending on the position and size, the location can differ a lot, even with values that are realistic to occur in an exercise. Since resizing of viewports and simulated regions is an operation of exercise creation, this is a breaking change according to our definition (see <https://github.com/hpi-sam/digital-fuesim-manv#versions>).

This should only be a breaking change for exercises with history. With my current implementation, states should be losslessly converted to the new definition. Only certain Move-Actions in the history should pose problems.

> We though about a solution for this problem in our team. Our idea was to keep a log which "optical" corner of the rectangle the current location refers to. This log could be updated from the `newSize` of each resize action before migrating it (so one can deduct the corner from the sign of the width and height). With the information about the corner, one could restore the exact desired position and extent on the map after a move action.
> 
> What do you think about this?

If I understand it correctly, this would be similar to my previous approach (look at the migration in this PR from a few weeks ago), in which I kept a log of all old viewport sizes. (Though, I believe I forgot to populate the log with the previous sizes of viewports which didn't have a resize or create action.) To still do this, one would have to adjust the migration system again, to realise this log.

Alternatively one could create two migrations for this change and use the first migration to write this log in the intermediate exerciseState and perform the lossless migration. The second one would only remove this log from the state again.

lukasrad02: > > Depending on the position and size, the location can differ a lot, even with values that are realistic to occur in an exercise. Since resizing of viewports and simulated regions is an operation of exercise creation, this is a breaking change according to our definition (see https://github.com/hpi-sam/digital-fuesim-manv#versions).
> 
> This should only be a breaking change for exercises with history. With my current implementation, states should be losslessly converted to the new definition. Only certain Move-Actions in the history should pose problems.

At the moment, our definition only differentiates between prepared and played exercises. But prepared exercises can also be exported with history which I think is a legitimate choice.

> > We though about a solution for this problem in our team. Our idea was to keep a log which "optical" corner of the rectangle the current location refers to. This log could be updated from the `newSize` of each resize action before migrating it (so one can deduct the corner from the sign of the width and height). With the information about the corner, one could restore the exact desired position and extent on the map after a move action.
> > What do you think about this?
> 
> If I understand it correctly, this would be similar to my previous approach (look at the migration in this PR from a few weeks ago), in which I kept a log of all old viewport sizes. [...]

Yes, this is what we though of

> To still do this, one would have to adjust the migration system again, to realise this log.
> 
> Alternatively one could create two migrations for this change and use the first migration to write this log in the intermediate exerciseState and perform the lossless migration. The second one would only remove this log from the state again.

Probably not the cleanest solution, but the log could also be saved in a local variable in your typescript file, requiring no changes to the migration system

Dassderdie: > At the moment, our definition only differentiates between prepared and played exercises. But prepared exercises can also be exported with history, which I think is a legitimate choice.

What is the purpose of not yet played exports with a history? Do you want to keep the option to undo and redo changes in the prepared exercise?

> Probably not the cleanest solution, but the log could also be saved in a local variable in your typescript file, requiring no changes to the migration system

This would introduce strange side effects to this particular migration function.
You would have to manually empty the log if the `exerciseId` of the `intermediaryState` changes. In addition, you would have to rely on the call of the state-migration function to be directly followed by all the action migration calls of the same exercise (first migrating the initial states of all exercises, and then all their actions would result in incorrect migrations). Otherwise, you would have to save all logs for all exercises, which would be a memory leak.
I would rather opt for my approach as this state introduction could set a bad precedent for future programmers.

lukasrad02: > What is the purpose of not yet played exports with a history? Do you want to keep the option to undo and redo changes in the prepared exercise?

Yes, this is what I've though of. Trainers might have complex scenarios and then just omit the last couple of actions to create an easier version of them. (Although I think there is no option to do this from the user interface yet).

Another reason for me is that I would expect some people to be in doubt of the meaning of the export options "Kompletter Verlauf" (full history) and "Aktueller Zustand" (current state) and export their prepared exercises as full history just to be safe.

> I would rather opt for my approach as this state introduction could set a bad precedent for future programmers.

Are you talking about the two migrations or about the changes to the migration system?

lukasrad02: > The two migrations.

Okay, sounds good.

> A new migration system would need more discussion and would probably also only make sense after we have more experience with the needs of migrations. The last refactoring removed some capabilities (adding actions, combining actions, reordering actions, in general, more context-aware operations on actions). Though I believe all except adding actions can be worked around via two consecutive migrations.

@Nils1729 might be the right person if you'd like to go into the discussion on the migration system (not for this PR but in general)

PR #677, opened 2023-02-18
Refactor to use StoreService instead of Store

Dassderdie:
* There is now a `StoreService` that wraps the ngrx-store
* This improves the readability and especially discoverability of synchronous access to the store (`storeService.select(selector)` instead of `getStateSnapshot(selector, store)`)
* It is now possible to add comments to the store functions (like that we usually don't dispatch actions but propose them via the `exerciseService`)
* Previously, one always had to import `Store<AppState>` to get the typings - now `StoreService` is enough
* Makes it easier to replace ngrx should the need arise as we only use a very small feature set of it

Nearly all of the changes are just straightforward renaming.


benn02 (commented):
LGTM, but this still has conflicts to be resolved. Please resolve them for approval.

frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/tabs/behavior-tab/behaviors/assign-leader/simulated-region-overview-behavior-assign-leader.component.ts
```diff
@@ -23,13 +22,13 @@ export class SimulatedRegionOverviewBehaviorAssignLeaderComponent
...
-        if (this.assignLeaderBehaviorState.leaderId) {
-            this.currentLeader = this.store.select(
-                createSelectPersonnel(this.assignLeaderBehaviorState.leaderId)
-            );
-        }
+        this.currentLeader = this.assignLeaderBehaviorState.leaderId
+            ? this.storeService.select$(
+                  createSelectPersonnel(this.assignLeaderBehaviorState.leaderId)
+              )
+            : undefined;
```
Dassderdie, line 32: I believe I fixed a potential bug here:
If the `leaderId` would have changed back from an id to `undefined`, the component would still have displayed the previous leader.

PR #680, opened 2023-02-22
Check on Translate End if the Popup needs to close

benn02:
There is now an Array with UUIDs that indicate that if one of the elements corresponding to one of the UUIDs move, the pop-up shall be closed.

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/popup-manager.ts
+++ b/frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/popup-manager.ts
@@ -146,7 +146,7 @@ export class PopupManager {
 }
 
 /**
- * {@link closingUUIDs} is an array containing the UUIDs of elements that when clicked shall close the pop up
+ * {@link closingUUIDs} is an array containing the UUIDs of elements that when clicked shall close the pop-up
  */
 export interface OpenPopupOptions<
     Component extends PopupComponent,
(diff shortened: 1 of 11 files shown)


frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/popup-manager.ts
```diff
@@ -32,6 +33,15 @@ export class PopupManager {
...
     private popupsEnabled = true;
 
+    private _currentClosingIds: UUID[] = [];
+
+    public get currentClosingIds(): UUID[] {
+        return this._currentClosingIds;
+    }
+    public set currentClosingIds(value: UUID[]) {
+        this._currentClosingIds = value;
+    }
```
lukasrad02, line 43: I think it would be enough to have a public getter that returns `currentlyOpenPopupOptions.closingUUIDs`. This would also prevent storing this information in two places, which could lead to inconsistent data otherwise.

benn02: Done

lukasrad02: In your current solution, you don't need the `?` in `return this.currentlyOpenPopupOptions?.closingUUIDs;`, since you're checking for `undefined` in the if statement above.

If you prefer a shorter solution, you could use `return this.currentlyOpenPopupOptions?.closingUUIDs ?? [];` and omit the if statement.

lukasrad02 (approved):

PR #684, opened 2023-02-23
Fix the wrong message that occures when dragging a vehicle to a simulated region

benn02:
For more details, see the issue.
This just checks whether the vehicle was not there a tick ago before displaying a message. So addition/deletion in the same tick is not possible.


frontend/src/app/core/exercise.service.ts
```diff
@@ -243,10 +251,15 @@ export class ExerciseService {
                         !client.isInWaitingRoom
                 ),
                 switchMap((client) => this.store.select(selectVisibleVehicles)),
+                withLatestFrom(this.store.select(selectCurrentTime)),
                 pairwise(),
                 takeUntil(this.stopNotifications$)
             )
-            .subscribe(([oldVehicles, newVehicles]) => {
+            .subscribe(([[oldVehicles, oldTime], [newVehicles, newTime]]) => {
```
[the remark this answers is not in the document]
benn02: I was hesitant to use 0 here because in my mind debouncing by 0ms means just not debouncing (even if it's interpreted different and working fine), so I used a timer for 1ms to debounce since readding a vehicle multiple times in a span of 1ms should also not trigger a message.

frontend/src/app/core/exercise.service.ts
```diff
@@ -251,15 +251,12 @@ export class ExerciseService {
                         !client.isInWaitingRoom
                 ),
                 switchMap((client) => this.store.select(selectVisibleVehicles)),
-                withLatestFrom(this.store.select(selectCurrentTime)),
+                // Do not trigger the message if the vehicle was removed and added again in the same millisecond
+                debounce(() => timer(1)),
```
Dassderdie, line 255:
```suggestion
                debounceTime(0),
```

Or `1` if it didn't work with `0`. Though, I'm currently missing why it shouldn't :/

frontend/src/app/core/exercise.service.ts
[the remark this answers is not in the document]
Dassderdie: I would put it like this: 
The `0` confuses an inexperienced JS-developer, while the `1` confuses an experienced JS-developer.

If there is a `0`, I know it is just skipping multiple synchronous emits. If there is a `1`, I think this is about latency and could potentially break on different systems, etc.

frontend/src/app/core/exercise.service.ts
benn02: It works that way, but I don't find it obvious that one output per 0ms (which is the meaning of debounceTime(0)), does discard any events. So I rather use 1 here because that makes it obvious that there is a real time span during which outputs are discarded.

Dassderdie: I would put it like this:
The 0 confuses an inexperienced JS-developer, while the 1 confuses an experienced JS-developer.

If there is a 0, I know it is just skipping multiple synchronous emits. If there is a 1, I think this is about latency and could potentially break on different systems, etc.

benn02: Also, I realized there is one more Problem with the notification coming up when the viewport one is restricted to Changed, which also should not happen

benn02: > I would put it like this: The 0 confuses an inexperienced JS-developer, while the 1 confuses an experienced JS-developer.
> 
> If there is a 0, I know it is just skipping multiple synchronous emits. If there is a 1, I think this is about latency and could potentially break on different systems, etc.

I just found the word timeout in the documentation confusing since it is more of a time window that closes after that time so in the time window of 0 can be multiple elements but if I wait 0 seconds for more events it sounds like there is a maximum of one event, but I think 0 is the optimal solution after reading the documentation a second time

Dassderdie: I believe the naming is derived from the [`setTimeout()` function](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout) that's part of the browser- and nodejs-API.

benn02: I think I now fixed both things the movement in a simulated region issue with debounceTime and the switch viewport issue with moving pairwise inside switch map so only pairs from the same viewport are built

Dassderdie: I believe `pairwise` should come before `debounce `.
This way you should also be able to move them back to the outer pipe?

Dassderdie (approved):
LGTM - I haven't tested it though.

I will probably have a shot at making these debounceTime(0) unnecessary by tweaking the OptimisticUpdateHandler and ngrx-store a bit...

Nils1729 (approved):

PR #691, opened 2023-03-01
Clean up pipeline

Nils1729:
* Extracts caching of `node_module`s and `shared/dist` into own action
* Extracts updating the version into own action
* Removes unnecessary lfs checkouts
* Improve some string formatting in `yml` files


.github/actions/cache-node-modules/action.yml
```diff
@@ -0,0 +1,31 @@
...
+        - name: Cache node modules
+          uses: actions/cache@v3
+          id: cache-node-modules
+          with:
+              path: |
+                  node_modules
+                  shared/node_modules
+                  frontend/node_modules
+                  backend/node_modules
+              key: node-modules-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('package-lock.json') }}-${{ runner.os }}
```
Greenscreen23, line 24:
```suggestion
              key: node-modules-${{ hashFiles('package-lock.json', '**/package-lock.json') }}-${{ runner.os }}
```
https://docs.github.com/en/actions/learn-github-actions/expressions#hashfiles

.github/workflows/pipeline.yml
```diff
@@ -254,44 +188,28 @@ jobs:
...
-                  key: ${{ secrets.CACHE_VERSION }}-${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('shared/src/**') }}
+            - name: Cache node modules and shared/dist
+              uses: ./.github/actions/cache-node-modules
             - name: Cache Cypress binary
               uses: actions/cache@v3
               id: cache-cypress
               with:
                   path: ~/.cache/Cypress
-                  key: cypress-${{ secrets.CACHE_VERSION }}-${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('package-lock.json') }}
+                  key: cypress-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('package-lock.json') }}-${{ runner.os }}
```
Greenscreen23, line 202:
```suggestion
                  key: cypress-${{ hashFiles('package-lock.json', '**/package-lock.json') }}-${{ runner.os }}
```

.github/actions/cache-node-modules/action.yml
```diff
@@ -0,0 +1,31 @@
+name: Cache node_modules and dist
+description: Caches the node_modules and optionally dist.
+inputs:
+    shared-dist:
+        description: Whether to cache shared/dist too.
+        default: 'true'
```
Greenscreen23, line 6: I don't like the name of this action. The statement `uses: ./.github/actions/cache-node-modules` does not include the information that `shared/dist` is also cached. I'd suggest calling the action `cache` or `dfm-cache` and adding a second input for `node-modules` which also defaults to true.

Greenscreen23 (changes requested):

Dassderdie:
I'm not quite sure about it, but I thought GitHub calculates the pipeline minutes by rounding up the minutes of each action -> many actions that run very short cost more than few that run long.

Greenscreen23 (approved):

Greenscreen23:
> I'm not quite sure about it, but I thought GitHub calculates the pipeline minutes by rounding up the minutes of each action -> many actions that run very short cost more than few that run long.

We are a public repo: https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions
> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits.

Dassderdie:
> We are a public repo: https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

I see. I only remembered that a year ago we spent all the free pipeline minutes of the hpi-sam organization ^^
But at that time the repository was still private...

PR #729, opened 2023-03-07
Create a popup that shows all simulated regions

lukasrad02:


frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/simulated-regions-modal/simulated-regions-modal.component.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    ngOnInit(): void {
+        this.simulatedRegionIds = this.store.select(
+            createSelector(selectSimulatedRegions, (simulatedRegions) =>
+                Object.values(simulatedRegions).map((simulatedRegion) => ({
+                    name: simulatedRegion.name,
+                    id: simulatedRegion.id,
+                    simulatedRegion,
+                }))
+            )
```
Dassderdie, line 41: Don't create a new selector, but use the already existing one. You can then modify the observable's values via the rxjs pipe operators.

benn02 (approved):

PR #740, opened 2023-03-09
Fix deletion of simulated region now cascades

benn02:


shared/src/store/action-reducers/simulated-region.ts
```diff
@@ -29,6 +31,95 @@ import { TransferPointActionReducers } from './transfer-point';
...
+
+    Object.values(draftState.personnel)
+        .filter(
+            (personnel) =>
+                isInSpecificSimulatedRegion(personnel, simulatedRegionId) &&
+                relatedVehicles.includes(
+                    getElement(draftState, 'vehicle', personnel.vehicleId)
+                )
+        )
+        .forEach((personnel) => delete draftState.personnel[personnel.id]);
```
lukasrad02, line 86: You have already checked that `relatedVehicles` only includes vehicles where all materials and personnel of that vehicle are in the current simulated region. So I think the easier way would be to so something like (pseudocode, might have to be fixed to fit this situation):

```ts
realtedVehicles.forEach((vehicle) => {
    vehicle.materialIds.forEach((id) => delete draftState.material[id]);
    vehicle.personnelIds.forEach((id) => delete draftState.personel[id]);
}
```

shared/src/store/action-reducers/simulated-region.ts
```diff
@@ -29,6 +31,95 @@ import { TransferPointActionReducers } from './transfer-point';
...
+    // Find related vehicles
+
+    const relatedVehicles = Object.values(draftState.vehicles).filter(
+        (vehicle) =>
+            isInSpecificSimulatedRegion(vehicle, simulatedRegionId) &&
+            Object.keys(vehicle.materialIds).every((materialId) =>
+                isInSpecificSimulatedRegion(
+                    getElement(draftState, 'material', materialId),
+                    simulatedRegionId
+                )
```
lukasrad02, line 56: It would also be okay if the material is in the vehicle (same for the personnel).

shared/src/store/action-reducers/simulated-region.ts
```diff
@@ -146,25 +237,7 @@ export namespace SimulatedRegionActionReducers {
...
-                    (element) =>
-                        isInSpecificSimulatedRegion(element, simulatedRegion.id)
-                );
-                TransferPointActionReducers.removeTransferPoint.reducer(
-                    draftState,
-                    {
-                        type: '[TransferPoint] Remove TransferPoint',
-                        transferPointId: transferPoint.id,
-                    }
-                );
```
lukasrad02, line 166: The remove transfer point action reducer does more than just removing the transfer point, since it also handles existing transfers.

I think we should keep using its code (possibly by extracting all code from the reducer to a new function that can be called from the transfer point reducer as well as the simulated region reducer) and remove the transfer point before removing vehicles/materials/personnel, since removing the transfer point might cause elements to be added to the region.

shared/src/store/action-reducers/simulated-region.ts
benn02: Done

shared/src/store/action-reducers/simulated-region.ts
benn02: Done

lukasrad02 (commented):
Looks good to me! Just add this change to the `CHANGELOG.md` and I'll be fine with it

lukasrad02 (approved):

PR #741, opened 2023-03-10
Treat patients also if the current leader is a GF

lukasrad02:


Greenscreen23 (approved):
You could also add a note in the changelog :)

PR #767, opened 2023-03-17
Add Recurring Event

benn02:


shared/src/simulation/activities/recurring-event.ts
```diff
@@ -0,0 +1,72 @@
...
+    @IsUUID(4, uuidValidationOptions)
+    public readonly id: UUID;
+
+    @Type(...simulationEventTypeOptions)
+    @ValidateNested()
+    public readonly event: ExerciseSimulationEvent;
+
+    @IsInt()
+    @Min(0)
+    public readonly startTime: number;
```
Nils1729, line 27: You could use `nextRecurrence` instead so when creating an activity, it is clear when it triggers next.

shared/src/simulation/activities/recurring-event.ts
```diff
@@ -0,0 +1,72 @@
...
+            terminate
+        ) {
+            if (
+                draftState.currentTime >=
+                activityState.startTime + activityState.recurrenceIntervalTime
+            ) {
+                activityState.startTime += activityState.recurrenceIntervalTime;
+                sendSimulationEvent(
+                    simulatedRegion,
+                    cloneDeep(activityState.event)
```
Nils1729, line 65: You do not need to clone the event as it will be cloned in `sendSimulationEvent`. Immutable events are valid input for this function.

shared/src/simulation/activities/recurring-event.ts
```diff
@@ -0,0 +1,70 @@
...
+            activityState,
+            _tickInterval,
+            terminate
+        ) {
+            if (draftState.currentTime >= activityState.nextOccurrenceTime) {
+                activityState.nextOccurrenceTime +=
+                    activityState.recurrenceIntervalTime;
+                sendSimulationEvent(
+                    simulatedRegion,
+                    cloneDeep(activityState.event)
```
benn02, line 65:
```suggestion
                    activityState.event
```

Nils1729 (approved):

Issue #754, opened 2023-03-14
Populate the radiogram sidebar


lukasrad02:
Show radiograms in the sidebar.

### Acceptance Criteria:

- [x] Per default, the sidebar shows all radiograms that are not done
- [x] `Done` radiograms can be viewed on request (e.g. by clicking a checkbox, switching to another tab, ...)
- [x] Radiograms are sorted by their timestamp, the newest radiogram at the bottom
- [x] Accepted radiograms are highlighted (e.g. by a colored outline or by moving them to a dedicated region at the top)
- [x] For each radiogram, a box with an accept/done button and there should be a placeholder for the radiograms content
    - Which button is shown depends on the state of the radiogram

## Depends on:

- #751
- #753

PR #771, opened 2023-03-17
Create radiogram sidebar

Nils1729:


frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/simulated-regions-modal/simulated-regions-modal.component.html
```diff
@@ -8,35 +8,50 @@ <h4 class="modal-title">Simulierte Regionen</h4>
...
+    <div class="d-flex flex-row w-100 h-100">
+        <div
+            class="flex-shrink-1 mh-100 p-4 d-flex flex-column"
+            style="min-width: 20%"
         >
-            <ng-container
-                *ngFor="let id of simulatedRegionIds$ | async"
-                [ngbNavItem]="id"
+            <div class="flex-shrink-1">
+                <p>Example title</p>
```
lukasrad02, line 20: I would already specify the real title here, e.g. `<h5>Meldungen</h5>` (`h5` since the surrounding modal uses `h4` for its title).

We might also want to change the modal title and its button in the toolbar from *Simulierte Regionen* to something like *Simulations-Übersicht*, what do think about this?

frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/simulated-regions-modal/simulated-regions-modal.component.html
```diff
@@ -8,35 +8,50 @@ <h4 class="modal-title">Simulierte Regionen</h4>
...
-    <div class="h-100 d-flex flex-column">
-        <nav
-            #nav="ngbNav"
-            ngbNav
-            [(activeId)]="currentSimulatedRegionId"
-            class="nav-tabs"
+    <div class="d-flex flex-row w-100 h-100">
+        <div
+            class="flex-shrink-1 mh-100 p-4 d-flex flex-column"
+            style="min-width: 20%"
```
lukasrad02, line 17: There should also be a max width and I would suggest using units that do not depend on the screen size that much so that we know how much space we have when designing the components for the different radiograms

lukasrad02:
Please add a note to the changelog (can be short, since #754 will complete it).

lukasrad02 (approved):

PR #774, opened 2023-03-17
Test/765 import tests

benn02:
This adds migration tests that tests if all exercises from https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios/tree/main/migration-test-scenarios can be imported and migrated correctly. See the readme of that repo for more info


package.json
```diff
@@ -22,7 +22,8 @@
         "lint": "concurrently \"cd shared && npm run lint\" \"cd frontend && npm run lint\" \"cd backend && npm run lint\"",
         "lint:fix": "concurrently \"cd shared && npm run lint:fix\" \"cd frontend && npm run lint:fix\" \"cd backend && npm run lint:fix\"",
         "test:concurrently": "concurrently \"cd shared && npm run test\" \"cd frontend && npm run test\" \"cd backend && npm run test\"",
-        "test": "cd shared && npm run test && cd ../frontend && npm run test && cd ../backend && npm run test",
+        "test": "(mkdir test-scenarios || cd .) && cd test-scenarios && (git clone https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios.git || cd .) && cd digital-fuesim-manv-public-test-scenarios && git pull && cd ../../shared && npm run test && cd ../frontend && npm run test && cd ../backend && npm run test",
```
ClFeSc, line 25: Maybe consider [Git Submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) for this.

benn02: I will thanks for the hint

ClFeSc:
I suggest adding at least one of the old scenarios, possibly one created for Ahrweiler, to test the older migrations (in case this is possible from a legal point of view, they're currently stored in a private repo).

backend/test/http-migration.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+        describe(dir, () => {
+            const exercisePaths = fs.readdirSync(`${basePath}/${dir}`);
+            const exercisePathsToTest = exercisePaths.filter(
+                (exercisePath) =>
+                    exercisePath.split('-')[0] !== 'EXCLUDE' ||
+                    exercisePath.split('-')[1] !== 'FROM' ||
+                    exercisePath.split('-')[2] !== 'TEST'
+            );
+            test.each(exercisePathsToTest)(
+                'It imports %s correctly',
```
Nils1729, line 26:
```suggestion
                'It imports %s without error',
```

Nils1729:
Please add a line to the changelog and some hint to the test scenario repo and its usage in one of the READMEs.

backend/test/http-migration.spec.ts
benn02: Done

benn02:
> Please add a line to the changelog and some hint to the test scenario repo and its usage in one of the READMEs.

Done

.github/workflows/pipeline.yml
```diff
@@ -157,6 +157,58 @@ jobs:
...
+                  cat ./.env.example
+                  | sed -e "s/^DFM_DB_USER=.*$/DFM_DB_USER=${{ secrets.POSTGRES_USER }}/"
+                  | sed -e "s/^#DFM_DB_PASSWORD=.*$/DFM_DB_PASSWORD=${{ secrets.POSTGRES_PASSWORD }}/"
+                  | sed -e "s/^DFM_DB_NAME=.*$/DFM_DB_NAME=${{ secrets.POSTGRES_DB }}/"
+                  | sed -e "s/^DFM_DB_HOST=.*$/DFM_DB_HOST=127.0.0.1/" > ./.env
+            - name: Run migrations
+              run: cd backend && npm run migration:run && cd ..
+            - name: Fetch Test Scenarios
+              run: git submodule update --init --recursive
+              if: always()
```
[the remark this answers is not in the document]
benn02: They mean that this step is executed even if the previous one fails. This is useful to always run the test.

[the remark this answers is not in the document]
Nils1729: So when fetching the test scenarios fails, the job automatically succeeds?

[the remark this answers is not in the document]
benn02: No it fails, but it does the tests it only ensures that the job is not cancelled before even though it is certain it will fail

lukasrad02:
It should be referenced in the readme that test scenarios are stored in a submodule and how to clone the repo with its submodules.

Either in the installation section, step 3, for example

> 3. Clone the repo by running `git clone https://github.com/hpi-sam/digital-fuesim-manv`. To be able to run migration tests, you also have to clone the submodules: `git clone --recurse-submodules https://github.com/hpi-sam/digital-fuesim-manv`.

or in the migration tests section, for example

> The test scenarios are stored in a submodule. Use `--recurse-submodules` when cloning the repo or run `git submodule update --init --recursive` if you have cloned the repo already.

PR #776, opened 2023-03-20
Add the generate report activity

Greenscreen23:


shared/src/simulation/activities/generate-report.ts
```diff
@@ -0,0 +1,75 @@
...
+                sendSimulationEvent(
+                    simulatedRegion,
+                    activityState.collectEvent
+                );
+                activityState.hasSendEvent = true;
+                return;
+            }
+
+            sendRadiogram(activityState.radiogram, draftState);
+            terminate();
```
Nils1729, line 73: Because the if clause is not a guard clause, I would prefer having an explicit `else` branch.

shared/src/models/radiogram/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+import type { Mutable } from '../../utils';
+import { cloneDeepMutable } from '../../utils';
+import type { ExerciseRadiogram } from './exercise-radiogram';
+
+export function sendRadiogram(
+    radiogram: ExerciseRadiogram,
+    draftState: Mutable<ExerciseState>
+) {
+    const mutableRadiogram = cloneDeepMutable(radiogram);
+    mutableRadiogram.transmissionTime = draftState.currentTime;
```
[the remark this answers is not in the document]
Greenscreen23: Refactored in a different PR, see 4011ff955ae4a6a175809e4c3324e8b562a557fc

shared/src/simulation/activities/generate-report.ts
Greenscreen23: I've refactored it in 13197e9d7e83402dd099ac454ecf59ce08b6605e

Nils1729 (approved):

PR #777, opened 2023-03-20
Add report radiograms

Greenscreen23:


shared/src/models/radiogram/material-count-radiogram.ts
Nils1729, line 1: Because radiograms will be created in the reducers, we need the `id` as a constructor parameter to fill it deterministically. See https://github.com/hpi-sam/digital-fuesim-manv/pull/664#discussion_r1107193312

Greenscreen23: Done in 4bc46f3883a662581fc99b2f743ad9012026bbcf

Nils1729 (approved):

PR #786, opened 2023-03-24
Add component for personnel count radiograms

lukasrad02:


frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/radiogram-list/radiogram-card/radiogram-card-content-personnel-count/radiogram-card-content-personnel-count.component.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    radiogram$!: Observable<PersonnelCountRadiogram>;
+
+    readonly personnelCategories = [
+        'notarzt',
+        'notSan',
+        'rettSan',
+        'san',
+        'gf',
+    ] as const;
```
Greenscreen23, line 26: Could you please add this to `shared/src/models/utils/personnel-type.ts` and import it from there?
```ts
export const personnelTypes = StrictObject.keys(personnelTypeAllowedValues);
```

Its also used in the general tab, so it might also be sensible to change it there as well :)

frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/radiogram-list/radiogram-card/radiogram-card-content-personnel-count/radiogram-card-content-personnel-count.component.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    constructor(private readonly store: Store<AppState>) {}
+
+    ngOnInit(): void {
+        const selectRadiogram = createSelectRadiogram(this.radiogramId);
+        this.radiogram$ = this.store.select(
+            selectRadiogram
+        ) as Observable<PersonnelCountRadiogram>;
```
Greenscreen23, line 34: I don't think you need to use a separate variable for `selectRadiogram`, and we have never used these selectors like this.

Greenscreen23 (changes requested):

frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/radiogram-list/radiogram-card/radiogram-card-content-personnel-count/radiogram-card-content-personnel-count.component.ts
lukasrad02: I did not want to compute this dynamically since the order of the elements in the array defines the order in which the elements are shown in the user interface. Placing the array in shared could cause the UI to change suddenly and it would be hard to track the cause.

Greenscreen23 (approved):

PR #790, opened 2023-03-24
Add report of counts

benn02:


shared/src/simulation/behaviors/assign-leader.ts
```diff
@@ -35,51 +42,189 @@ export const assignLeaderBehavior: SimulationBehavior<AssignLeaderBehaviorState>
...
+                                            (material) =>
+                                                material.canCaterFor.green
+                                        )
+                                        .reduce((a, b) => a + b, 0);
+                                    radiogram.materialForPatients.red =
+                                        canCaterForRed;
+                                    radiogram.materialForPatients.yellow =
+                                        canCaterForYellow;
+                                    radiogram.materialForPatients.green =
+                                        canCaterForGreen;
```
Greenscreen23, line 151: In my opinion, you can remove the variables from above, as they serve no purpose other than to transfer one value once from a to b

shared/src/simulation/behaviors/assign-leader.ts
```diff
@@ -35,51 +42,189 @@ export const assignLeaderBehavior: SimulationBehavior<AssignLeaderBehaviorState>
...
+                                    const groupedVehicles = groupBy(
+                                        vehicles,
+                                        (vehicle) => vehicle.vehicleType
+                                    );
+                                    Object.entries(groupedVehicles).forEach(
+                                        ([vehicleType, vehicleGroup]) => {
+                                            radiogram.vehicleCount[
+                                                vehicleType
+                                            ] = vehicleGroup.length;
+                                        }
```
Greenscreen23, line 217: You could also use `map` instead of `forEach`. I think this would make it a bit more clear what is going on.

benn02: Since this executes side effects I refrained from doing so

shared/src/simulation/behaviors/assign-leader.ts
Greenscreen23: But whatever, I think we just think about code differently

benn02: Done

shared/src/simulation/behaviors/assign-leader.ts
Greenscreen23: I don't really see what side effects would be executed by something like this. Do you mind explaining (maybe as dm)
```ts
radiogram.vehicleCount = Object.fromEntries(
    StrictObject.entries(
        groupedVehicles
    ).map(([vehicleType, vehicleGroup]) => [
        vehicleType,
        vehicleGroup.length,
    ])
);
```

benn02: Done

Greenscreen23 (approved):

PR #808, opened 2023-03-27
Fix falsy values in ngIf

benn02:


Dassderdie:
@benn02 I believe we already have the [`appLet`](https://github.com/hpi-sam/digital-fuesim-manv/blob/dev/frontend/src/app/shared/directives/let.directive.ts) directive for that.

PR #810, opened 2023-03-27
Add flag that specifies whether radiograms have been answered by a behavior

lukasrad02:


Greenscreen23:
Where is the flag initially set? It should be `false` by default.

Greenscreen23 (approved):

PR #853, opened 2023-04-04
Refactoring/resolve circular dependencies

Greenscreen23:
### PR Checklist

-   [ ] ~If this PR adds or changes features or fixes bugs, this has been added to the changelog~

### DIFF
--- a/.eslint/typescript.eslintrc.cjs
+++ b/.eslint/typescript.eslintrc.cjs
@@ -7,6 +7,7 @@ module.exports = {
     extends: [
         'eslint:recommended',
         'plugin:@typescript-eslint/recommended',
+        'plugin:import/typescript'
         // TODO: doesn't work for some reason from this config and has to be imported separately
         // 'prettier',
     ],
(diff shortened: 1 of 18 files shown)


Greenscreen23:
This PR also introduces a pipeline job to check for circular dependencies

lukasrad02:
> This PR also introduces a pipeline job to check for circular dependencies

Instead of using madge as additional dependency which is only able to check the compiled files for circular dependencies, you could use the [import/no-cycle](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) ESLint rule.

Opposite to madge, this does not require additional dependencies, is able to check the TypeScript files and highlight the problematic imports in the IDE (when using the ESLint extension) and (in my experience) finds more cyclic dependencies than madge does.

When using the ESLint rule, keep im mind to add `'plugin:import/typescript'` to the `extends` array in the ESLint config for the rule to work properly (which, however, is something we should do no matter whether we add the cyclic rule or not...)

lukasrad02 (approved):

PR #919, opened 2023-04-26
Feature/798 customer feature request mark patient whose popup is open

Greenscreen23:
### PR Checklist


lukasrad02:
Toggling the popup seems to be buggy: I've noticed that sometimes when I click the same element a second time, the popup changes its location while sometimes the popup closes. I first thought that this was related to small movements of the element since I slightly moved my cursor while clicking. But observing the proposed actions on the websocket shows that even if I don't move the element, the popup behaves different from time to time.

Here is an example: The first time the popup closes is due to a slight move of the element. But the second, third and fourth time, the popup closes without the element being moved (see the websocket in the bottom-right corner) and after a varying number of clicks. I did not use any other inputs (like the Escape key) while recording this video.

https://user-images.githubusercontent.com/49586507/234797506-fc3070ea-93ef-45cd-947e-5a362002c4d2.mp4

lukasrad02 (approved):

PR #939, opened 2023-05-05
first proposal on Auto-Delete exercises

anonym-HPI:
### PR Checklist

### DIFF
--- a/backend/src/exercise/exercise-wrapper.ts
+++ b/backend/src/exercise/exercise-wrapper.ts
@@ -17,6 +17,7 @@ import {
 } from 'digital-fuesim-manv-shared';
 import type { EntityManager } from 'typeorm';
 import { LessThan } from 'typeorm';
+import { IsInt, Min } from 'class-validator';
 import { Config } from '../config';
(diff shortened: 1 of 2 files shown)


anonym-HPI:
@Dassderdie @ClFeSc What do you think of this way to implement this?

The discussion before is also here: https://github.com/hpi-sam/digital-fuesim-manv/issues/183

@Dassderdie You wrote about `setTimeout`, I did use it in some kind (used the existing `PeriodicEventHandler`), this way the server could be set to like interval over every exercise like once a day and delete exercises that have not been accessed in a while.

Also the question is how often the `sinceExerciseWithoutClients` should be updated (only when no client is connected anymore or every minute if at least one client is connected, of course it should be renamed then).

anonym-HPI:
If somebody wants to help write a test.
A test should do the following
1. Create a new exercise
2. Let a client connect to it
3. Disconnect this client
4. Wait at least Config.cleanupTime (maybe a bit more, to be sure)
5. Then try to connect again to this exercise
6. Expect that an error is returned that the exercise does not exist anymore

Another test could be:
1. An exercise is importet (without any client connecting to the exercise)
2. Follow step 4-6 above

PR #951, opened 2023-05-08
Increase cypress response timeout

Greenscreen23:
### PR Checklist

-   ~[ ] If this PR adds or changes features or fixes bugs, this has been added to the changelog~
-   ~[ ] If this PR adds new actions or other ways to alter the state, [test scenarios](https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios) have been added~

### DIFF
--- a/frontend/cypress.config.ts
+++ b/frontend/cypress.config.ts
@@ -8,5 +8,6 @@ export default defineConfig({
     e2e: {
         baseUrl: 'http://127.0.0.1:4200',
         video: false,
+        responseTimeout: 60000,
     },
 });


benn02 (commented):
Why didnt't you increase the wait on timeout as suggested in the issue?

lukasrad02:
> Why didnt't you increase the wait on timeout as suggested in the issue?

As it turned out, the Angular Dev server is already accepting requests but delays the answer until the build process is finished. So we do not have to increase the timeout for how long it may take until the server is reachable but the timeout how long an answer may take.

lukasrad02 (approved):

PR #974, opened 2023-05-10
Add events related to patient transport

benn02:
### PR Checklist

-   [x] If this PR adds new actions or other ways to alter the state, [test scenarios](https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios) have been added


shared/src/simulation/events/patient-category-transfer-to-hospital-finished.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    @IsLiteralUnion(alivePatientTriageCategoryAllowedValues)
+    readonly patientCategory: AlivePatientTriageCategory;
+
+    /**
+     * This is true, if this refers to its own one single region.
+     * This is false, if it refers to all regions managed by one behavior.
+     */
+    @IsBoolean()
+    readonly isRelatedToOwnRegion: boolean;
```
Nils1729, line 24: In my opinion, either these cases should be split into separate events or the name must be more descriptive, like `ownRegionOnly`.

shared/src/simulation/events/patient-category-transfer-to-hospital-finished.ts
```diff
@@ -0,0 +1,38 @@
...
+import type { SimulationEvent } from './simulation-event';
+
+export class PatientCategoryTransferToHospitalFinishedEvent
+    implements SimulationEvent
+{
+    @IsValue('patientCategoryTransferToHospitalFinishedEvent')
+    readonly type = 'patientCategoryTransferToHospitalFinishedEvent';
+
+    @IsLiteralUnion(alivePatientTriageCategoryAllowedValues)
+    readonly patientCategory: AlivePatientTriageCategory;
```
Nils1729, line 17: Is there any real reason why this should not be any patient category? What about `blue` or `white`? They are mostly alive, aren't they?

shared/src/simulation/events/patient-transfer-to-hospital-successful.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { SimulationEvent } from './simulation-event';
+
+export class PatientTransferToHospitalSuccessfulEvent
+    implements SimulationEvent
+{
+    @IsValue('patientTransferToHospitalSuccessfulEvent')
+    readonly type = 'patientTransferToHospitalSuccessfulEvent';
+
+    @IsLiteralUnion(alivePatientTriageCategoryAllowedValues)
+    readonly patientCategory: AlivePatientTriageCategory;
```
Nils1729, line 18: Same as above

shared/src/simulation/events/patient-transfer-to-hospital-successful.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    @IsUUID(4, uuidValidationOptions)
+    readonly originSimulatedRegion: UUID;
```
Nils1729, line 21: Which region Id does this hold? `departure` or `management`?

Nils1729 (commented):
I don't think the whole `AlivePatientTriageCategory` thing is helpful here.

shared/src/simulation/events/patient-category-transfer-to-hospital-finished.ts
benn02: I think a name like RelevantForHospitalTransportPatientCategory makes it more clear what it is. And all other patients than SK1-3 will not be relevant to the Transfer Behavior so I let them out

shared/src/simulation/events/patient-transfer-to-hospital-successful.ts
benn02: the region where the patient is from (i made the name more clear)

shared/src/simulation/events/patient-transfer-to-hospital-successful.ts
benn02: See above

shared/src/simulation/events/patient-category-transfer-to-hospital-finished.ts
benn02: Its now called `isRelatedOnlyToOwnRegion`

shared/src/simulation/events/patient-category-transfer-to-hospital-finished.ts
Nils1729: It would have been less work *and* more robust if left in. What if a patient dies while the transport is being prepared? What if the participants decide to transport a blue patient regardless?

benn02: In case 1) The patient must not be transported to not waste important capacity, In case 2) It is not relevant to the behavior and therefore this event should not be sent. 
The Transfer patients to Hospital behavior only knows 3 categorie (we could change this but that is how its currently planned) and therefor this data type should represent exactly those 3 categories

benn02: done

Nils1729 (approved):

PR #982, opened 2023-05-12
Feature/964 report behavior can send radiograms on last patient of a category

lukasrad02:
### PR Checklist

### DIFF
--- a/shared/src/simulation/behaviors/report.ts
+++ b/shared/src/simulation/behaviors/report.ts
@@ -44,7 +44,7 @@ export class ReportBehaviorState implements SimulationBehaviorState {
 
     @IsBoolean()
     public readonly reportTransferOfCategoryInSingleRegionCompleted: boolean =
-        true;
+        false;
 
     @IsBoolean()
     public readonly reportTransferOfCategoryInMultipleRegionsCompleted: boolean =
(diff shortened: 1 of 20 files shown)


shared/src/simulation/behaviors/report.ts
```diff
@@ -39,6 +42,14 @@ export class ReportBehaviorState implements SimulationBehaviorState {
...
     public readonly reportTreatmentProgressChanges: boolean = true;
 
+    @IsBoolean()
+    public readonly reportTransferOfCategoryInSingleRegionCompleted: boolean =
+        true;
+
+    @IsBoolean()
+    public readonly reportTransferOfCategoryInMultipleRegionsCompleted: boolean =
+        true;
+
```
Nils1729, line 52: In the PA template, could you please set this value to `false` so simple patient trays do not send duplicate radiograms by default?

lukasrad02: A PA template should not have the manage transports behavior at all, so this should not be a problem.

What I could (and should, if I think about this) do is to disable the single region report per default on all simulated regions, since even a self-managed region would need a management behavior (that would be assigned to the region itself and only manage this region), so the single and multiple region reports would almost be equivalent (despite from the text "this region"/"all managed regions"), but at least some radiograms would be created.

lukasrad02: I've changed the default for the single region to `false` in 3735008f.

Nils1729 (approved):

PR #984, opened 2023-05-12
Feature/968 create transport patients to hospital behavior

lukasrad02:
### PR Checklist

### DIFF
--- a/shared/src/simulation/activities/transfer-patient-to-hospital.ts
+++ b/shared/src/simulation/activities/transfer-patient-to-hospital.ts
@@ -1,0 +1,115 @@
+import { IsUUID } from 'class-validator';
+import { IntermediateOccupation, getCreate } from '../../models/utils';
+import {
+    UUID,
+    UUIDSet,
+    cloneDeepMutable,
+    uuidValidationOptions,
+} from '../../utils';
+import { IsUUIDSet, IsValue } from '../../utils/validators';
+import { sendSimulationEvent } from '../events/utils';
+import {
+    PatientTransferToHospitalSuccessfulEvent,
+    TransferPatientsInSpecificVehicleRequestEvent,
+} from '../events';
+import { catchAllHospitalId } from '../../data/default-state/catch-all-hospital';
+import { getElement } from '../../store/action-reducers/utils';
+import { Patient } from '../../models/patient';
+import type {
+    SimulationActivity,
+    SimulationActivityState,
+} from './simulation-activity';
+
+export class TransferPatientToHospitalActivityState
+    implements SimulationActivityState
+{
+    @IsValue('transferPatientToHospitalActivity' as const)
+    public readonly type = 'transferPatientToHospitalActivity';
+
+    @IsUUID(4, uuidValidationOptions)
+    public readonly id: UUID;
+
+    @IsUUIDSet()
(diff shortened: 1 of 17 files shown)


shared/src/simulation/behaviors/transfer-to-hospital.ts
```diff
@@ -0,0 +1,218 @@
...
+                            getVisiblePatientStatus(patient, draftState)
+                        ]++;
+                    });
+
+                    // The tick event itself is the last event per tick
+                    // Therefore, we can reset our state here
+                    behaviorState.patientIdsSelectedForTransfer = {};
+
+                    break;
+                }
```
Nils1729, line 196: This pattern with temporary `patientIdsSelectedForTransfer` is really confusing and computationally expensive. How hard would it be to listen on `TransferPatientsInSpecificVehicleRequestEvent` or even `PatientTransferToHospitalSuccessfulEvent` and use this to catch the last patient instead?

shared/src/simulation/behaviors/transfer-to-hospital.spec.ts
```diff
@@ -0,0 +1,590 @@
...
+    const newState = cloneDeepMutable(state);
+
+    Object.values(newState.simulatedRegions).forEach((simulatedRegion) => {
+        simulatedRegion.inEvents = [];
+    });
+
+    return newState;
+}
+
+describe('reassign treatment', () => {
```
Nils1729, line 108: :thinking:

shared/src/simulation/activities/transfer-patient-to-hospital.ts
```diff
@@ -0,0 +1,115 @@
...
+                sendSimulationEvent(
+                    transferManagementRegion,
+                    PatientTransferToHospitalSuccessfulEvent.create(
+                        Patient.getVisibleStatus(
+                            patient,
+                            draftState.configuration.pretriageEnabled,
+                            draftState.configuration.bluePatientsEnabled
+                        ),
+                        simulatedRegion.id
+                    )
```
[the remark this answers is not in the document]
lukasrad02: We have decided to keep it as it is, since

- it would require a lot of changes to pass the information that a report should be sent trough all the layers of the transfer framework
- in a well-configured scenario (i.e., all required behaviors exist on the simulated region), the transfer should not fail
- it is realistic that the success of some operation might be reported to early

shared/src/simulation/behaviors/transfer-to-hospital.ts
lukasrad02: I have to keep `patientIdsSelectedForTransfer`, since otherwise, the behavior would try to load the same patient into multiple vehicles if more than one vehicle arrives in one tick, since there are some ticks between the selection and loading of a patient.

Therefore, I will also keep the calculation of the progress in the tick event handler, since I won't be able to remove `patientIdsSelectedForTransfer` anyways.

shared/src/simulation/behaviors/transfer-to-hospital.spec.ts
lukasrad02: :facepalm:, fixed it in 85d5121c

Nils1729 (approved):

PR #986, opened 2023-05-12
Display which button was clicked on a resource request radiogram

Greenscreen23:
### PR Checklist

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/radiogram-list/radiogram-card/radigoram-card-content-resource-request/radigoram-card-content-resource-request.component.html
+++ b/frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/radiogram-list/radiogram-card/radigoram-card-content-resource-request/radigoram-card-content-resource-request.component.html
@@ -1,8 +1,11 @@
-<ng-container *ngIf="requiredResource$ | async as requiredResource">
+<ng-container *ngIf="radiogram$ | async as radiogram">
     <h5>Anfrage von Ressourcen</h5>
 
     <ng-container
-        *ngIf="(requiredResource | keys).length > 0; else noResourcesRequired"
+        *ngIf="
+            (radiogram.requiredResource | keys).length > 0;
+            else noResourcesRequired
+        "
     >
         <table class="table card-text">
             <thead>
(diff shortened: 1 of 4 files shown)


shared/src/models/radiogram/resource-request-radiogram.ts
```diff
@@ -26,6 +32,10 @@ export class ResourceRequestRadiogram implements Radiogram {
     @ValidateNested()
     readonly status: ExerciseRadiogramStatus;
 
+    @IsOptional()
+    @IsBoolean()
+    readonly resourcesPromised?: boolean;
```
lukasrad02, line 37: Could you please add a migration for this property to ensure that this key is present on the class and not implicitly set to `undefined` since it is missing.

frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/radiogram-list/radiogram-card/radigoram-card-content-resource-request/radigoram-card-content-resource-request.component.html
```diff
@@ -14,31 +17,54 @@ <h5>Anfrage von Ressourcen</h5>
...
-                type="button"
-                [disabled]="(enableActionButtons$ | async) === false"
-                (click)="denyRequest()"
-                class="btn btn-danger col-sm mx-1"
+            <ng-container
+                *ngIf="
+                    radiogram.resourcesPromised === undefined;
+                    else requestAnswered
+                "
             >
```
lukasrad02, line 41: If the radiogram is in done status, the buttons should not be displayed even if the success of the request is unknown. I would suggest to show a text that the success in unknown instead.

lukasrad02 (changes requested):

shared/src/models/radiogram/resource-request-radiogram.ts
lukasrad02: And keep in mind to add test scenarios

Greenscreen23:
Merge hpi-sam/digital-fuesim-manv-public-test-scenarios#11 after this

frontend/src/app/pages/exercises/exercise/shared/simulated-region-overview/radiogram-list/radiogram-card/radigoram-card-content-resource-request/radigoram-card-content-resource-request.component.html
lukasrad02, line 1: I would suggest to use negation instead of `=== false` at some if statements in this file.

For the `[disabled]` property of the buttons, the changed behavior would be even better in my opinion, since I'd rather like to disable the buttons if `(enableActionButtons$ | async)` is `undefined` so that one can't click them by accident while the radiogram isn't even selected.

In line 66 (`*ngIf="radiogram.resourcesPromised === false"`) it wouldn't change the behavior, but I think the code would look better.

Just for the `showAnswer$` I don't have any preference whether the buttons or the text should be visible while is is `undefined`, so you may keep it like it is if you want to show the buttons in this case.

lukasrad02 (approved):

PR #1016, opened 2023-05-24
Create vehicle logs

Greenscreen23:
### PR Checklist

### DIFF
--- a/shared/src/models/utils/occupations/exercise-occupation.ts
+++ b/shared/src/models/utils/occupations/exercise-occupation.ts
@@ -48,3 +48,15 @@ export const occupationTypeOptions: Parameters<typeof Type> = [
         },
     },
 ];
+
+export const occupationToGermanDictionary: {
+    [key in ExerciseOccupationType]: string;
+} = {
+    noOccupation: 'Das Fahrzeug wird nicht genutzt',
+    intermediateOccupation: 'Das Fahrzeug wird gerade übergeben',
+    unloadingOccupation: 'Das Fahrzeug wird gerade ausgeladen',
+    loadOccupation: 'Das Fahrzeug wird gerade beladen',
+    waitForTransferOccupation: 'Das Fahrzeug wartet auf den Transfer',
+    patientTransferOccupation:
+        'Das Fahrzeug ist für den Transport von Patienten ins Krankenhaus reserviert',
+};
--- a/shared/src/models/utils/occupations/exercise-occupation.ts
+++ b/shared/src/models/utils/occupations/exercise-occupation.ts
@@ -52,11 +52,11 @@ export const occupationTypeOptions: Parameters<typeof Type> = [
 export const occupationToGermanDictionary: {
     [key in ExerciseOccupationType]: string;
 } = {
-    noOccupation: 'Das Fahrzeug wird nicht genutzt',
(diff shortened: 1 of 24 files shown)


shared/src/models/utils/tag-helpers.ts
```diff
@@ -134,3 +136,80 @@ export function createTreatmentProgressTag(
...
+    );
+}
+
+export function createOccupationTag(
+    _draftState: Mutable<ExerciseState>,
+    occupation: ExerciseOccupation
+): Tag {
+    return new Tag(
+        'Tätigkeit',
+        'white',
```
lukasrad02, line 168: Tags with white background might not be recognizable as tags...

shared/src/models/utils/tag-helpers.ts
```diff
@@ -134,3 +136,80 @@ export function createTreatmentProgressTag(
...
+        'black',
+        occupationToGermanDictionary[occupation.type],
+        occupation.type
+    );
+}
+
+export function createVehicleActionTag(
+    _draftState: Mutable<ExerciseState>,
+    vehicleAction: 'arrived' | 'departed' | 'load' | 'unload'
```
lukasrad02, line 177: Could you plase normalize the strings so thex're all in the same tense (e.g. `loaded` and `unloaded`)?

shared/src/models/utils/occupations/exercise-occupation.ts
```diff
@@ -48,3 +48,15 @@ export const occupationTypeOptions: Parameters<typeof Type> = [
...
+export const occupationToGermanDictionary: {
+    [key in ExerciseOccupationType]: string;
+} = {
+    noOccupation: 'Das Fahrzeug wird nicht genutzt',
+    intermediateOccupation: 'Das Fahrzeug wird gerade übergeben',
+    unloadingOccupation: 'Das Fahrzeug wird gerade ausgeladen',
+    loadOccupation: 'Das Fahrzeug wird gerade beladen',
+    waitForTransferOccupation: 'Das Fahrzeug wartet auf den Transfer',
+    patientTransferOccupation:
+        'Das Fahrzeug ist für den Transport von Patienten ins Krankenhaus reserviert',
```
lukasrad02, line 61: I think these strings are pretty long for a tag. For example, "Das Fahrzeug" is completely redundant information. You could just say: "nicht genutzt", "in Übergabe", ...

lukasrad02 (changes requested):

shared/src/store/action-reducers/hospital.ts
```diff
@@ -137,7 +137,7 @@ export namespace HospitalActionReducers {
                 logVehicle(
                     draftState,
                     [createHospitalTag(draftState, hospitalId)],
-                    'Ein Fahrzeug hat ein Krankenhaus angefahren',
+                    `${vehicle.name} hat ein Krankenhaus angefahren`,
```
lukasrad02, line 140: 1. Why did you change it only here and not for all the other log messages?
2. Why do you query for the vehicle name instead of letting the helper do this? (The way you're doing it right now requires having an instance of the vehicle while for doing it in the helper, the vehicle id would suffice.)

lukasrad02 (changes requested):

PR #1020, opened 2023-05-25
Track a patient

Greenscreen23:
### PR Checklist

### DIFF
--- a/shared/src/store/action-reducers/exercise.ts
+++ b/shared/src/store/action-reducers/exercise.ts
@@ -134,7 +134,7 @@ export namespace ExerciseActionReducers {
                     logPatient(
                         draftState,
                         [],
-                        `Die Sichtungskategorie von ${currentPatient.personalInformation.name} hat sich geändert.`,
+                        `Die Sichtungskategorie des Patienten hat sich geändert.`,
                         currentPatient.id
                     );
                 }
@@ -252,17 +252,17 @@ function evaluateTreatmentReassignment(
                     .map(([personnelType]) =>
                         createPersonnelTypeTag(draftState, personnelType)
                     ),
-                `Die Einsatzkräfte zu diesem Patienten wurden neu zugeteilt. Er hat nun die folgenden Einsatzkräfte: ${StrictObject.entries(
-                    newTreatmentAssignment[patientId]!
-                )!
-                    .filter(([, count]) => count > 0)
-                    .map(
-                        ([personnelType, count]) =>
-                            `${count.toPrecision(2)} ${
-                                personnelTypeNames[personnelType]
-                            }`
-                    )
-                    .join(', ')}.`,
+                `Diese Einsatzkräfte wurden dem Patienten neu zugeteilt: ${
+                    StrictObject.entries(newTreatmentAssignment[patientId]!)!
+                        .filter(([, count]) => count > 0)
+                        .map(
+                            ([personnelType, count]) =>
+                                `${+count.toFixed(2)} ${
+                                    personnelTypeNames[personnelType]
+                                }`
+                        )
+                        .join(', ') ?? 'Keine Einsatzkräfte'
+                }.`,
                 patientId
             );
         });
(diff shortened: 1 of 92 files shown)


shared/src/store/action-reducers/exercise.ts
```diff
@@ -171,3 +194,76 @@ function refreshTransfer(
...
+                    draftState.previousTreatmentAssignment![patientId]?.[
+                        personnelType
+                    ]
+            )
+        )
+        .forEach((patientId) => {
+            logPatient(
+                draftState,
+                StrictObject.entries(newTreatmentAssignment[patientId]!)
+                    .filter(([personnelType, count]) => count > 0)
```
lukasrad02, line 251: Discard the first parameter

lukasrad02 (changes requested):

shared/src/store/action-reducers/exercise.ts
```diff
@@ -171,3 +194,76 @@ function refreshTransfer(
...
+                    newTreatmentAssignment[patientId]!
+                )!
+                    .filter(([, count]) => count > 0)
+                    .map(
+                        ([personnelType, count]) =>
+                            `${count.toPrecision(2)} ${
+                                personnelTypeNames[personnelType]
+                            }`
+                    )
+                    .join(', ')}.`,
```
lukasrad02, line 265: I think you could omit the "er hat..." part, since it makes the message unnecessary long. Just say "... neu zugeteilt: X NA, Y NFS, ...".

You should also handle empty personnel (it currently just says "die folgenden Einsatzkräfte: .") and display integer values without a trailing ".0".

shared/src/store/action-reducers/exercise.ts
```diff
@@ -125,6 +131,12 @@ export namespace ExerciseActionReducers {
                     currentPatient.visibleStatusChanged
                 ) {
                     updateTreatments(draftState, currentPatient);
+                    logPatient(
+                        draftState,
+                        [],
+                        `Die Sichtungskategorie von ${currentPatient.personalInformation.name} hat sich geändert.`,
```
lukasrad02, line 137: You're inconsistent whether you include the patient name in the message or not. For example, you don't include the name in the treatment reassignment log messages.

By the way, using the name does not make that much sense, as there is currently no way to find out the name of a patient who is part of the simulation (except of moving it to the map, opening the popup, reading the name and putting it back to the simulated region). So maybe we should implement #1022 (was just an idea in our internal list for a couple of weeks but it seems to be worth implementing it now).

lukasrad02 (changes requested):

shared/src/store/action-reducers/exercise.ts
```diff
@@ -171,3 +194,76 @@ function refreshTransfer(
...
+                `Diese Einsatzkräfte wurden dem Patienten neu zugeteilt: ${
+                    StrictObject.entries(newTreatmentAssignment[patientId]!)!
+                        .filter(([, count]) => count > 0)
+                        .map(
+                            ([personnelType, count]) =>
+                                `${+count.toFixed(2)} ${
+                                    personnelTypeNames[personnelType]
+                                }`
+                        )
+                        .join(', ') ?? 'Keine Einsatzkräfte'
```
lukasrad02, line 264: This doesn't work since `[].join()` does not return undefined.

shared/src/store/action-reducers/exercise.ts
lukasrad02: What do you think about the mentioned issue?

shared/src/store/action-reducers/exercise.ts
Greenscreen23: You're right

shared/src/store/action-reducers/exercise.ts
lukasrad02: We've decided to work on the issue so this PR can be merged as it is for now and the tags will be updated when the new IDs are ready

lukasrad02 (approved):

PR #1028, opened 2023-05-31
Feature/1022 add identity to patients

lukasrad02:
### PR Checklist


lukasrad02:
It might be useful to add the identifier to the `HospitalPatient`, too, so patients can be tracked even when they're not longer on the map.

frontend/src/app/pages/exercises/exercise/shared/exercise-settings/exercise-settings-modal/exercise-settings-modal.component.html
```diff
@@ -62,24 +62,49 @@ <h4 class="modal-title">Einstellungen</h4>
...
-                    (ngModelChange)="setBluePatientsFlag($event)"
-                    class="form-check-input"
-                    type="checkbox"
-                    data-cy="settingsBluePatientsCheckbox"
+                    #patientIdentifierPrefixInput="ngModel"
+                    [ngModel]="configuration.patientIdentifierPrefix"
+                    (appSaveOnTyping)="updatePatientIdentifierPrefix($event)"
+                    class="form-control"
+                    type="text"
+                    placeholder="B-"
```
benn02, line 104: The Placeholder could be confusing because it could seem like the Prefix is B- when it is in fact not existing. Maybe put something like Your prefix here, e.g. B.

lukasrad02:
> 1. It would be nice to Inform the user that once a patient is created the change in prefix does not apply to that patient.

Good idea, I've added a notice for this.

> 2. It would be nice to limit the length of the prefix so that a user can not mess up the UI accidentally

I've added a warning that shows up for prefixes with a length > 3 (since I think two letters and the dash will occur frequently). I do not want to enforce a specific limit as the number of letters causing layout problems could differ from user to user (different fonts etc) and there are legitimate reasons to choose longer prefixes (e.g. realism).

> 3. When using a prefix like `Berlin-` In the Patient Tab, the UI converts to a scrollbar even in the modal. I would suggest either making it bigger or talk to Christian whether it is acceptable to leave one of the 4 displayed information points out.

I've slightly increased the size so that prefixes with up to 3 characters (the soft limit from above) should fit into the table (with enough space left if people have a larger monospace font than me).

> 4. When using this prefix again, the patient popup does weird things like displaying a white border. Maybe make the top of the popup higher or remove one of the information points.

This happens since the `max-width` rule was only applied to the popup content and not the header. I've tried limiting the header length and cutting off long names with an ellipsis (`…`), but I didn't manage to get this running, since there are too many other elements in the header that should not be truncated.

What I've finally done was to remove the name from the header entirely. It is now displayed in the popup body instead. I think this should be fine since the name was only used to identify patients (which can now be done trough the identifier).

frontend/src/app/pages/exercises/exercise/shared/exercise-settings/exercise-settings-modal/exercise-settings-modal.component.html
lukasrad02: Changes it in https://github.com/hpi-sam/digital-fuesim-manv/pull/1028/commits/ab547349eb22da6adf4e81c7a48f07aeccf0994c

benn02 (approved):

PR #1050, opened 2024-10-01
Adjust frontend components content

fhuebscher:


christianschaeffer:
I know we already discussed this elsewhere, but let's clean up this PR...

Regrading the specific changes:
1. The assets (imprint and privacy) are dependent on the server operator and should be deployment-specific, and not part of the general repo. I guess we could add the license fine as default proposal for all deployments.
2. The mail contact address in `exercise.component.html` has already be updated in #1058, the one in `footer.component.html` is still open.
3. Regarding the docker file chances and the new `build:deployment` in `package.json`, I cannot really judge the difference and necessity.

@fhuebscher please either modify your PR or close it if you think the remaining changes are not worth it. Since we have a critical upcoming bugfix (i.e. need to make a hotfix release) we might add the stuff from this PR (if you think anything should remain).

@lukasrad02 please advice if you have something to add.

lukasrad02:
> 2. The mail contact address in `exercise.component.html` has already be updated in [Release v0.8.0 #1058](https://github.com/hpi-sam/digital-fuesim-manv/pull/1058), the one in `footer.component.html` is still open.

Oh, I missed that one. Will be fixed by #1063.

PR #1055, opened 2025-04-12
Fix Angular dev server reload on shared package change

lukasrad02:
Due to the Angular updates from #1051, we're now using Vite as build system. Vite [prebundles all dependencies](https://angular.dev/tools/cli/build-system-migration#prebundling) to enable fast reloads and only rebuilds them when the `package-lock.json` changes. Hence, changes to code in the shared package are not reflected in the frontend (see <https://github.com/hpi-sam/digital-fuesim-manv/pull/522#issuecomment-2796336803>).

This PR excludes the shared package from prebundling, as well as its (direct and indirect) CommonJS dependencies to resolve this issue. Without excluding the CJS dependencies, they're not found during build. [Vite recommends](https://vite.dev/config/dep-optimization-options.html#optimizedeps-exclude) to explicitly include CJS dependencies of excluded ESM dependencies, but Angular does not expose the `include` option.

---

### PR Checklist

### DIFF
--- a/frontend/angular.json
+++ b/frontend/angular.json
@@ -149,6 +149,17 @@
                             "buildTarget": "digital-fuesim-manv:build:development"
                         }
                     },
+                    "options": {
+                        "prebundle": {
+                            "exclude": [
+                                "digital-fuesim-manv-shared",
+                                "@noble/hashes/sha256",
+                                "@noble/hashes/crypto",
+                                "rbush-knn",
+                                "tinyqueue"
+                            ]
+                        }
+                    },
                     "defaultConfiguration": "development"
                 },
                 "extract-i18n": {


frontend/angular.json
```diff
@@ -149,6 +149,17 @@
...
                         }
                     },
+                    "options": {
+                        "prebundle": {
+                            "exclude": [
+                                "digital-fuesim-manv-shared",
+                                "@noble/hashes/sha256",
+                                "@noble/hashes/crypto",
+                                "rbush-knn",
+                                "tinyqueue"
```
ClFeSc, line 159: Uh, I don't like having to keep such a list here. But I guess there is no other way to achieve this. And at least it will yell quite extensive at the developer when a *new* dependency has to be added here.

ClFeSc (approved):
Verified the change solves the issue.

ClFeSc (approved):

PR #1066, opened 2025-04-28
fix: no longer transform import before sending it to the backend

### LINKED ISSUE #1065: Importing an exercise is not working through the frontend
When using the frontend to import an exercise the import will fail since #1055. This probably stems from a different import behavior of modules with `vite`. Note that importing the same exports directly into the backend works as expected.

ClFeSc:
Previously, the frontend would convert the JSON import to an instance before sending it to the backend. This, however, lead to the backend rejecting the import as an instance is only meant to be used during validation. Now, only the backend validates the import, with the frontend just passing on the JSON.

This problem was introduced in #1055 as before that the conversion to an instance was faulty in a way that the backend still understood and accepted the imports.

Fixes #1065

### PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 
 ## [Unreleased]
 
+### Fixed
+
+- Imports of exercise once again can be performed in the frontend
+
 ## [0.8.0] - 2025-04-14
 
 ### Added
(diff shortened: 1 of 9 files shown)


lukasrad02 (approved):
LGTM.

As #1065 only handles the specific bug when importing trough the frontend, I've created #1067 to track the root cause for this issue (which is not resolved by this PR, correct me if I'm wrong)

PR #1073, opened 2025-05-05
Fix/1072 map interactions are not rolled back in case of an error

### LINKED ISSUE #1072: Map interactions are not rolled back in case of an error
When moving elements on the map, the actions are not applied optimistically on the client, so there is no state change. However, the element stays at its new location as its not actively moved back.

Thus, if applying the action fails, there is no mechanism to undo the move, as no state change has ever happened on the client, so the element stays at its new position on the map. This makes it easy to overlook errors, as it seems as everything worked fine

lukasrad02:
When an element is moved to the trash bin to remove it from the map, both a move action and a delete action are sent, so if the delete fails, the element will stay at the new location. As the move is processed first, we would have to add some additional logic to handle this edge case.

I would argue that its okay do not fix this, as in contrast to the failed move (where users might not notice that the action has failed, as the element is shown at the new position on their map), a failed delete action will be noticed by the users and if there were about to delete the element, its also not important to restore its exact location. Do you agree with this (or do you have an idea how to fix this part of the bug), @ClFeSc?

Closes #1072 

### PR Checklist


ClFeSc (changes requested):
While it solves the issue in most cases this has two problems I'm not sure I like:

1. The code is a little brittle I think. Having to remember to make the move optimistic is something I'd rather not have as it is not really obvious at first glance why this is required. For example by changing this parameter from a function proposing an action to a function returning the action, proposing in `frontend/src/app/pages/exercises/exercise/shared/exercise-map/feature-managers/moveable-feature-manager.ts` instead, always optimistically.
2. The rollback does not work in case of incorrect IDs. I've tested it with replacing `patient.id` with `uuid()`. This obviously fails the action but does not revert the element's position. It, however, is enough to fix this issue in the problem with the alarm groups, so only if the ID used in the action sent by the frontend does not exist in the frontend this breaks.

ClFeSc:
> Do you agree with this (or do you have an idea how to fix this part of the bug), @ClFeSc?

In my opinion this is a good solution for the moves. For the other actions like inserting into a vehicle or deleting, I think the better solution would be to not send move actions in the first place. But this is definitely nothing for this PR and also, as you've said, something the people are more likely to notice themselves.

lukasrad02:
Thank you for the feedback!

> The code is a little brittle I think. Having to remember to make the move optimistic is something I'd rather not have as it is not really obvious at first glance why this is required. For example by changing this parameter from a function proposing an action to a function returning the action

I see you point, but I wonder whether that kind of refactoring might introduce other problems. The current approach of a generic move callback provides some more flexibility than just being able to construct an action. For example, this could limit whether/how we could fix "moving an element to the trash bin sends a move action first" in the future.

So if I do this refactoring, I would propose to still allow this function to perform side effects and to decide to not return an action. Would this be fine for you?

> The rollback does not work in case of incorrect IDs. I've tested it with replacing `patient.id` with `uuid()`. This obviously fails the action but does not revert the element's position. It, however, is enough to fix this issue in the problem with the alarm groups, so only if the ID used in the action sent by the frontend does not exist in the frontend this breaks.

I'm not sure whether I understand you correctly: To me, this reads as a general reminder without a call to action, as the wrong UUID is a problem by itself (and thus, this bugfix only treats symptoms instead of causes, which does not have to be perfect). Do I understand this correctly?

lukasrad02:
> But this is definitely nothing for this PR

Moved to #1075

ClFeSc:
> I see you point, but I wonder whether that kind of refactoring might introduce other problems. The current approach of a generic move callback provides some more flexibility than just being able to construct an action. For example, this could limit whether/how we could fix "moving an element to the trash bin sends a move action first" in the future.
> 
> So if I do this refactoring, I would propose to still allow this function to perform side effects and to decide to not return an action. Would this be fine for you?

Hm, I see what you mean. What side effects would you expect from this function, though? And while I can see the point of not returning any action to not even propose one, I'm not sure this would be a good approach to preventing deletions from emitting move actions first as I think this would be better suited as a check directly in the movable feature manager and not in each feature manager separately.

> I'm not sure whether I understand you correctly: To me, this reads as a general reminder without a call to action, as the wrong UUID is a problem by itself (and thus, this bugfix only treats symptoms instead of causes, which does not have to be perfect). Do I understand this correctly?

Yes, I think it would be fine to leave it as it is.

lukasrad02:
> Hm, I see what you mean. What side effects would you expect from this function, though?

Potential side effects would be features like:
- Undo-able moves (e. g. showing a "are you sure" popup in case of especially critical actions) and deletions
- Limiting the distance of a single move
- Disallowed move targets (e. g. vehicles may only be moved onto roads)
- Things related to the technische Rettung that some users are asking for (e. g. allow moving obstacles only if there is enough personnel available)

I haven't thought about this in detail and some of these features might better to implement in a different way (for example in the feature manager of the element at the target location)

> And while I can see the point of not returning any action to not even propose one, I'm not sure this would be a good approach to preventing deletions from emitting move actions first as I think this would be better suited as a check directly in the movable feature manager and not in each feature manager separately.

For the specific case of deleting elements, I agree that this should not be implemented in every single feature manager, but some of the examples above are more specific to only certain element types

ClFeSc:
> > Hm, I see what you mean. What side effects would you expect from this function, though?
> 
> Potential side effects would be features like:
> 
>     * Undo-able moves (e. g. showing a "are you sure" popup in case of especially critical actions) and deletions
> 
>     * Limiting the distance of a single move
> 
>     * Disallowed move targets (e. g. vehicles may only be moved onto roads)
> 
>     * Things related to the technische Rettung that some users are asking for (e. g. allow moving obstacles only if there is enough personnel available)
> 
> 
> I haven't thought about this in detail and some of these features might better to implement in a different way (for example in the feature manager of the element at the target location)
> 
> > And while I can see the point of not returning any action to not even propose one, I'm not sure this would be a good approach to preventing deletions from emitting move actions first as I think this would be better suited as a check directly in the movable feature manager and not in each feature manager separately.
> 
> For the specific case of deleting elements, I agree that this should not be implemented in every single feature manager, but some of the examples above are more specific to only certain element types

Fair enough. Keep it like it is then. Update the branch and this should be ready.

ClFeSc (approved):

Issue #1077, opened 2025-05-24
`if: always()` in GitHub workflows causes additional, confusing errors


lukasrad02:
Most of our pipeline jobs currently rely on the `build` job populating the cache with `node_modules`. This is generally fine, since these jobs all have `needs: build` set, so the cache entry will be present.

However, we have encountered pipeline runs where the subsequent jobs did not find the expected cache entry (in private forks, hence no link/reference to the pipeline run). We should consider handling those cases with an `if: ${{ steps.cache-build.outputs.node-modules-cache-hit != 'true' }}` step that prints some error message and cancels the job to provide a more clear explanation of why the job failed. Otherwise, the pipeline will just fail at some later point when trying to load node modules.

This proposal only aims to improve developer experience and does not affect the stability/reliability of the pipeline. Investigating why the cache entry could not be found is out of scope for this issue; the pipeline will still fail in this case even with this issue addressed. Hence, I consider it low-prio.

lukasrad02:
Besides the cache, there are some more instances where the `if: always()` causes issues (e. g., if docker containers could not be started).

In some jobs, we can remove the always-constraints. For example, in the `migration-test` job, there is no need to run "Fetch Test Scenarios" and "Run Migration Tests" if any of the previous steps failed (in contrast to the `test` and `lint` jobs, where we actually want to run all tests/formatters/linters no matter whether one of them has already failed)

PR #1089, opened 2025-06-23
Improve CI

lukasrad02:
This PR adds some improvements to the CI pipeline:

- Use random port for PostgreSQL (test, migration test, cypress) to support parallel execution (especially on self-hosted runners)
- Use tmpfs for PostgreSQL data to speed up DB operations (especially recreating an empty database for each test)
- Also compile front- and backend code in the pipeline. So far, only shared was built, so errors in the front- or backend code could stay undetected
- Remove some unnecessary `if: always()` statements (see <https://github.com/hpi-sam/digital-fuesim-manv/issues/1077#issuecomment-2934131597>)

### PR Checklist

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -217,10 +217,8 @@ jobs:
               run: cd backend && npm run migration:run && cd ..
             - name: Fetch Test Scenarios
               run: git submodule update --init --recursive
-              if: always()
             - name: Run Migration Tests
               run: cd backend && npm run test:ci-migration; cd ..
-              if: always()
 
     cypress:
         timeout-minutes: 20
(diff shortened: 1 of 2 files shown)


ClFeSc (approved):
If you are sure the tmpfs is working as intended, go ahead and merge. I'm just curious because the times did not improve and your code differs from what I've found online, e.g. here: https://stackoverflow.com/a/75726458/13849454

lukasrad02:
> If you are sure the tmpfs is working as intended, go ahead and merge. I'm just curious because the times did not improve and your code differs from what I've found online, e.g. here: https://stackoverflow.com/a/75726458/13849454

Fair point. I haven't checked whether my changes actually achieve the desired result. Since `volumes` in GitHub workflow services don't support `tmpfs` mounts, I've used the plain Docker command line option as documented at <https://docs.docker.com/engine/storage/tmpfs/>. I also found the SO post you've mentioned, but prefer my approach, since it moves the responsibility for the tmpfs to docker instead of relying on the existence of `/dev/shm`.

I suppose there is just no measurable improvement as the data to be written is rather enough and possibly wasn't written to the disk even before the change but remained in the system's IO cache

lukasrad02:
Do you want me to conduct some additional analysis?

Honestly, I don't know what to look at besides the timings and since those didn't get worse, I would consider it acceptable

ClFeSc:
No, it's fine. Thanks!

PR #1120, opened 2025-10-01
Migrate to ESLint flat config format

lukasrad02:
Unfortunately, the `total-functions` plugin seems to not be maintained any longer and is not compatible with the latest ESLint

### PR Checklist


.eslint/typescript.eslintrc.mjs
```diff
@@ -291,7 +285,12 @@ module.exports = {
         ],
         '@typescript-eslint/explicit-function-return-type': 'off',
         '@typescript-eslint/explicit-module-boundary-types': 'off',
-        '@typescript-eslint/explicit-member-accessibility': 'warn',
+        '@typescript-eslint/explicit-member-accessibility': [
```
ClFeSc, line 288: What is the reason for this change from `warn` to `off` with one set to `explicit`?

backend/eslint.config.js
```diff
@@ -0,0 +1,46 @@
...
+
+export default defineConfig([
+    includeIgnoreFile(gitignorePath),
+    {
+        files: ['**/*.ts', '**/*.js'],
+
+        languageOptions: {
+            parserOptions: {
+                projectService: true,
+                tsconfigRootDir: import.meta.dirname,
```
ClFeSc: `__dirname`?

benchmark/eslint.config.js
```diff
@@ -0,0 +1,46 @@
...
+
+export default defineConfig([
+    includeIgnoreFile(gitignorePath),
+    {
+        files: ['**/*.ts', '**/*.js'],
+
+        languageOptions: {
+            parserOptions: {
+                projectService: true,
+                tsconfigRootDir: import.meta.dirname,
```
ClFeSc: `__dirname`?

frontend/eslint.config.js
```diff
@@ -0,0 +1,108 @@
...
+    includeIgnoreFile(gitignorePath),
+    {
+        files: ['**/*.ts', '**/*.js'],
+
+        ignores: ['cypress.config.ts'],
+
+        languageOptions: {
+            parserOptions: {
+                projectService: true,
+                tsconfigRootDir: import.meta.dirname,
```
ClFeSc: `__dirname`?

frontend/eslint.config.js
```diff
@@ -0,0 +1,108 @@
...
+        },
+
+        processor: angular.processInlineTemplates,
+    },
+    {
+        files: ['**/*.js', 'cypress.config.ts'],
+
+        extends: [ts.configs.disableTypeChecked],
+
+        // Manually disable all rxjs-x and rxjs-angular-x rules that need type information
```
ClFeSc, line 59: Is this really the best solution?

shared/src/models/radiogram/status/radiogram-done-status.ts
```diff
@@ -14,7 +14,7 @@ export class RadiogramDoneStatus implements RadiogramStatus {
     /**
      * @deprecated Use {@link create} instead.
      */
-    // eslint-disable-next-line @typescript-eslint/no-useless-constructor, @typescript-eslint/no-empty-function
+
```
ClFeSc: Remove empty line.

shared/src/models/radiogram/status/radiogram-unread-status.ts
```diff
@@ -14,7 +14,7 @@ export class RadiogramUnreadStatus implements RadiogramStatus {
     /**
      * @deprecated Use {@link create} instead.
      */
-    // eslint-disable-next-line @typescript-eslint/no-useless-constructor, @typescript-eslint/no-empty-function
+
```
ClFeSc: Remove empty line.

shared/src/utils/validators/is-patients-per-uuid.ts
```diff
@@ -15,7 +15,7 @@ export const isPatientsPerUUID = createMapValidator<
 >({
     // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
     keyValidator: <(key: unknown) => key is UUID>((key) => isUUID(key, 4)),
-    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+
```
ClFeSc: Remove empty line.

shared/eslint.config.js
```diff
@@ -0,0 +1,46 @@
...
+
+export default defineConfig([
+    includeIgnoreFile(gitignorePath),
+    {
+        files: ['**/*.ts', '**/*.js'],
+
+        languageOptions: {
+            parserOptions: {
+                projectService: true,
+                tsconfigRootDir: import.meta.dirname,
```
ClFeSc: `__dirname`?

.eslint/typescript.eslintrc.mjs
lukasrad02: I don't remember any important reason for this change.

However, when I switch back to `warn`, there are over 800 warnings in the frontend only. The warnings are also triggered on files that haven't changed recently, so it seems like this rule has never worked at all (some rules actually did break after the dependency updates in #1051 due to a mismatch of the TS version and the ESLint/TS-ESLint version, but). The changelog of ts-eslint does not mention any change of the default value of the `accessibility` option, so this can also not be the reason for the sudden high number of warnings.

I think I noticed the high number of warnings triggered by this rule when updating the config files and disabled the rule temporarily to remove the clutter from the stdout. I also did this with some other rules but must have forgotten to re-enable this one afterwards.

Because of the high number of warnings, I would suggest the following options:

1. Leave the rule disabled (as it was obviously never used/needed)
2. Enable the rule in a separate PR (since checking a sensible visibility for all >800 occurrences will be cumbersome and simply assigning the default `public` defeats the purpose of this rule)

Personally, I prefer explicit access modifiers and thus option 2

backend/eslint.config.js
lukasrad02: Oh, nice one. Didn't notice the redundancy after combining auto-converted and manually rewritten config.

Fixed in 98e9479336847205f54d9cae37a1440d3fcd813c (also for occurrences in the other configs)

shared/src/models/radiogram/status/radiogram-done-status.ts
lukasrad02: Removed all empty lines in 3ae00c5e03b2eab49bb3be81c164c926622dbe13

frontend/eslint.config.js
lukasrad02: I think so.

There is only one alternative I came up with: Move all `rxjs-x` and `rxjs-angular-x` rules in the shared configs to a dedicated file that is only referenced for TS files, not JS. However, in this case we still would have issues with the rules from `rxjsX.configs.recommended`

lukasrad02: But I didn't find an equivalent of `ts.configs.disableTypeChecked` for `rxjs-x`/`rxjs-angular-x`

.eslint/typescript.eslintrc.mjs
ClFeSc: Sure, sounds reasonable.

frontend/eslint.config.js
ClFeSc: Okay, then leave it as it is, I would say.

.eslint/typescript.eslintrc.mjs
ClFeSc: I've opened #1121 for this.

ClFeSc (approved):

Issue #1095, opened 2025-07-16
Selector can return undefined despite its type definition


lukasrad02:
The `createSelectElementFromMapFactory` creates selectors to select basic elements (patients, vehicle, personnel, …) from the state based on their UUID. The created selectors contain an assertion that the element is defined. This is not accurate, as for example, one might still have the UUID of an already deleted element.

Even defensive code does not protect against access to undefined values in these cases. For example, in the user interface added in #1076, one can trigger access to `undefined` by deleting a simulated region while its opened in the IFS modal. Although the modal switches back to the overview tab if a region does not exist any longer, the observables on the simulated region already trigger with the new value (`undefined`), thus causing some errors printed to the console.

Changing the type definition to be more accurate might require bigger changes, as the base selectors created by `createSelectElementFromMapFactory` are used in many places and also as the base for many other selectors.

PR #1124, opened 2025-10-27
Generalize materials and personnel

### LINKED ISSUE #1123: Generalise materials and personnel
- [x] Introduce Material in the same way as VehicleType
- [x] Introduce MaterialTemplate
- [ ] Use new types with existing features

Use: https://github.com/hpi-sam/swt25-projekt-4/commit/07fd071527c9eeacbc7e278de59d314aeef7c394#diff-347b6f9f37ca99d23a7a65486991ca24a6b5dc8d16663324057de846a6d78e1a

hansegucker:
Close #1123 

### PR Checklist

- [ ] If this PR adds new actions or other ways to alter the state, [test scenarios](https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios) have been added.

### Screenshots

grafik


hansegucker:
Test scenarios are still missing, but I want to wait with that until #1125 is merged.

lukasrad02:
This PR also contains the Webstorm config files that should be added in #1128. Is this on purpose?

(Since this branch has been rebased, there might be some conflicts upon merge, as this branch now has a different commit for the config files than #1128)

hansegucker:
> This PR also contains the Webstorm config files that should be added in #1128. Is this on purpose?
> 
> (Since this branch has been rebased, there might be some conflicts upon merge, as this branch now has a different commit for the config files than #1128)

I will rebase this PR another time when #1128 got merged.

frontend/src/app/pages/exercises/exercise/shared/simulation/trainer-modal/radiogram-list/radiogram-card/radiogram-card-content-personnel-count/radiogram-card-content-personnel-count.component.html
```diff
@@ -4,9 +4,15 @@ <h5>Personalstatus</h5>
 
 <table *ngIf="radiogram$ | async as radiogram" class="table table-sm">
     <tbody>
-        <tr *ngFor="let personnelCategory of personnelCategories">
-            <td>{{ personnelCategory | personnelName }}</td>
-            <td>{{ radiogram.personnelCount[personnelCategory] }}</td>
+        <tr
+            *ngFor="
+                let personnelTemplate of personnelTemplates$ | async | values
+            "
```
lukasrad02, line 10: It might make sense to limit this to only those templates where count > 0, so the table does not get to big once there are more templates than the current default

frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/exercise-statistics-modal/exercise-statistics-modal.component.ts
```diff
@@ -136,7 +136,7 @@ export class ExerciseStatisticsModalComponent implements OnInit {
 
     // Personnel statistics
     private readonly personnelConfig: {
-        [key in PersonnelType]: { label: string; color: string };
+        [key in string]: { label: string; color: string };
```
lukasrad02: Is it possible to drop this fixed color assignment and let Chart.js choose colors for only those types that are present on-demand?

It is not necessary right know, but it might make sense to already start thinking about this…

lukasrad02 (commented):
As discussed in-person, I'd suggest to add a UUIID to the material and personnel types as unique identifier to avoid potential collisions once custom types can be defined.

I'll pause my review in the meantime. Here are already some small suggestions regarding the files I've reviewed so far

hansegucker:
@lukasrad02 I have now introduced ids for the templates, therefore I would appreciate a review of this PR. Heads up: Test scenarios are still missing and the migration might not cover all cases yet. I would like to fix this when we are sure that there won't be any further changes to the data model in this PR.

shared/src/state.ts
```diff
@@ -126,14 +125,16 @@ export class ExerciseState {
...
+    @IsArray()
+    @ValidateNested()
+    @Type(() => MaterialTemplate)
+    public readonly materialTemplates: readonly MaterialTemplate[] =
+        Object.values(defaultMaterialTemplates);
+    @IsArray()
+    @ValidateNested()
+    @Type(() => PersonnelTemplate)
+    public readonly personnelTemplates: readonly PersonnelTemplate[] =
+        Object.values(defaultPersonnelTemplates);
```
lukasrad02: Why aren't you using UUIDMaps here?

shared/src/models/utils/personnel-type.ts
```diff
@@ -1,21 +1,5 @@
...
-} = {
-    gf: 'Gruppenführer',
-    notarzt: 'Notarzt',
-    notSan: 'Notfallsanitäter',
-    rettSan: 'Rettungssanitäter',
-    san: 'Sanitäter',
-};
+export function getPersonnelTypes(state: ExerciseState) {
+    return Object.keys(state.personnelTemplates);
+}
```
lukasrad02: Identifying the personnel types by their array index while they also have a UUID feels really weird. But this should be fixed by using UUIDMaps in the state.

And in general: Please use `StrictObject` instead of `Object`. It provides better type-safety

shared/src/models/utils/rescue-resource.ts
```diff
@@ -32,13 +30,13 @@ export class PersonnelResource {
...
+    public readonly personnelCounts: ResourceDescription;
 
     /**
      * @deprecated Use {@link create} instead
      */
-    constructor(personnelCounts?: ResourceDescription<PersonnelType>) {
+    constructor(personnelCounts?: ResourceDescription) {
         this.personnelCounts = personnelCounts ?? {
             gf: 0,
             notarzt: 0,
```
lukasrad02: With dropping the fixed list of personnel types, I would prefer to also drop this default assignment.

I hope that nothing breaks by doing this, but I think TypeScript should catch every place where we access a specific personnelType without checking whether it is defined now that this ResourceDescription is string-based

shared/src/data/default-state/material-templates.ts
```diff
@@ -31,7 +30,7 @@ const bigMaterialTemplate = MaterialTemplate.create(
 );
 
 export const defaultMaterialTemplates: {
-    [key in MaterialType]: MaterialTemplate;
```
lukasrad02, line 34: I'd suggest to not annotate a type for those default templates objects and use `as const` for the assignment for both material and personnel templates, e.g.

```ts
export const defaultMaterialTemplates = {
    standard: standardMaterialTemplate,
    big: bigMaterialTemplate,
} as const;
```

This way, you don't have to define a type that we initially wanted to get rid of (literal union of material/personnel types), but still can use dot-notation to access the templates in the default object as long as this default object is still used.

shared/src/models/material.ts
```diff
@@ -25,6 +25,12 @@ export class Material {
     @IsValue('material' as const)
     public readonly type = 'material';
 
+    @IsUUID()
+    public readonly baseTemplateId: UUID;
```
lukasrad02: "base template" sounds a bit redundant.

But I like the idea of the backlink

shared/src/models/material.ts
```diff
@@ -25,6 +25,12 @@ export class Material {
     @IsValue('material' as const)
     public readonly type = 'material';
 
+    @IsUUID()
```
lukasrad02: This validator is lacking the args `(4, uuidValidationOptions)`

shared/src/models/personnel.ts
```diff
@@ -32,8 +26,17 @@ export class Personnel {
     @IsUUID(4, uuidValidationOptions)
     public readonly vehicleId: UUID;
 
-    @IsLiteralUnion(personnelTypeAllowedValues)
-    public readonly personnelType: PersonnelType;
+    @IsString()
+    public readonly personnelType: string;
+
+    @IsUUID()
+    public readonly baseTemplateId: UUID;
```
lukasrad02: Same thing as with the material template: Name is redundant and validator is lacking args

shared/src/models/personnel.ts
```diff
@@ -82,23 +85,29 @@ export class Personnel {
...
+        baseTemplateId: UUID,
+        name: string,
         assignedPatientIds: UUIDSet,
         image: ImageProperties,
         canCaterFor: CanCaterFor,
         treatmentRange: number,
         overrideTreatmentRange: number,
-        position: Position
+        position: Position,
+        abbreviation: string = ''
```
lukasrad02: I'd prefer if this parameter would appear directly after `name` to match the order the class members are defined in as well as the logical grouping

shared/src/models/radiogram/personnel-count-radiogram.ts
```diff
@@ -10,12 +10,11 @@ import { uuidValidationOptions } from '../../utils/index.js';
 import { IsValue } from '../../utils/validators/index.js';
 import { IsPersonnelCount } from '../../utils/validators/is-personnel-count.js';
 import { IsRadiogramStatus } from '../../utils/validators/is-radiogram-status.js';
-import type { PersonnelType } from '../utils/index.js';
 import { getCreate } from '../utils/get-create.js';
 import type { Radiogram } from './radiogram.js';
 import type { ExerciseRadiogramStatus } from './status/exercise-radiogram-status.js';
 
-export type PersonnelCount = { [key in PersonnelType]: number };
+export type PersonnelCount = { [key in string]: number };
```
lukasrad02: I think we can remove this type completely. This was always just a `ResourceDescription` (until now a `ResourceDescription<PersonnelType>`, now a plain `ResourceDescription`).

After changing this radiogram model, you can also remove `shared/src/utils/validators/is-personnel-count.ts`.

And the key doesn't have to stay a `personnelType`. We could either use the template id here (and perform a lookup of the name on use) or, since a radiogram contains user-facing information, directly use the display names of the personnel types here. Depending on how you marketplace-approach should look like, display names might be the better approach, since they will work even if the personnel template is no longer available for whatever reason. (We could still group the personnel by template id, but couldn't derive the display name from that id in the frontend).

No matter whether you refactor this to use IDs or display names, keep in mind to update the `radiogram-card-content-personnel-count` component and the log function for this radiogram (<https://github.com/hpi-sam/digital-fuesim-manv/blob/1a8ba5347f06d9341a4bf0af0e35a222ce2d0915/shared/src/store/action-reducers/utils/log.ts#L844>)

shared/src/models/personnel.ts
```diff
@@ -32,8 +26,17 @@ export class Personnel {
     @IsUUID(4, uuidValidationOptions)
     public readonly vehicleId: UUID;
 
-    @IsLiteralUnion(personnelTypeAllowedValues)
-    public readonly personnelType: PersonnelType;
+    @IsString()
+    public readonly personnelType: string;
```
lukasrad02, line 33:
```suggestion
    /**
     * @deprecated This will be refactored into a capability-based system. Please already consider using {@link baseTemplateId} if you only need an opaque identifier of the type and you don't assert any properties of the personnel.
     */
    @IsString()
    public readonly personnelType: string;
```

shared/src/models/material.ts
```diff
@@ -25,6 +25,12 @@ export class Material {
     @IsValue('material' as const)
     public readonly type = 'material';
 
+    @IsUUID(4, uuidValidationOptions)
+    public readonly baseTemplateId: UUID;
+
+    @IsString()
+    public readonly name: string;
```
lukasrad02: `name` is commonly used for user-defined, exercise-specific properties (like names of vehicles, viewports, etc).

Although material and personnel do not have such names (yet), I think it would make sense to differentiate, for example by naming this `typeName` and also adjusting the other properties, such as the personnel name and abbreviation.

shared/src/models/utils/tag-helpers.ts
```diff
@@ -284,13 +284,13 @@ export function createBehaviorTag(
...
+    personnel: Personnel | PersonnelTemplate
 ): Tag {
     return new Tag(
         'Personaltyp',
         'chocolate',
         'white',
-        personnelTypeNames[personnelType],
-        personnelType
+        personnel.name,
+        personnel.personnelType
```
lukasrad02: This `specifier` is just an internal identifier for aggregation and filtering. You can already use the template ID here:

```suggestion
        personnel.type === 'personnelTemplate'
            ? personnel.id
            : personnel.baseTemplateId
```

shared/src/models/vehicle.ts
```diff
@@ -23,6 +23,9 @@ export class Vehicle {
     @IsString()
     public readonly name: string;
 
+    @IsUUID(4, uuidValidationOptions)
+    public readonly baseTemplateId: UUID;
```
lukasrad02: Also rename this property (see comment on `Material` model)

shared/src/models/personnel.ts
```diff
@@ -32,8 +26,17 @@ export class Personnel {
...
+
+    @IsUUID(4, uuidValidationOptions)
+    public readonly baseTemplateId: UUID;
+
+    @IsString()
+    public readonly name: string;
+
+    @IsString()
+    public readonly abbreviation: string = '';
```
lukasrad02: Rename (see comment on `Material.name`)

shared/src/simulation/activities/provide-personnel-from-vehicles.ts
```diff
@@ -162,8 +160,8 @@ function personnelInVehicleTemplate(
         (tp) => tp.id === templateId
     )[0];
     if (template) {
-        template.personnel.forEach((pt) => {
-            resource[pt]++;
+        template.personnelTemplateIds.forEach((pt) => {
+            resource[pt]!++;
```
lukasrad02: Using the template IDs here does not match with the `personnelType` that is still used in the `requiredPersonnelCounts` of the activity state and thus breaks the activity.

shared/src/simulation/activities/provide-personnel-from-vehicles.ts
```diff
@@ -192,7 +190,7 @@ function personnelInUnloadingVehicles(
                 tryGetElement(draftState, 'personnel', personnelId)
                     ?.personnelType
         )
-        .filter((pt): pt is PersonnelType => pt !== undefined)
-        .forEach((pt) => resource[pt]++);
+        .filter((pt): pt is string => pt !== undefined)
```
lukasrad02: This type guard is unnecessary

```suggestion
        .filter((pt) => pt !== undefined)
```

shared/src/simulation/behaviors/assign-leader.ts
```diff
@@ -51,7 +50,7 @@ export class AssignLeaderBehaviorState implements SimulationBehaviorState {
     static readonly create = getCreate(this);
 }
 
-const personnelPriorities: { [Key in PersonnelType]: number } = {
+const personnelPriorities: { [Key in string]: number } = {
```
lukasrad02: Use `as const` instead of the type definition.

Alternatively, to ensure that the map contains all personnel types (which was the original intention of this typedef and would still work as long as personnel types are still in use), you could also use `[Key in keyof<typeof<defaultPersonnelTemplates>>]` (handwritten in GitHub web UI, might contain syntax errors or not work out of the box)

shared/src/simulation/behaviors/assign-leader.ts
```diff
@@ -198,17 +197,17 @@ export const assignLeaderBehavior: SimulationBehavior<AssignLeaderBehaviorState>
...
+                                    personnelCount['notarzt'] =
                                         groupedPersonnel['notarzt']?.length ??
                                         0;
-                                    personnelCount.rettSan =
+                                    personnelCount['rettSan'] =
                                         groupedPersonnel['rettSan']?.length ??
                                         0;
-                                    personnelCount.san =
+                                    personnelCount['san'] =
                                         groupedPersonnel['san']?.length ?? 0;
```
lukasrad02: Since the `PersonnelCountRadiogram` does not require all properties to be set with the new type definition, I think this can be simplified by using `.map()` and `StrictObject.fromEntries()`.

However, you would have to update the `radiogram-card-content-personnel-count` component, as this currently asserts that every personnel type is present in the `personnelCount`

shared/src/simulation/behaviors/assign-leader.ts
```diff
@@ -365,8 +364,8 @@ function selectNewLeader(
 
     personnel.sort(
         (a, b) =>
-            personnelPriorities[b.personnelType] -
-            personnelPriorities[a.personnelType]
+            (personnelPriorities[b.personnelType] ?? 0) -
+            (personnelPriorities[a.personnelType] ?? 0)
```
lukasrad02: Rather use `-1` here, since `0` is actually used as a priority

lukasrad02 (commented):
Partial review, so you can already start working on addressing my suggestions.

shared/src/data/default-state/personnel-templates.ts
```diff
@@ -1,73 +1,82 @@
...
         aspectRatio: 1,
     },
     CanCaterFor.create(0, 0, 0, 'and'),
     0,
-    0
+    0,
+    'GF'
 );
 
 export const defaultPersonnelTemplates: {
```
fekoch: same as https://github.com/hpi-sam/digital-fuesim-manv/pull/1124/files#r2533896864

frontend/src/app/pages/exercises/exercise/shared/editor-panel/map-editor-card/map-editor-card.component.html
```diff
@@ -18,6 +18,14 @@
         >
             <i class="bi bi-pencil-fill"></i>
         </button>
+        <button
+            *ngIf="enableDeleteButton"
+            (mousedown)="elementDelete.emit(); $event.stopPropagation()"
+            class="btn btn-outline-danger btn-sm position-absolute"
+            style="top: 0; right: 0"
```
lukasrad02: Currently, it would not be possible to enable both the edit and delete button, since ther're both placed at the same, absolute position. You could wrap both buttons in an absolutely positioned `div` instead, so that both buttons can be enabled at the same time without having to calculate coordinates

shared/src/state.ts
hansegucker: I did it consistent with `vehicleTemplates` which also is only an array.

shared/src/simulation/activities/reassign-treatments.ts
```diff
@@ -299,7 +296,7 @@ function createCateringMaterials(
     }));
 }
 
-const personnelPriorities: { [Key in PersonnelType]: number } = {
+const personnelPriorities: { [Key in string]: number } = {
```
lukasrad02: Use `as const` or `keyof<typeof<defaultPersonnelTemplates>>` (more information in one of my previous comments)

shared/src/simulation/activities/reassign-treatments.ts
```diff
@@ -380,7 +377,7 @@ function createCateringPersonnel(
 ): CateringPersonnel[] {
     return personnel.map((pers) => ({
         personnel: pers,
-        priority: personnelPriorities[pers.personnelType],
+        priority: personnelPriorities[pers.personnelType] ?? 0,
```
lukasrad02: Use `-1`, since `0` is an actually used priority:


```suggestion
        priority: personnelPriorities[pers.personnelType] ?? -1,
```

shared/src/store/action-reducers/exercise.ts
```diff
@@ -287,19 +287,18 @@ export function preparePartialExportForImport(
 }
 
 export interface TreatmentAssignment {
-    [patientId: UUID]: { [personnelType in PersonnelType]: number };
+    [patientId: UUID]: { [personnelType in string]: number };
```
lukasrad02: Use `ResourceDescription` instead of equivalent inline type (for the right hand side, i.e., `[patientId: UUID]: ResourceDescription`)

shared/src/models/utils/personnel-type.ts
lukasrad02: Also: Why is this not a proper selector?

shared/src/store/action-reducers/exercise.ts
```diff
@@ -311,7 +310,7 @@ function calculateTreatmentAssignment(
         StrictObject.keys(personnel.assignedPatientIds)
             .filter((patientId) => treatmentAssignment[patientId])
             .forEach((patientId) => {
-                treatmentAssignment[patientId]![personnel.personnelType] +=
+                treatmentAssignment[patientId]![personnel.personnelType]! +=
```
lukasrad02: Since `getPersonnelTypes` returns the array index, the object returned by this function will also use those indices instead of the string-based personnel types, breaking its consumers

shared/src/store/action-reducers/exercise.ts
```diff
@@ -341,15 +340,25 @@ function evaluateTreatmentReassignment(
                 StrictObject.entries(newTreatmentAssignment[patientId]!)
                     .filter(([, count]) => count > 0)
                     .map(([personnelType]) =>
-                        createPersonnelTypeTag(draftState, personnelType)
+                        createPersonnelTypeTag(
+                            draftState,
+                            draftState.personnelTemplates.find(
+                                (template) =>
+                                    template.personnelType === personnelType
```
lukasrad02: Related to the previous comment: This comparison will never be true, as you're comparing indices with type names here

shared/src/state-migrations/43-generalize-materials-personnel.ts
```diff
@@ -0,0 +1,206 @@
...
+} = {
+    gf: uuid(),
+    notarzt: uuid(),
+    notSan: uuid(),
+    rettSan: uuid(),
+    san: uuid(),
+};
+
+const personnelTypeAbbreviations: {
+    [Key in string]: string;
```
lukasrad02:
```suggestion
    [Key in PersonnelType]: string;
```

shared/src/state-migrations/43-generalize-materials-personnel.ts
```diff
@@ -0,0 +1,206 @@
...
+} = {
+    gf: 'GF',
+    notarzt: 'NA',
+    notSan: 'NFS',
+    rettSan: 'RS',
+    san: 'San',
+};
+
+export const personnelTypeNames: {
+    [key in string]: string;
```
lukasrad02:
```suggestion
    [key in PersonnelType]: string;
```

shared/src/state-migrations/43-generalize-materials-personnel.ts
```diff
@@ -0,0 +1,206 @@
...
+} = {
+    gf: 'Gruppenführer',
+    notarzt: 'Notarzt',
+    notSan: 'Notfallsanitäter',
+    rettSan: 'Rettungssanitäter',
+    san: 'Sanitäter',
+};
+export const generalizeMaterialsPersonnel43: Migration = {
```
lukasrad02:
```suggestion
};

export const generalizeMaterialsPersonnel43: Migration = {
```

shared/src/state-migrations/43-generalize-materials-personnel.ts
```diff
@@ -0,0 +1,206 @@
...
+                (vehicleParameters) =>
+                    vehicleParameters.personnel.forEach((personnel) => {
+                        personnel.name =
+                            personnelTypeNames[personnel.personnelType];
+                        personnel.baseTemplateId =
+                            personnelTemplateIds[personnel.personnelType];
+                        personnel.abbreviation =
+                            personnelTypeAbbreviations[personnel.personnelType];
+                    })
+            );
```
lukasrad02, line 198: This code is largely identical to the code for the add vehicle action. Could you extract common logic into a function? (For example, extracting the bodies of the inner `.forEach()` loops into a `migrateMaterial` (or `migratePersonnel`) function)

Those functions could also be used in the state migrations below

shared/src/state-migrations/44-generalize-materials-personnel.ts
```diff
@@ -0,0 +1,206 @@
...
+                    vehicleParameters.personnel.forEach((personnel) => {
+                        personnel.name =
+                            personnelTypeNames[personnel.personnelType];
+                        personnel.baseTemplateId =
+                            personnelTemplateIds[personnel.personnelType];
+                        personnel.abbreviation =
+                            personnelTypeAbbreviations[personnel.personnelType];
+                    })
+            );
+        } else if (actionType === '[VehicleTemplate] Edit vehicleTemplate') {
```
lukasrad02, line 204: What about the add vehicle template action?

shared/src/state-migrations/43-generalize-materials-personnel.ts
```diff
@@ -0,0 +1,206 @@
+import type { UUID } from '../utils/index.js';
+import type { ImageProperties } from '../models/index.js';
```
lukasrad02: Please define this type within the migration instead relying on external types (that could change in newer state versions)

frontend/src/app/pages/exercises/exercise/shared/simulation/trainer-modal/radiogram-list/radiogram-card/radiogram-card-content-personnel-count/radiogram-card-content-personnel-count.component.html
```diff
@@ -4,9 +4,11 @@ <h5>Personalstatus</h5>
 
 <table *ngIf="radiogram$ | async as radiogram" class="table table-sm">
     <tbody>
-        <tr *ngFor="let personnelCategory of personnelCategories">
-            <td>{{ personnelCategory | personnelName }}</td>
-            <td>{{ radiogram.personnelCount[personnelCategory] }}</td>
+        <tr *ngFor="let personnelTemplate of personnelTemplates$ | async">
+            <td>{{ personnelTemplate.name }}</td>
+            <td>
+                {{ radiogram.personnelCount[personnelTemplate.id] }}
```
lukasrad02: This is (currently) not correct, since the radiogram uses personnel type names instead of template IDs.

However, since I already suggested to refactor the radiogram, this can stay as-is. I've just added the comment in case you decide to not refactor the radiogram.

frontend/src/app/pages/exercises/exercise/shared/simulation/trainer-modal/tabs/general-tab/simulated-region-overview-general-tab.component.ts
```diff
@@ -123,20 +112,62 @@ export class SimulatedRegionOverviewGeneralTabComponent implements OnInit {
...
-            this.personnel[`${category}$`] = this.store.select(
-                createSelectByPredicate(
-                    containedPersonnelSelector,
-                    this.createPersonnelTypePredicate(category)
-                )
-            );
-        });
+        this.personnel$ = this.store.select(
+            createSelector(
+                selectPersonnelTemplates,
```
lukasrad02: The templates aren't used in this selector

frontend/src/app/pages/exercises/exercise/shared/simulation/trainer-modal/tabs/general-tab/simulated-region-overview-general-tab.component.ts
```diff
@@ -123,20 +112,62 @@ export class SimulatedRegionOverviewGeneralTabComponent implements OnInit {
...
+                    });
+
+                    return categorizedPersonnel;
+                }
+            )
+        );
+
+        this.materials$ = this.store.select(
+            createSelector(
+                selectMaterialTemplates,
```
lukasrad02: The templates aren't used in this selector

shared/src/models/utils/personnel-type.ts
```diff
@@ -1,21 +1,5 @@
-import type { AllowedValues } from '../../utils/validators/index.js';
+import type { ExerciseState } from '../../state.js';
 
-export type PersonnelType = 'gf' | 'notarzt' | 'notSan' | 'rettSan' | 'san';
```
lukasrad02, line 3: This type is still used in `shared/src/simulation/activities/reassign-treatments.spec.ts`

lukasrad02 (changes requested):
I've now reviewed all changes and did some basic testing. Thorough testing will follow after you've addressed my suggestions so I don't have to test multiple times.

frontend/src/app/pages/exercises/exercise/shared/simulation/trainer-modal/radiogram-list/radiogram-card/radiogram-card-content-personnel-count/radiogram-card-content-personnel-count.component.html
lukasrad02: Why did you resolve this comment? As far as I can tell, this is not yet addressed.

(I unresolved the comment now so that this comment does not get overlooked)

shared/src/state.ts
lukasrad02: Ah, I see. But currently, the rest of your code is not that consistent. For example, `getPersonnelTypes` is implemented as if the templates were stored in an ID map (see <https://github.com/hpi-sam/digital-fuesim-manv/pull/1124#discussion_r2533799906>).

My idea why this should be an ID map is that we now have to lookup personnel (and sometimes, but way less frequently) templates by ID in the state, for example, if we want to get the personnel type name from a template id in a `ResourceDescription`. Currently, `state.personnelTemplates.find(pt => pt.id === idWeAreSearchingFor)` is used multiple times. Of course, this could just be factored out into a helper function, but with frequent lookups, maps seem more reasonable to me. If you strive for consistency, I'd rather also store the vehicle templates in a map than personnel and material in an array.

hansegucker: I agree with you and will migrate the vehicle templates to a map.

frontend/src/app/pages/exercises/exercise/shared/simulation/trainer-modal/radiogram-list/radiogram-card/radiogram-card-content-personnel-count/radiogram-card-content-personnel-count.component.html
hansegucker: Ah, I thought it was simulated region and not radiogram content

lukasrad02:
@hansegucker `shared` does not build on this branch. Could you please fix this so I can review a working state?

hansegucker:
> @hansegucker `shared` does not build on this branch. Could you please fix this so I can review a working state?

Sorry, I am not sure why the issue didn't occur to me. Should be fixed now.

shared/src/simulation/behaviors/assign-leader.ts
lukasrad02: This comment is not addressed yet

shared/src/simulation/activities/reassign-treatments.ts
lukasrad02: Unaddressed comment

(In general: If you disagree with my suggestions, feel free to add a comment. I just want to point out parts of my review that you might have missed by accident)

shared/src/models/utils/personnel-type.ts
lukasrad02: This is not fixed yet

shared/src/simulation/activities/provide-personnel-from-vehicles.ts
lukasrad02: There is still a mismatch somewhere here. `shared/src/simulation/activities/reassign-treatments.ts` operates completely on type names and also uses them to send the `ResourceRequiredEvent`. But the data is then interpreted as template IDs.

Perhaps it helps to add doc comments to properties like `ResourceRequiredEvent.requiredResource`, `ProvidePersonnelFromVehiclesActivityState.requiredPersonnelCounts`, etc to indicate whether they are based on type names or template IDs  
Edit: Even differentiating between `ResourceDescription<string>` (for type names) and `ResourceDescription<UUID>` (for template IDs) might help to spot some issues. This won't trigger any errors at compile time, since `string === UUID`, but at least provides a hint on what values to expect

frontend/src/app/pages/exercises/exercise/shared/editor-panel/edit-vehicle-template-modal/edit-vehicle-template-modal.component.ts
```diff
@@ -41,6 +48,20 @@ export class EditVehicleTemplateModalComponent implements OnInit {
...
+                this.vehicleTemplate!.materialTemplateIds.includes(template.id)
+            )
+        );
+        this.personnelTemplates = cloneDeepMutable(
+            Object.values(
+                selectStateSnapshot(selectPersonnelTemplates, this.store)
+            ).filter((template) =>
+                this.vehicleTemplate!.personnelTemplateIds.includes(template.id)
+            )
+        );
```
lukasrad02: This does not take the count of the personnel/material into account. If I create a vehicle with three instances of big material, save the template and open the editor again, it only shows one instance. Immediately saving the template again (without performing any changes, which should be a no-op) then actually writes the wrong number of big material into the template so that subsequently added vehicles only have a single big material.

Same issue for the personnel.

You can easily reproduce the issue by editing the "GW San" template that consists of 2 RS, 2 San, 4 big material (and one GF and NA each). The editor does not show those counts.

frontend/src/app/pages/exercises/exercise/shared/editor-panel/material-template-display/material-template-display.component.ts
lukasrad02, line 1: This file should be deleted (not just empty)

frontend/src/app/pages/exercises/exercise/shared/editor-panel/templates/simulated-region.ts
```diff
@@ -144,9 +144,9 @@ function reconstituteBehavior(
     behavior.id = uuid();
     switch (behavior.type) {
         case 'providePersonnelBehavior':
-            behavior.vehicleTemplatePriorities = state.vehicleTemplates.map(
-                (template) => template.id
-            );
+            behavior.vehicleTemplatePriorities = Object.values(
```
lukasrad02: Use `StrictObject`

lukasrad02 (changes requested):
Not a full review yet, but should be enough that you can start fixing the issues already pointed out.

(I've also un-resolved some of my previous comments where I had the impression that they're not yet addressed.)

shared/src/simulation/activities/reassign-treatments.ts
hansegucker: I actually implemented your suggestion which was the reason why I didn't answer on this comment. Unfortunately, this doesn't work quite well type-wise, as accessing arbitrary strings on a const-typed object is somehow forbidden. (And so I undid the change.)

frontend/src/app/pages/exercises/exercise/shared/editor-panel/edit-vehicle-template-modal/edit-vehicle-template-modal.component.ts
hansegucker: Ups

shared/src/simulation/activities/provide-personnel-from-vehicles.ts
hansegucker: For now, I decided to continue with personnel types as there will be some further changes here when adapting a capability system.

shared/src/state-migrations/43-generalize-materials-personnel.ts
```diff
@@ -0,0 +1,302 @@
...
+    );
+    delete vehicleTemplate.personnel;
+    return vehicleTemplate;
+}
+
+function migrateMapImage(
+    mapImage: MapImage,
+    mapImageTemplateIds: { [key in string]: UUID }
+) {
+    mapImage.id = mapImageTemplateIds[mapImage.image.url];
```
lukasrad02: I suppose this should be

```suggestion
    mapImage.templateId = mapImageTemplateIds[mapImage.image.url];
```

Afterwards, the migration test for `state-23 › combined-scenarios › everything-dropped-on-the-map-once-complete.json` should succeed. (You also have to change `id` to `templateId` in the `MapImage` type definition.)

lukasrad02 (commented):
I've investigated the failing migration tests and found the following issues:

- from version 23: Should be fixed by the suggestion below
- from versions 28 and 37 (1+4 failing scenarios): Not your fault, I've created #1178 for the bug
- from version 40: Not your fault, should be fixed by <https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios/pull/21>

shared/src/simulation/activities/reassign-treatments.ts
lukasrad02: Fair point. Since we don't have a `PersonnelType` any longer that we can rely on, a `const` object does not help us here.

Hence, you can keep it as-is.

shared/src/simulation/activities/provide-personnel-from-vehicles.ts
```diff
@@ -153,17 +151,18 @@ function personnelInVehicleTemplate(
     templateId: UUID
 ): {
     vehicleType: string | undefined;
-    vehiclePersonnel: ResourceDescription<PersonnelType>;
+    vehiclePersonnel: ResourceDescription;
 } {
     const resource = cloneDeepMutable(
         PersonnelResource.create()
     ).personnelCounts;
```
lukasrad02: This resource description is not 0-initialized and thus the `++` leads to `NaN` values that break the activity.

shared/src/simulation/activities/provide-personnel-from-vehicles.ts
```diff
@@ -172,7 +171,7 @@ function personnelInVehicleTemplate(
 function personnelInUnloadingVehicles(
     draftState: ExerciseState,
     simulatedRegion: SimulatedRegion
-): ResourceDescription<PersonnelType> {
+): ResourceDescription {
     const resource = cloneDeepMutable(
         PersonnelResource.create()
     ).personnelCounts;
```
lukasrad02: Same issue with non-initialized data as before.

shared/src/store/action-reducers/vehicle-templates.ts
```diff
@@ -46,13 +44,13 @@ export class EditVehicleTemplateAction implements Action {
     @Type(() => ImageProperties)
     public readonly image!: ImageProperties;
 
-    @IsString({ each: true })
+    @IsUUID(4, { each: true })
```
lukasrad02: Please use `uuidArrayValidationOptions` (combines the default `uuidValidationOptions` with `{ each: true }`)

shared/src/store/action-reducers/vehicle-templates.ts
```diff
@@ -46,13 +44,13 @@ export class EditVehicleTemplateAction implements Action {
...
     @IsArray()
-    public readonly materials!: readonly MaterialType[];
+    public readonly materialTemplateIds!: readonly UUID[];
 
-    @IsString({ each: true })
+    @IsUUID(4, { each: true })
```
lukasrad02: `uuidArrayValidationOptions`, see before

lukasrad02 (commented):
Some comments from my WIP review for you to continue working

shared/src/state-helpers/create-vehicle-parameters.ts
```diff
@@ -18,34 +17,39 @@ import type { UUID } from '../utils/index.js';
...
-    );
-    const personnel = vehicleTemplate.personnel.map((currentPersonnel) =>
-        Personnel.generatePersonnel(
-            personnelTemplates[currentPersonnel],
-            vehicleId,
-            vehicleTemplate.name,
-            VehiclePosition.create(vehicleId)
-        )
-    );
+    const materials: Material[] = vehicleTemplate.materialTemplateIds
```
lukasrad02: Unnecessary type annotation

shared/src/models/personnel.ts
```diff
@@ -101,6 +106,13 @@ export class Personnel {
         this.position = position;
     }
 
+    /**
+     * @deprecated Do not access directly, use helper methods from models/utils/position/position-helpers(-mutable) instead.
+     */
+    @IsPosition()
+    @ValidateNested()
+    public readonly position: Position;
```
lukasrad02: Why did you move this below the constructor?

shared/src/export-import/file-format/partial-export.ts
```diff
@@ -15,7 +15,6 @@ export class PartialExport extends BaseExportImportFile {
     @Type(() => PatientCategory)
     public readonly patientCategories?: PatientCategory[];
 
-    @IsOptional()
```
lukasrad02, line 18: Why did you remove this?

lukasrad02 (dismissed):
Not much to go, I now have reviewed everything except your changes to migration 38 (you pushed the changes in just the moment I finished my review) and the files changed in `shared/src/simulation/activities`.

Regarding the activities: I've noticed that the server hangs in exercises with simulation, but I didn't have the time to investigate. If you want to reproduce:

1. Create a new exercise
2. Add a simulated patient tray
3. Add 4x red patient and 1x RTW to the patient tray
4. Unmount the personnel from the RTW (just for convenience so you don't have to wait 2 minutes for the auto-unmount)
5. Adjust the settings of the treat patients behavior and set all intervals to 1 second (just for convenience so the bug triggers faster)
6. Adjust the settings of the request vehicles behavior and set the request interval to 0.5 seconds (just for convenience so the bug triggers faster)
7. Start the exercise and let it run until 1:07

You will notice that the timer stops, but the exercise still claims to be running. In case of an exception, the exercise would be paused and the buttons therefore say "Start", but they still say "Pause". Hence, I suspect that the server runs into some endless loop while calculating the simulation.

backend/src/exercise/patient-ticking.ts
```diff
@@ -13,7 +12,7 @@ import {
 /**
  * The count of assigned personnel and material that cater for a {@link Patient}.
  */
-type Catering = { [key in PersonnelType | 'material']: number };
+type Catering = { [key in string | 'material']: number };
```
fekoch: can just be key in string

backend/src/exercise/patient-ticking.ts
```diff
@@ -97,10 +99,10 @@ function getNextPatientHealthPoints(
...
 ): HealthPoints {
-    let material = treatedBy.material;
-    const notarzt = treatedBy.notarzt;
-    const notSan = treatedBy.notSan;
-    const rettSan = treatedBy.rettSan;
+    let material = treatedBy['material'] ?? 0;
+    const notarzt = treatedBy['notarzt'] ?? 0;
+    const notSan = treatedBy['notSan'] ?? 0;
+    const rettSan = treatedBy['rettSan'] ?? 0;
     // TODO: Sans should be able to treat patients too.
```
fekoch: maybe add a //TODO that this will have to be revisited when implementing capabilities

frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/exercise-statistics-modal/exercise-statistics-modal.component.ts
```diff
@@ -91,94 +83,64 @@ export class ExerciseStatisticsModalComponent implements OnInit {
...
+                const vehicleTemplateIds = new Set<UUID>();
                 for (const statistic of statistics) {
-                    for (const vehicleType of Object.keys(
+                    for (const templateId of Object.keys(
                         statistic.value.vehicles
                     )) {
-                        vehicleTypes.add(vehicleType);
+                        vehicleTemplateIds.add(templateId);
                     }
                 }
```
fekoch:
```suggestion
                const vehicleTemplateIds = new Set<UUID>(
                    statistics.flatMap((statistic) =>
                        Object.keys(statistic.value.vehicles)
                    )
                );
```

frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/exercise-statistics-modal/exercise-statistics-modal.component.ts
```diff
@@ -91,94 +83,64 @@ export class ExerciseStatisticsModalComponent implements OnInit {
...
 
                 return {
-                    datasets: [...vehicleTypes].map((vehicleType, index) => ({
-                        label: vehicleType,
+                    datasets: [...vehicleTemplateIds].map((templateId) => ({
+                        label: vehicleTemplates[templateId]?.vehicleType ?? '',
```
fekoch: I would like a more explizit/visible default value

frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/exercise-statistics-modal/exercise-statistics-modal.component.ts
```diff
@@ -91,94 +83,64 @@ export class ExerciseStatisticsModalComponent implements OnInit {
...
+                // Get all vehicle template IDs
+                const personnelTemplateIds = new Set<UUID>();
+                for (const statistic of statistics) {
+                    for (const templateId of Object.keys(
+                        statistic.value.personnel
+                    )) {
+                        personnelTemplateIds.add(templateId);
+                    }
+                }
+
```
fekoch: also use flatmap, remove the comment

frontend/src/app/pages/exercises/exercise/shared/exercise-statistics/exercise-statistics-modal/exercise-statistics-modal.component.ts
```diff
@@ -91,94 +83,64 @@ export class ExerciseStatisticsModalComponent implements OnInit {
...
+                return {
+                    datasets: [...personnelTemplateIds].map((templateId) => ({
+                        label: personnelTemplates[templateId]?.name ?? '',
```
fekoch: see above

frontend/src/app/state/application/selectors/exercise.selectors.ts
```diff
@@ -95,36 +95,24 @@ export const createSelectSimulatedRegion = createSelectElementFromMapFactory(
...
-
-// Element from Array
-export const createSelectMapImageTemplate = createSelectElementFromArrayFactory(
+export const createSelectMapImageTemplate = createSelectElementFromMapFactory(
     selectMapImagesTemplates
 );
-export const createSelectVehicleTemplate = createSelectElementFromArrayFactory(
-    selectVehicleTemplates
-);
+export function createSelectRadiogram<R extends ExerciseRadiogram>(id: UUID) {
```
fekoch, line 110: `radiograms[id]` can also be undefined. I think the selector functions should not mask this

shared/src/data/default-state/vehicle-templates.ts
```diff
@@ -84,35 +90,47 @@ const ktwKatSchutzVehicleTemplate = VehicleTemplate.create(
...
     `RTH ???`,
     rthImage,
     1,
-    ['notarzt', 'notSan'],
-    ['standard']
+    [defaultPersonnelTemplates.notarzt.id, defaultPersonnelTemplates.notSan.id],
+    [defaultMaterialTemplates.standard.id]
 );
 
 export const defaultVehicleTemplates: readonly VehicleTemplate[] = [
```
fekoch, line 136:
```suggestion
const defaultVehicleTemplates: readonly VehicleTemplate[] = [
```

`defaultVehicleTemplates` is only used locally

shared/src/export-import/file-format/partial-export.ts
```diff
@@ -15,7 +15,6 @@ export class PartialExport extends BaseExportImportFile {
...
     @IsArray()
```
fekoch, line 19: Is it possible to do `vehicleTemplates?` -> `vehicleTemplates` if we are also removing @IsOptional?

shared/src/simulation/activities/reassign-treatments.ts
```diff
@@ -274,20 +271,20 @@ interface CateringPersonnel {
...
             [sk in TreatablePatientStatus]?: number;
         };
     };
 };
 
 interface PersonnelSubstitution {
-    from: ResourceDescription<PersonnelType>;
-    to: Mutable<PersonnelType>;
+    from: ResourceDescription;
+    to: Mutable<string>;
```
fekoch:
```suggestion
    to: string;
```

shared/src/simulation/activities/reassign-treatments.ts
```diff
@@ -841,14 +842,14 @@ function recordPatientTreatedByPersonnel(
...
-): ResourceDescription<PersonnelType> {
+): ResourceDescription {
     const groupedPersonnel = groupBy(personnel, (p) => p.personnelType);
     const havePersonnel = Object.fromEntries(
         StrictObject.keys(personnelPriorities).map((pt) => [
             pt,
             groupedPersonnel[pt]?.length ?? 0,
         ])
-    ) as ResourceDescription<PersonnelType>;
+    ) as ResourceDescription;
```
fekoch: This type assertion is not needed, you can just do `havePersonnel: ResourceDescription = ...`

shared/src/store/action-reducers/map-image-templates.ts
```diff
@@ -46,18 +46,13 @@ export namespace MapImageTemplatesActionReducers {
...
                     throw new ReducerError(
                         `MapImageTemplate with id ${mapImageTemplate.id} already exists`
                     );
                 }
-                draftState.mapImageTemplates.push(
-                    cloneDeepMutable(mapImageTemplate)
-                );
+                draftState.mapImageTemplates[mapImageTemplate.id] =
+                    cloneDeepMutable(mapImageTemplate);
                 return draftState;
```
fekoch, line 56: is there a reason for cloning the mapImageTemplate?

fekoch (changes requested):

backend/src/exercise/patient-ticking.ts
lukasrad02: I also thought about suggesting this change, but I think that it actually does make sense to explicitly specify `material` to express that "the key is either some string (a personnel type) or literal `'material'`"

frontend/src/app/state/application/selectors/exercise.selectors.ts
lukasrad02: We have this issue with (almost) all selectors, see #1095. Since this specific selector wasn't added in this PR (the line is shown as addition, but only due to reordering of the selectors), I don't think that changing this selector is in scope for this already quite large PR.

backend/src/exercise/patient-ticking.ts
fekoch: Then we should do it properly and require it, I think:

```suggestion
type Catering = { [key in string]: number } & { material: number };
```

shared/src/export-import/file-format/partial-export.ts
hansegucker: It's not removable.

shared/src/store/action-reducers/map-image-templates.ts
hansegucker: I have no idea, but it's nothing I introduced so I will resolve this comment.

shared/src/models/personnel.ts
hansegucker: I am not sure, undid the change

shared/src/data/default-state/vehicle-templates.ts
hansegucker: But it could be used externally and is allowed to be, so I will keep it this way.

hansegucker:
> Not much to go, I now have reviewed everything except your changes to migration 38 (you pushed the changes in just the moment I finished my review) and the files changed in `shared/src/simulation/activities`.
> 
> Regarding the activities: I've noticed that the server hangs in exercises with simulation, but I didn't have the time to investigate. If you want to reproduce:
> 
>     1. Create a new exercise
> 
>     2. Add a simulated patient tray
> 
>     3. Add 4x red patient and 1x RTW to the patient tray
> 
>     4. Unmount the personnel from the RTW (just for convenience so you don't have to wait 2 minutes for the auto-unmount)
> 
>     5. Adjust the settings of the treat patients behavior and set all intervals to 1 second (just for convenience so the bug triggers faster)
> 
>     6. Adjust the settings of the request vehicles behavior and set the request interval to 0.5 seconds (just for convenience so the bug triggers faster)
> 
>     7. Start the exercise and let it run until 1:07
> 
> 
> You will notice that the timer stops, but the exercise still claims to be running. In case of an exception, the exercise would be paused and the buttons therefore say "Start", but they still say "Pause". Hence, I suspect that the server runs into some endless loop while calculating the simulation.

I can reproduce the exact same behavior, but I didn't found the issue.

lukasrad02:
I've found the issue that causes the exercise to freeze: `addResourceDescription` (as well as most of the other helpers) do not handle missing properties correctly:

```js
addResourceDescription({a:1,b:2}, {})
// → {a: NaN, b: NaN}
addResourceDescription({}, {a:1, b:2})
// → {}
addResourceDescription({a:1}, {a:2})
// → {a: 3}
```

This causes the loop <https://github.com/hpi-sam/digital-fuesim-manv/blob/87286086ac0e3320f0f3e33835e4813a3e80c3c9/shared/src/simulation/activities/provide-personnel-from-vehicles.ts#L97-L101> as soon as `NaN` slips in

lukasrad02:
I just noticed that `addPartialResourceDescriptions` and `subtractPartialResourceDescriptions` already existed. Now that the default functions also support partial descriptions, I think you can remove them

shared/src/state-migrations/38-deterministic-alarm-groups.ts
```diff
@@ -38,13 +39,24 @@ interface CanCaterFor {
...
+    type: 'materialTemplate';
+    materialType: MaterialType;
+    canCaterFor: CanCaterFor;
+    overrideTreatmentRange: number;
+    treatmentRange: number;
+    image: ImageProperties;
+}
+
 interface MaterialTemplate {
+    id?: UUID;
```
lukasrad02: This property seems unused.

shared/src/state-migrations/38-deterministic-alarm-groups.ts
```diff
@@ -38,13 +39,24 @@ interface CanCaterFor {
...
 
+interface MaterialTemplateIntermediary {
+    id?: UUID;
+    type: 'materialTemplate';
+    materialType: MaterialType;
+    canCaterFor: CanCaterFor;
+    overrideTreatmentRange: number;
+    treatmentRange: number;
+    image: ImageProperties;
+}
```
lukasrad02: I wonder why you added this additional interface. With the (unused) `id?: UUID` added to `MaterialTemplate`, those two interfaces are identical.

If you want to make this distinction to highlight that this migration deals with two potentially different variants of the templates (from state 37 and from the current state), you could still derive one type from the other one (e.g. `type MaterialTemplateIntermediary = MaterialTemplate & { id?: UUID }`). However, this intention is also not obvious to my from the name "Intermediary". Wouldn't it rather be "Current"?

While writing this comment, I thought that we could even change the typedef <https://github.com/hpi-sam/digital-fuesim-manv/blob/6227eb495edda5e6157c5e95eeff7d60308aec7e/shared/src/state-migrations/migration-functions.ts#L52-L55> so that the `intermediaryState` parameter is typed as `ExerciseState` (i.e., the current version from `state.ts`). This is correct according to how the migrations currently behave and would help us catch errors like the one in this PR even more reliably than by the migration tests.

shared/src/state-migrations/38-deterministic-alarm-groups.ts
```diff
@@ -57,18 +69,28 @@ interface Material {
...
 
+interface PersonnelTemplateIntermediary {
+    id?: UUID;
+    type: 'personnelTemplate';
+    personnelType: PersonnelType;
+    canCaterFor: CanCaterFor;
+    overrideTreatmentRange: number;
+    treatmentRange: number;
+    image: ImageProperties;
+}
```
lukasrad02: Same question as regarding `MaterialTemplateIntermediary`

hansegucker:
> I just noticed that `addPartialResourceDescriptions` and `subtractPartialResourceDescriptions` already existed. Now that the default functions also support partial descriptions, I think you can remove them

I think I would like to not further change things here, as at least the `subtract...` does a little bit more magic (filtering out values <= 0), so I don't want to risk breaking more things again as we have to finish this PR as soon as possible.

fekoch (approved):

PR #1076, opened 2025-05-20
Feature/interface signaller UI

lukasrad02:
This PR adds a new modal that allows interface signallers to interact with the software

### PR Checklist

Please make sure to fulfil the following conditions before marking this PR ready for review:

- [x] If this PR adds or changes features or fixes bugs, this has been added to the changelog
- [x] If this PR adds new actions or other ways to alter the state, [test scenarios](https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios) have been added.
- [x] I have the right to license the code submitted with this Pull Request under the mentioned license in the file [LICENSE-README.md](LICENSE-README.md) (i.e., this is my
      own code or code licensed under a license compatible to AGPL v3.0 or later, for exceptions look into [LICENSE-README.md](LICENSE-README.md)) and
      hereby license the code in this Pull Request under it.
      I certify that by signing off my commits (see In case of using third party code, I have given appropriate credit.
      We are using DCO for that, see [here](https://github.com/dcoapp/app#how-it-works) for more information.
- [x] If I have used third party code and I mentioned it in the code, I also updated the [inspired-by-or-copied-from-list.html](inspired-by-or-copied-from-list.html) list to include the links.

PR #1125, opened 2025-10-27
Implement EOC Viewport + Private EOC Logs

Quixelation:
This PR implements:
- the ability for trainers to assign participants to the Emergency Operations Center via a "Virtual"/"Pseudo"-Viewport in the participant-Overview.
    - Assigned Participants will see the old interface of the Emergency Operations Center instead of a map
- the distinction between private and public messages in the eoc-log
    - participants can only read and write public messages
    - trainers can read and write public and private messages

Test Scenarios have been added in https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios/pull/19

---

### PR Checklist

- [X] If this PR adds or changes features or fixes bugs, this has been added to the changelog
- [X] If this PR adds new actions or other ways to alter the state, [test scenarios](https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios) have been added.
- [X] I have the right to license the code submitted with this Pull Request under the mentioned license in the file [LICENSE-README.md](LICENSE-README.md) (i.e., this is my
- [X] If I have used third party code and I mentioned it in the code, I also updated the [inspired-by-or-copied-from-list.html](inspired-by-or-copied-from-list.html) list to include the links.


frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/emergency-operations-center-full/emergency-operations-center-full.component.html
```diff
@@ -0,0 +1,10 @@
+<div class="modal-body" style="min-height: 60vh">
+    <app-eoc-alarm-group-triggers />
+    <br />
+    <div class="card">
+        <h5 class="card-header">Einsatztagebuch</h5>
+        <div class="card-body">
+            <app-eoc-log-interface />
+        </div>
+    </div>
```
lukasrad02: I'd prefer if the subcomponents would use the same nesting depth, i.e., both contain only the card body (like `app-eoc-log-interface`) or contain the full card (like `app-eoc-alarm-group-triggers`)

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-alarm-group-triggers/eoc-alarm-group-triggers.component.html
lukasrad02, line 1: Regarding bringing back the old alarm group interface (prior to #1076): @christianschaeffer you mentioned that some users had issues with the new UI. I wonder whether it would make sense to also revert the UI change in the trainer view.

In any case, I noticed that the hotkeys of the new UI can't be used in the trainer view, as the prevent entering log messages that contain the hotkey characters ("a" and "z"), and the hotkeys have mainly been added for the IFS.

I removed the old UI so we don't have to maintain two different user interfaces for the same purpose, but if we bring back the old one anyways, we could also use it in the trainer view.

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-alarm-group-triggers/eoc-alarm-group-triggers.component.html
```diff
@@ -0,0 +1,82 @@
...
+    <h5 class="card-header">Alarmieren</h5>
+
+    <div *ngIf="alarmGroups$ | async | values as alarmGroups" class="card-body">
+        <div
+            *appLet="
+                transferPoints$
+                    | async
+                    | values
+                    | orderBy: getTransferPointOrderByValue as transferPoints
+            "
```
lukasrad02: This variable does not seem to be used

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-alarm-group-triggers/eoc-alarm-group-triggers.component.ts
lukasrad02, line 1: I think it would make sense to deduplicate with `frontend/src/app/shared/components/send-alarm-group-interface/send-alarm-group-interface.component.ts` (the meanwhile new UI), especially more complex parts like `function sendAlarmGroup`.

Either extract some common helpers or combine both components into a single one with an input to toggle the two HTML templates.

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-log-interface/eoc-log-interface.component.html
```diff
@@ -25,23 +36,44 @@
...
+                        </span>
+                    </div>
 
-                <div class="col" style="white-space: pre-wrap">
-                    {{ eocLogEntry.message }}
+                    <div
+                        class="col"
+                        style="white-space: pre-wrap"
+                        [style.color]="
+                            eocLogEntry.isPrivate ? '#131350' : 'unset'
```
lukasrad02: I can hardly see the difference of `#131350` and the default color. Could you please increase the contrast?

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-log-interface/eoc-log-interface.component.ts
```diff
@@ -21,6 +21,12 @@ export class EocLogInterfaceComponent {
 
     public newLogEntry = '';
 
+    public sendingPrivateLog = true;
+    public readonly clientisTrainer$ = this.store.select(selectOwnClient).pipe(
+        first(),
```
lukasrad02: This seems like a use case for `selectStateSnapshot`

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/transfer-target-dropdown/transfer-target-dropdown.component.html
```diff
@@ -0,0 +1,45 @@
...
+        class="btn btn-outline-primary btn-sm"
+        data-cy="sendAlarmGroupChooseTargetButton"
+        [disabled]="transferPoints.length === 0"
+    >
+        <app-transfer-point-name
+            *ngIf="selectedTransferPointId; else noTargetTransferPoint"
+            [transferPointId]="selectedTransferPointId"
+        />
+        <ng-template #noTargetTransferPoint>
+            <span class="text-muted"> Kein Ziel ausgewählt </span>
```
lukasrad02:
```suggestion
            <span class="text-muted">Kein Ziel ausgewählt</span>
```

frontend/src/app/pages/exercises/exercise/shared/exercise-map/exercise-map.component.html
```diff
@@ -1,6 +1,11 @@
+<app-emergency-operations-center-full
+    *ngIf="(restrictedToViewport$ | async)?.viewportType === 'eoc'"
+/>
+
 <!-- We want keyboard events to go to the OpenLayers map, not the container -> tabindex="-1" -->
 <div
     #openLayersContainer
+    [hidden]="(restrictedToViewport$ | async)?.viewportType === 'eoc'"
```
lukasrad02: Is there a reason why this element is only hidden, but not completely removed from the DOM (i.e., `*ngIf`)? The map does not need that much performance, but we could still save some computations if the map is not in the DOM at all

frontend/src/app/pages/exercises/exercise/shared/exercise-map/utility/ol-map-manager.ts
```diff
@@ -137,6 +137,11 @@ export class OlMapManager {
             .select(selectRestrictedViewport)
             .pipe(takeUntil(this.destroy$))
             .subscribe((viewport) => {
+                if (viewport?.viewportType === 'eoc') {
+                    // Do not restrict the view for EOCs, since it's not a map (no size)
+                    return;
+                }
```
lukasrad02: Wouldn't it make sense to combine this with the `if (!viewport)` a couple of lines later?

Or, if the map is not in the DOM (see previous comment), you should be able to completely remove this additional check

shared/src/data/default-state/emergency-operations-viewport.ts
lukasrad02, line 1: I'm not really a fan of creating a virtual viewport for this feature, since the EOC viewport does not use any viewport features except of the viewport restrictions.  
Especially, there is no need for the EOC viewport to appear on the map, so you already had to implement some workarounds, like the `filter((v) => v.viewportType !== 'eoc')` in the `ol-map-manager` or the permission check for actions.

What do you think about extending the `role` attribute of a `Client` instead? This could also be beneficial to other features, like a separate role for interface signallers (#1076). Since many features currently assume a strictly binary separation (trainer or participant), there could be sub-roles, implemented either as a separate `participantRole` or by transforming the current `role` from a string literal to an object with the general role and the specific role.

shared/src/state-migrations/42-add-eoc-viewport.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        let hasEmergencyOperationsCenter = false;
+        for (const viewportId of Object.keys(typedState.viewports)) {
+            const viewport = typedState.viewports[viewportId]!;
+            if (viewport.viewportType === 'eoc') {
+                hasEmergencyOperationsCenter = true;
+            } else {
+                viewport.viewportType = 'default';
+            }
+        }
```
lukasrad02: You can assume that migration 42 will always be running on a version 41 state, so you don't have to implement the additional checks. Setting the type of all existing viewports to `default` and always adding the EOC viewport should be fine.

shared/src/store/action-reducers/emergency-operation-center.ts
```diff
@@ -67,19 +73,39 @@ export class SendAlarmGroupAction implements Action {
...
+                name,
+                isPrivate
             );
             draftState.eocLog.push(cloneDeepMutable(logEntry));
             return draftState;
         },
-        rights: 'trainer',
+        rights: (client, action) => {
+            if (
+                action.type === '[Emergency Operation Center] Add Log Entry' &&
```
lukasrad02: Checking the type of the action in a permission validator that is used exclusively for a single action seems weird.

I don't think that there will be cases where a single function is used for multiple different actions (as this single function would only be a large switch-case for the different action types), so I'd prefer typing the `action` parameter with the correct, specific action.

And since the `client` is passed into the validator, what do you think about implementing the role check (trainer or participant) in the validator and return a boolean (allow/deny) instead? This would allow even more complex checks (such as "nobody is allowed to set `x` higher than 123).

Quixelation: > I don't think that there will be cases where a single function is used for multiple different actions (as this single function would only be a large switch-case for the different action types), so I'd prefer typing the `action` parameter with the correct, specific action.

I agree with you - I will fix that :+1:

> And since the `client` is passed into the validator, what do you think about implementing the role check (trainer or participant) in the validator and return a boolean (allow/deny) instead? This would allow even more complex checks (such as "nobody is allowed to set `x` higher than 123).

sounds good to me. that would also make the code far more readable :+1:

shared/src/data/default-state/emergency-operations-viewport.ts
Quixelation: Would there be any benefit of changing role to an object except that we could have different "types" of participants, that presumably just have different "views" in the software?

maybe we could also add another attribute to the client that represents the currently shown interface. like `activeView` or `visibleInterface`. 
This way, if - for some reason - trainers want to open a full-page EOC while still being able to access all other administrative settings of the exercise, they could do so.
i don't know how popular this use-case would be though...

shared/src/store/action-reducers/emergency-operation-center.ts
Quixelation: `((client: Client, action: A) => boolean | Role)` could maybe be a good way to implement such a thing to enable for more granular rights, but still allowing the old system of returning a "role-string":

```ts
(client, action) => {
    if(action.value > 123) return false;
    else return "trainer";
}
```

shared/src/data/default-state/emergency-operations-viewport.ts
lukasrad02: Feature-wise, it doesn't matter whether `activeView`/`visibleInterface` is a separate property or part of a complex `role` object. In both cases, we could decide on the semantics of a specific combination of values (e.g., whether the restriction should also apply on trainers. I rather wonder how you would allow trainers to set (and especially unset) the value while they are in the fullscreen EOC, but this is also a UI issue, not an issue of the data model.

> \[T\]hat we could have different "types" of participants, that presumably just have different "views" in the software?

is the main benefit in my opinion, but is again unrelated to the data model (complex object vs. separate properties). Only implementing multiple different types/views with more and more virtual viewports does not seem sensible to me.

shared/src/store/action-reducers/emergency-operation-center.ts
lukasrad02: Also fine for me

frontend/src/app/pages/exercises/exercise/shared/exercise-map/exercise-map.component.html
Quixelation: `*ngIf` causes the openLayerContainer to unload and loading it again would require calling `new OlMapManager()` again - this would require listening to changes on `selectRestrictedViewport`, and [waiting till angular updates the DOM](https://stackoverflow.com/questions/68624678/viewchild-nativeelement-is-undefined-if-content-is-wrapped-up-with-ngif) to be able to access `#openLayersContainer`.

looking at the code again, i think it would be better, if `app-emergency-operations-center-full` (and the logic switching between them) wasn't included in `ExerciseMapComponent` at all, but rather in some kind of Wrapper Component, switching between `app-exercise-map` and `app-emergency-operations-center-full`.
This way, we don't need to care about the problems described above with `*ngIf` and the TimeTravel component woudn't have an ExerciseMap that could also be many different other things.

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-alarm-group-triggers/eoc-alarm-group-triggers.component.html
Quixelation: i could try to combine the new UI with the old one...
the only difference seems to be "one button per alarmgroup" vs. "dropdown for alarmgroups and a submit button".
we could implement a parameter in the `SendAlarmGroupInterfaceComponent` that prevents the initialization of the Hotkey-Layer and switches between the above mentioned changes via `*ngIf`? 
also regarding your other comment https://github.com/hpi-sam/digital-fuesim-manv/pull/1125#discussion_r2469721862
imageimage (1)

lukasrad02: Yes, a combination would be fine for me. However, please check that the UI still works in the IFS modal using only keyboard navigation, since the new UI is used 1:1 in the EOC modal (your second screenshot) and the IFS modal.

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/send-alarm-groups-card/send-alarm-groups-card.component.html
```diff
@@ -0,0 +1,25 @@
...
+        <div class="form-check form-switch mt-2" style="font-size: 16px">
+            <input
+                class="form-check-input"
+                type="checkbox"
+                [(ngModel)]="useComplexLayout"
+                id="complexLayoutToggle"
+            />
+            <label class="form-check-label" for="complexLayoutToggle"
+                >Komplexes Layout</label
+            >
```
hansegucker:
```suggestion
            <label class="form-check-label" for="complexLayoutToggle">
                Komplexes Layout
            </label>
```

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-log-interface/eoc-log-interface.component.ts
```diff
@@ -21,6 +24,10 @@ export class EocLogInterfaceComponent {
 
     public newLogEntry = '';
 
+    public sendingPrivateLog = true;
+    public readonly clientisTrainer =
```
hansegucker:
```suggestion
    public readonly clientIsTrainer =
```

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-log-interface/eoc-log-interface.component.ts
```diff
@@ -31,9 +38,12 @@ export class EocLogInterfaceComponent {
             type: '[Emergency Operation Center] Add Log Entry',
             message: this.newLogEntry,
             name: selectStateSnapshot(selectOwnClient, this.store)!.name,
+            isPrivate: this.clientisTrainer ? this.sendingPrivateLog : false,
```
hansegucker:
```suggestion
            isPrivate: this.clientIsTrainer ? this.sendingPrivateLog : false,
```

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-log-interface/eoc-log-interface.component.html
```diff
@@ -1,47 +1,87 @@
...
+                                [style.color]="
+                                    eocLogEntry.isPrivate ? '#0b02bf' : 'unset'
+                                "
+                            >
+                                <em
+                                    *ngIf="eocLogEntry.isPrivate"
+                                    style="display: block; opacity: 0.8"
+                                >
+                                    [Private Übungsnotiz]</em
+                                >
```
hansegucker:
```suggestion
                                <em
                                    *ngIf="eocLogEntry.isPrivate"
                                    style="display: block; opacity: 0.8"
                                >
                                    [Private Übungsnotiz]
                                </em>
```

frontend/src/app/pages/exercises/exercise/shared/emergency-operations-center/eoc-log-interface/eoc-log-interface.component.html
```diff
@@ -1,47 +1,87 @@
...
+            <div class="form-check form-switch mt-2" *ngIf="clientisTrainer">
+                <input
+                    class="form-check-input"
+                    type="checkbox"
+                    [ngModel]="!sendingPrivateLog"
+                    (ngModelChange)="sendingPrivateLog = !$event"
+                />
+                <label class="form-check-label"
+                    >Als öffentliche Leitstellen-Notiz speichern</label
+                >
```
hansegucker:
```suggestion
                <label class="form-check-label">
                    Als öffentliche Leitstellen-Notiz speichern
                </label>
```

hansegucker (changes requested):
Only some small changes ...

hansegucker (approved):

PR #1127, opened 2025-10-29
Use PostgreSQL 18

hansegucker:
### PR Checklist

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -150,7 +150,7 @@ jobs:
 
         services:
             postgres:
-                image: postgres:14
+                image: postgres:18
                 env:
                     POSTGRES_DB: ${{ env.POSTGRES_DB }}
                     POSTGRES_PASSWORD: ${{ env.POSTGRES_PASSWORD }}
@@ -164,7 +164,7 @@ jobs:
                     --health-interval 10s
                     --health-timeout 5s
                     --health-retries 5
-                    --tmpfs /var/lib/postgresql/data
+                    --tmpfs /var/lib/postgresql
 
         steps:
             - uses: actions/checkout@v6
@@ -210,7 +210,7 @@ jobs:
 
         services:
             postgres:
-                image: postgres:14
+                image: postgres:18
                 env:
                     POSTGRES_DB: ${{ env.POSTGRES_DB }}
                     POSTGRES_PASSWORD: ${{ env.POSTGRES_PASSWORD }}
@@ -224,7 +224,7 @@ jobs:
                     --health-interval 10s
                     --health-timeout 5s
                     --health-retries 5
-                    --tmpfs /var/lib/postgresql/data
+                    --tmpfs /var/lib/postgresql
 
         steps:
             - uses: actions/checkout@v6
@@ -256,7 +256,7 @@ jobs:
 
         services:
             postgres:
-                image: postgres:14
+                image: postgres:18
                 env:
                     POSTGRES_DB: ${{ env.POSTGRES_DB }}
                     POSTGRES_PASSWORD: ${{ env.POSTGRES_PASSWORD }}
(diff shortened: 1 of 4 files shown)


ClFeSc:
**Are you aware of the steps you need to take to successfully deploy these changes?**

You **CANNOT** upgrade PostgreSQL by simply changing a number in some YAML file. This would work for new deployments, but migrating the data has to be a manual step. See https://www.postgresql.org/docs/current/upgrading.html.

ClFeSc:
I've now opened #1129 for this. If you're ready to take all necessary steps in the near future (and this is the reason you've opened this PR) you can obviously close this issue again once done.

hansegucker:
> **Are you aware of the steps you need to take to successfully deploy these changes?**
> 
> You **CANNOT** upgrade PostgreSQL by simply changing a number in some YAML file. This would work for new deployments, but migrating the data has to be a manual step. See https://www.postgresql.org/docs/current/upgrading.html.

Yes, I am aware. This PR can wait until we coordinate the upgrade.

ClFeSc:
> Yes, I am aware. This PR can wait until we coordinate the upgrade.

Okay, nice :) Then disregard my very direct words, I just wanted to prevent a huge problem down the line in case you were not aware.

lukasrad02:
As far as I'm aware, the compose file is provided (1) as a reference *template* for production setups and (2) for local development setups (if using a database is desired). Therefore, I'd argue that changing versions should be fine, given that somebody validated the compatibility with the new version, and there is no need for a coordinated update. Production setups should be running on their own local copy of the compose file and in dev setups, there is typically no need for reliable persistence.

lukasrad02:
Does someone has an idea why the pipeline is failing? PostgreSQL complains about existing data from an older version, which is expected behavior after changing the PostgreSQL version, but the data should not be persisted in the first place…

.github/workflows/pipeline.yml
lukasrad02, line 1: I'd suggest to specify the PostgreSQL version as an environment variable (just like db name and password), so we only have to change it on one place for future updates.

ClFeSc:
> Does someone has an idea why the pipeline is failing? PostgreSQL complains about existing data from an older version, which is expected behavior after changing the PostgreSQL version, but the data should not be persisted in the first place…

Maybe we can circumvent it with https://github.com/docker-library/postgres/pull/1259#issuecomment-2215493289? From the PR linked in the error message from the CI, e.g. https://github.com/hpi-sam/digital-fuesim-manv/actions/runs/18905647202/job/53963317653?pr=1127, `Initialize containers` step.

Or changing `--tmpfs /var/lib/postgresql/data` to `--tmpfs /var/lib/postgresql` although I'm not sure this will actually help as this should just instruct it to use a RAM disk; it should work also on persistent storage, even if slower.

lukasrad02:
> Maybe we can circumvent it with [docker-library/postgres#1259 (comment)](https://github.com/docker-library/postgres/pull/1259#issuecomment-2215493289)? From the PR linked in the error message from the CI, e.g. https://github.com/hpi-sam/digital-fuesim-manv/actions/runs/18905647202/job/53963317653?pr=1127, `Initialize containers` step.
> 
> Or changing `--tmpfs /var/lib/postgresql/data` to `--tmpfs /var/lib/postgresql` although I'm not sure this will actually help as this should just instruct it to use a RAM disk; it should work also on persistent storage, even if slower.

Oh, I didn't notice that the error message even specifies "unused mount/volume". So I think the error is just a precaution of PostgreSQL to ensure that users adjust their mounts and the data does not get lost due to a change of the path.

I'd prefer to adjust the mount, since this is also the officially suggested way:

> The suggested container configuration for 18+ is to place a single mount at /var/lib/postgresql which will then place PostgreSQL data in a subdirectory

We don't need any persistence or other guarantees, but by using the official solution, I hope that we won't run into other issues in the next time. Changing `PGDATA` might also have some side effects in the future…

hansegucker:
I am bringing this PR back to review, as we migrated to PostgreSQL 18 in production last week.

Jogius (approved):
Should we mention somewhere in the Changelog/Docs that an upgrade to this version required manual changes in deployments?

Apart from this, the changes look good to me.

Jogius (approved):

Quixelation (approved):

PR #1128, opened 2025-10-29
Include Webstorm project configuration

fekoch:
This adds a IDE configuration for Jetbrains IDEs, that roughly matches the vscode configuration.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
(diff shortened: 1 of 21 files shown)


lukasrad02:
I'm not using Webstorm (and don't even have it installed), so I can't really review this PR.

One rather general suggestion: For VS Code, we decided to not include the actual config files in the repo, but only templates/examples with a different name. This allows everybody to edit the files according to their preferences (especially the `settings.json`), without worrying about committing those changes by accident, since the "real" config files are part of the `.gitignore`.

If some of the config files added in this PR could contain additional, developer-specific settings, I'd suggest to also use the example file approach. If every file only serves one specific purpose and there is no need for individual adjustments in those files, I'd be fine with this PR.

---

Just a small nitpick: Most of the files don't have a trailing newline

hansegucker (approved):

fekoch:
I checked all files again. I had one wrongly configured runConfiguration that was not system agnostic. If someone wants additional runConfigurations, Webstorm allows setting them without saving them into project configuration files. I also made sure, that there are run configurations for running both the windows and linux-macos backend-start-script 

In my opinion, all settings regarding running prettier on save, codestyle configurations, etc. should be tracked in git.

PR #1138, opened 2025-11-03
Implement `ClientRole` class

Quixelation:
This PR adds a `ClientRole` class which is used to replace and extend the current role-system.
It extends the binary `trainer | participant` to a `mainRole: 'trainer' | 'participant'`, and `specificRole` which can be used to differentiate between different participant types (EOC, interface signallers, ...) and to show different views/interfaces based on the `specificRole`.

See https://github.com/hpi-sam/digital-fuesim-manv/pull/1125#discussion_r2469943874

Test-Scenarios in https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios/pull/20

### PR Checklist

- [X] If this PR adds or changes features or fixes bugs, this has been added to the changelog
- [X] If this PR adds new actions or other ways to alter the state, [test scenarios](https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios) have been added.

### DIFF
--- a/shared/src/models/client-role.ts
+++ b/shared/src/models/client-role.ts
@@ -1,0 +1,29 @@
+import { IsLiteralUnion, IsValue } from '../utils/validators/index.js';
+import type { Role } from './utils/index.js';
+import { getCreate } from './utils/index.js';
+import {
+    roleAllowedValues,
+    type SpecificRole,
+    specificRolesAllowedValues,
+} from './utils/role.js';
+
+export class ClientRole {
+    @IsValue('clientrole' as const)
+    public readonly type = 'clientrole';
+
+    @IsLiteralUnion(roleAllowedValues)
+    public readonly mainRole: Role;
+
+    @IsLiteralUnion(specificRolesAllowedValues)
+    public readonly specificRole: SpecificRole | null;
+
+    /**
+     * @deprecated Use {@link create} instead
+     */
+    constructor(mainRole: Role, specificRole: SpecificRole | null) {
+        this.mainRole = mainRole;
+        this.specificRole = specificRole;
+    }
+
+    static readonly create = getCreate(this);
+}
--- a/shared/src/models/client-role.ts
+++ b/shared/src/models/client-role.ts
@@ -15,12 +15,12 @@ export class ClientRole {
     public readonly mainRole: Role;
 
     @IsLiteralUnion(specificRolesAllowedValues)
-    public readonly specificRole: SpecificRole | null;
+    public readonly specificRole: SpecificRole;
 
     /**
      * @deprecated Use {@link create} instead
      */
-    constructor(mainRole: Role, specificRole: SpecificRole | null) {
+    constructor(mainRole: Role, specificRole: SpecificRole) {
         this.mainRole = mainRole;
         this.specificRole = specificRole;
     }
--- a/shared/src/models/client-role.ts
+++ b/shared/src/models/client-role.ts
@@ -4,17 +4,17 @@ import { getCreate } from './utils/index.js';
 import {
     roleAllowedValues,
     type SpecificRole,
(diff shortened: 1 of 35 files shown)


shared/src/models/client-role.ts
```diff
@@ -0,0 +1,29 @@
...
+import { getCreate } from './utils/index.js';
+import {
+    roleAllowedValues,
+    type SpecificRole,
+    specificRolesAllowedValues,
+} from './utils/role.js';
+
+export class ClientRole {
+    @IsValue('clientrole' as const)
+    public readonly type = 'clientrole';
```
lukasrad02: Please use `camelCase`, matching the class name. Keep in mind to update the migration accordingly.
```suggestion
    @IsValue('clientRole' as const)
    public readonly type = 'clientRole';
```

shared/src/models/utils/role.ts
```diff
@@ -5,3 +5,9 @@ export const roleAllowedValues: AllowedValues<Role> = {
     participant: true,
     trainer: true,
 };
+export type SpecificRole = 'eoc' | 'map-operator' | 'trainer';
```
lukasrad02: Please use `camelCase`

```suggestion
export type SpecificRole = 'eoc' | 'mapOperator' | 'trainer';
```

shared/src/models/utils/role.ts
```diff
@@ -5,3 +5,9 @@ export const roleAllowedValues: AllowedValues<Role> = {
...
+export const specificRolesAllowedValues: AllowedValues<SpecificRole> = {
```
lukasrad02: Use consistent naming

```suggestion
export const specificRoleAllowedValues: AllowedValues<SpecificRole> = {
```

shared/src/state-migrations/42-replace-clientroles.ts
```diff
@@ -0,0 +1,43 @@
+import type { UUID } from '../utils/index.js';
+import type { Migration } from './migration-functions.js';
+
+type OldRole = 'participant' | 'trainer';
+
+interface ClientRole {
+    type: 'clientrole';
+    mainRole: 'participant' | 'trainer';
+    specificRole: 'map-operator' | 'trainer';
+}
```
lukasrad02: Keep in mind to update those values to camelCasing, after changing the main type definitions:

```suggestion
    type: 'clientRole';
    mainRole: 'participant' | 'trainer';
    specificRole: 'mapOperator' | 'trainer';
}
```

shared/src/store/validate-permissions.ts
```diff
@@ -18,13 +18,22 @@ export function validatePermissions(
...
+        (rights === 'participant' || rights === client.role.specificRole)
     ) {
-        return false;
+        return true;
     }
-    // TODO: Validate e.g. only actions in own viewport
-    return true;
+    if (
+        client.role.mainRole === 'trainer' ||
+        client.role.specificRole === rights
```
lukasrad02: Isn't this additional check for the specific role redundant?

If the client is a participant, this check is already part of the previous if-statement's condition. If the client is a trainer, the current check will already short-circuit to `true` and not even evaluate the second option.

lukasrad02 (changes requested):
Since there is no user-facing change, you don't have to add something to the changelog. However, please add test scenarios for your migration!

lukasrad02:
I have approved the test scenarios (<https://github.com/hpi-sam/digital-fuesim-manv-public-test-scenarios/pull/20>), you can merge them now. Afterwards, please update the submodule to the new commit on the test scenarios' `main` branch.

Feel free to re-request a review when you're done to notify me! Apart from updating the test scenarios and the failing DCO check, this PR looks good to me now!

lukasrad02 (approved):

PR #1161, opened 2025-11-18
Use npm 11 and ensure up-to-date lockfiles

lukasrad02:
After #1158, the lockfiles were not up-to-date and running `npm run install:all` caused changes to them.

Since the changes were different depending on the `npm` version used, I've also updated the required npm version to ensure consistent lockfiles.

### PR Checklist

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -44,7 +44,7 @@
             },
             "engines": {
                 "node": ">=22",
-                "npm": ">=10"
+                "npm": ">=11"
             }
         },
         "../shared": {
@@ -81,7 +81,7 @@
             },
             "engines": {
                 "node": ">=22",
-                "npm": ">=10"
+                "npm": ">=11"
             }
         },
         "node_modules/@babel/code-frame": {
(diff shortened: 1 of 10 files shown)


ClFeSc:
> After https://github.com/hpi-sam/digital-fuesim-manv/pull/1158, the lockfiles were not up-to-date and running npm run install:all caused changes to them.

Shouldn't [`npm ci`](https://docs.npmjs.com/cli/v9/commands/npm-ci) have caught those changes and let the CI fail?

lukasrad02:
> > After #1158, the lockfiles were not up-to-date and running npm run install:all caused changes to them.
> 
> Shouldn't [`npm ci`](https://docs.npmjs.com/cli/v9/commands/npm-ci) have caught those changes and let the CI fail?

The only difference were the `"peer": true` lines. As they aren't reflected in the `package.json`, I'm not sure whether `npm ci` is supped to fail, since it only fails if

> dependencies in the package lock do not match those in package.json

However, we have confirmed that the lockfiles were out of sync on multiple systems, so this shouldn't be a issue specific to my (or somebody else's) system

ClFeSc:
> The only difference were the "peer": true lines. As they aren't reflected in the package.json, I'm not sure whether npm ci is supped to fail, since it only fails if
> 
>     dependencies in the package lock do not match those in package.json

Ah, I see.

hansegucker:
@lukasrad02 There seems to be an issue with glob now.

lukasrad02:
> @lukasrad02 There seems to be an issue with glob now.

Yes, I'm aware. But that is not in scope of this PR. I was already planning to resolve the glob issue once this PR is merged to avoid rebasing.

hansegucker (approved):

PR #1165, opened 2025-11-20
Use locale-agnostic string compare in simulation

### LINKED ISSUE #1162: Use locale agnostic string comparison for deterministic reducers
In some reducers, especially those related to the simulation, we're sorting object keys to ensure deterministic behavior. The current approach uses `.sort((a, b) => a.id.localeCompare(b.id))`, for example here: <https://github.com/hpi-sam/digital-fuesim-manv/blob/c187cbabd0d36523c19210c06eaf803d43b2d5d3/shared/src/simulation/activities/reassign-treatments.ts#L91>

`localeCompare` seems to be the only built-in way to compare strings and get a numeric (-1, 0, 1) result. However, sorting can differ among clients, if different locales are used. Although I don't expect there to be issues, since we're mostly sorting UUIDs that don't contain any special characters or accents and our software is targeting German emergency services, we should switch to a locale-agnostic approach to prevent inconsistencies in the first place. (For example, the current approach would already be unstable if names with umlauts were used instead of UUIDs.)

The only option to get a numerical result of a locale-agnostic string comparison in JS seems to be writing a helper function using `===`, `<` and `>`, see <https://stackoverflow.com/a/25150594>.

(Side note: When *displaying* (with no effects on actions or reducers) information in the frontend, we should still use local-aware operations)

lukasrad02:
Closes #1162 

### PR Checklist


fekoch (changes requested):
I think it makes sense to also change the call in `objectToHash()` (`frontend/src/app/shared/pipes/track-by-property/object-to-hash.ts`)

There is also another call in the migration `37-add-patient-identifieres.ts`, but I am unsure, if we want to touch existing migrations...

lukasrad02:
> I think it makes sense to also change the call in `objectToHash()` (`frontend/src/app/shared/pipes/track-by-property/object-to-hash.ts`)

Yes, I missed that since I didn't investigate the usages in `frontend` in that detail. With the way `objectToHash()` is currently used, the actual sort order doesn't matter and must just be stable on a single system. But it should also be fine to replace it so that potential future usages of that helper function work as expected, even across multiple systems.

However, I've just noticed an issue with our testing setup: Tests that (indirectly, i.e., the test imports `object-to-hash.ts` which then imports `stringCompare` from `shared`) import code from `shared` seem to fail. I'll investigate and fix after the weekend.

> There is also another call in the migration `37-add-patient-identifieres.ts`, but I am unsure, if we want to touch existing migrations...

I also wasn't sure about this. In general, migrations *should* even be fixed if we notice a bug in them afterwards (especially since fixing the bug in a later migration might not always be possible, if the faulty migration has already overwritten the original data) and changing the behavior of a migration is not an issue as the migration (in contrast to reducers) will only be run once per exercise and on only one, well-defined system (backend/server).

In case of migration 37, it basically does not matter whether the patients are sorted at all. We just have to assign some identifiers to them so that the field is not empty. But since the IDs didn't exist before, there cannot be any feature that references those IDs and expects them to map to specific patients.

But I think it does still make sense to remove the `localeCompare` from the migration to ensure it cannot serve as a buggy example for future migrations, so I'll update it accordingly.

lukasrad02:
I've now updated the migration in 1d251f9875a87adf1761791bf54010fa928bf184.

For the update in `object-to-hash.ts`, we first need the fix from #1166, since otherwise, the tests would fail on this PR.

fekoch (approved):

lukasrad02:
Now also with `object-to-hash.ts` fixed

fekoch (approved):

Issue #181, opened 2022-02-10
Add exercise-exists route guard


Dassderdie:
>  If an exercise is accessed that does not exist, I must enter a name before being redirected to the starting page.

* Add exercise-exists route guard
* if it fails: display an error message #180 
* Maybe also add an async validator to the joinExercise input at the landing page

PR #1173, opened 2025-11-26
migrate to drizzle

Quixelation:
This is a **draft** PR for a proof-of-concept of how to replace TypeORM with [Drizzle](https://orm.drizzle.team/docs/overview).

There are still some important points missing, like migration and documentation.

### Important Considerations

- The `USE_DB=false` option is implemented by using pglite instead of postgres.

### Todo-List
- [x] `fuesim-server.spec.ts` test needs to be rewritten! --> new architecture
- [x] add db connection test?
- [x] test if existing database from typeORM can safely be switched to use with Drizzle
    - [x]  Tested creating exercise with typeorm --> switching to drizzle orm branch --> do migrations with drizzle --> resume server --> all exercises were loaded correctly and could be used.
- [x] purge script rewrite
- [x] npm test with USE_DB=false (in-memory db) has problems with `generate_uuid_v4()` extension
     - [x] How was this handled in TypeORM --> where should we enable this extension?
- [x] Remove TypeORM package
- [x] Change README
- [x] Test if action.id is set after saving to prevent double creation of action in database

### PR Checklist

- [ ] If this PR adds or changes features or fixes bugs, this has been added to the changelog


backend/src/database/schema.ts
```diff
@@ -0,0 +1,82 @@
...
+            columns: [table.exerciseId],
+            foreignColumns: [exerciseWrapperTable.id],
+            name: 'FK_180a58767f06b503216ba2b0982',
+        })
+            .onUpdate('cascade')
+            .onDelete('cascade'),
+    ]
+);
+
+export const migrations = pgTable('migrations', {
```
fekoch: We should call this also `migrationTable` to be consistent

Quixelation: This table probably still belongs to TypeORM.
Since Drizzle uses it's own table for keeping track of migrations, we can delete this table with a migration done by drizzle.

backend/src/database/services/database-service.ts
```diff
@@ -1,29 +1,130 @@
...
+        if (client instanceof Pool) {
+            await client.end();
+        } else {
+            await client.close();
+        }
+        this._initialized = false;
     }
 }
+
+export type DatabaseConnection = Awaited<
```
fekoch, line 125: we do not need to export this type

backend/src/database/services/database-service.ts
```diff
@@ -1,29 +1,130 @@
...
 
     /**
-     * Wrap all operations into one database transaction. Note that all calls inside MUST use the provided manager.
+     * DO NOT USE WITHOUT `DatabaseService` AS A WRAPPER
      */
-    public async transaction<T>(
-        operation: (manager: EntityManager) => Promise<T>
+    public static async createNewDatabaseConnection(
+        mode: DatabaseConnectionMode = 'default'
     ) {
```
fekoch: this function can return a DatabaseService instance

fekoch (changes requested):
as discussed, we want to keep a simplified version of the layer architecture. Additionally, we want to replace ExerciseWrapper with an ExerciseService class

backend/src/database/repositories/action-repository.ts
```diff
@@ -0,0 +1,109 @@
...
+    public async saveActions(actions: ActionWrapper[], exerciseId: string) {
+        if (actions.length === 0) return;
+
+        const actionsPatch: InferInsertModel<typeof actionTable>[] =
+            actions.map((actionWrapper) => {
+                const action = actionWrapper.getAction();
+                const actionPatch: typeof actionTable.$inferInsert = {
+                    actionString: action.actionString,
+                    emitterId: action.emitterId,
+                    exerciseId,
```
fekoch: Is it a plausible use case to apply actions to a different exercise? Otherwise, we could probably just use the exerciseId of the actionWrapper?

backend/src/database/repositories/action-repository.ts
```diff
@@ -0,0 +1,109 @@
...
+    public async deleteAllForExercise(exerciseId: string) {
+        await this.databaseConnection
+            .delete(actionTable)
+            .where(eq(actionTable.exerciseId, exerciseId));
+    }
+
+    public async saveActions(actions: ActionWrapper[], exerciseId: string) {
+        if (actions.length === 0) return;
+
+        const actionsPatch: InferInsertModel<typeof actionTable>[] =
```
fekoch: This seems very complex. Why do we have to use patches? I would expect that upserting the actions should be fairly straightforward? (Maybe you could add a short comment)

backend/src/database/repositories/action-repository.ts
```diff
@@ -0,0 +1,109 @@
...
+                    index: action.index,
+                };
+                if (action.id !== undefined) {
+                    actionPatch.id = action.id;
+                }
+                return actionPatch;
+            });
+
+        return this.databaseConnection.transaction(async (tx) => {
+            const actionPromises = actionsPatch.map(async (action) => {
```
fekoch: is it not feasible to do this upsert in a single database request? Doing a single Upsert for every action seems a bit wasteful to me

backend/src/database/repositories/exercise-repository.ts
```diff
@@ -0,0 +1,97 @@
...
+import type { ActiveExercise } from '../../exercise/active-exercise.js';
+import { BaseRepository } from './base-repository.js';
+
+export class ExerciseRepository extends BaseRepository {
+    public getExerciseById(id: string) {
+        return this.databaseConnection
+            .select()
+            .from(exerciseTable)
+            .where(eq(exerciseTable.id, id));
+    }
```
fekoch, line 14: this should only ever return a single result or none at all

backend/src/database/repositories/exercise-repository.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Loads the exercise with the corresponding trainer id
```
fekoch:
```suggestion
     * Loads the exercise with the corresponding trainer key
```

backend/src/database/repositories/exercise-repository.ts
```diff
@@ -0,0 +1,97 @@
...
+     */
+    public getExerciseByTrainerKey(trainerKey: string) {
+        return this.databaseConnection
+            .select()
+            .from(exerciseTable)
+            .where(eq(exerciseTable.trainerId, trainerKey));
+    }
+
+    /**
+     * Loads the exercise with the corresponding participant id
```
fekoch:
```suggestion
     * Loads the exercise with the corresponding participant key
```

backend/src/database/repositories/exercise-repository.ts
```diff
@@ -0,0 +1,97 @@
...
+            .values(exercisePatchWithId)
+            .onConflictDoUpdate({
+                target: exerciseTable.id,
+                set: exercisePatchWithId,
+            })
+            .returning();
+        return result;
+    }
+
+    public async createExerciseIfNotExists(activeExercise: ActiveExercise) {
```
fekoch, line 85: same as above, this should only ever return a single exercise or none at all

backend/src/database/repositories/base-repository.ts
```diff
@@ -0,0 +1,51 @@
+import type { Constructor } from 'digital-fuesim-manv-shared';
+import type {
+    DatabaseConnection,
+    DatabaseTransaction,
+} from './../../database/services/database-service.js';
```
fekoch:
```suggestion
} from '../services/database-service.js';
```

backend/src/database/scripts/purge.ts
```diff
@@ -1,28 +1,24 @@
...
-const dataSource = await createNewDataSource().initialize();
+const databaseService = await DatabaseService.createNewDatabaseConnection();
 
-const databaseService = new DatabaseService(dataSource);
+await databaseService.databaseConnection.transaction(async (manager) => {
+    if (
+        DatabaseService.isInMemoryConnection(databaseService.databaseConnection)
+    ) {
+        console.warn(
+            'PURGING IN-MEMORY DATABASE! --> Set DFM_USE_DB=true to purge Postgres Database.'
```
hansegucker:
```suggestion
            'PURGING IN-MEMORY DATABASE! Set DFM_USE_DB=true to purge PostgreSQL database.'
```

backend/src/exercise/http-handler/api/exercise.ts
```diff
@@ -56,27 +54,22 @@ export async function postExercise(
...
+export function getExercise(
+    exerciseId: string,
+    exerciseService: ExerciseService
+): HttpResponse {
+    const exerciseExists = exerciseService.hasExerciseKey(exerciseId);
     return {
         statusCode: exerciseExists ? 200 : 404,
         body: undefined,
     };
 }
```
fekoch, line 64: is there something missing? or should getExercise() not return an exercise?

backend/src/database/services/database-service.ts
```diff
@@ -1,29 +1,106 @@
...
+        });
+        const db = pgliteDrizzle({
+            client: pgLite,
+            schema,
+            logger: Config.dbLogging,
+        });
+        await this.testConnection(db);
+        await db.execute(`CREATE EXTENSION IF NOT EXISTS "uuid-ossp";`);
+        await this.migrate(db);
+        return new DatabaseService(db);
```
hansegucker: I think I would split this into two private methods `createNewPostgresDatabaseConnection` and `createNewPgliteDatabaseConnection` and use an explicit if/else (similar to `migrate`).

backend/src/database/schema.ts
```diff
@@ -0,0 +1,68 @@
...
+});
+export type ExerciseEntry = InferSelectModel<typeof exerciseTable>;
+
+export const actionTable = pgTable(
+    'action_entity',
+    {
+        id: uuid()
+            .default(sql`uuid_generate_v4()`)
+            .primaryKey()
+            .notNull(),
```
hansegucker: Maybe use `BaseEntity.table` here?

backend/src/exercise/active-exercise.ts
```diff
@@ -0,0 +1,303 @@
...
+            initialState
+        );
+
+        return exercise;
+    }
+
+    /**
+     * Be very careful when using this. - Use {@link create} instead for most use cases.
+     * This constructor does not guarantee a valid entity.
+     */
```
fekoch: does this still hold true?

fekoch (changes requested):

backend/src/exercise/http-handler/secure-http.ts
```diff
@@ -21,7 +25,7 @@ export async function secureHttp<Result extends object | undefined>(
     } catch (error: unknown) {
         // Try sending 500 response
         try {
-            const message = `An error occurred on http request: ${error}`;
+            const message = `An error occurred on http request ${req.path}: ${error}`;
```
hansegucker, line 28: Nice!

backend/package.json
```diff
@@ -17,16 +17,10 @@
...
-        "migration:generate:windows": ".\\src\\database\\migrations\\generate-migration.bat",
-        "migration:run": "npm run typeorm:cli migration:run",
-        "migration:run:prod": "npm run typeorm:cli:prod migration:run",
-        "migration:run:windows": "npm run typeorm:cli:windows migration:run",
-        "migration:revert": "npm run typeorm:cli migration:revert",
-        "migration:revert:windows": "npm run typeorm:cli:windows migration:revert",
+        "migration:generate": "npx drizzle-kit generate",
+        "migration:run": "npx drizzle-kit migrate",
+        "migration:check": "npx drizzle-kit check",
+        "db:studo": "npx drizzle-kit studio",
```
hansegucker:
```suggestion
        "db:studio": "npx drizzle-kit studio",
```

test-scenarios
hansegucker, line 1: I am not sure why there is an update of the test scenarios here?

hansegucker (changes requested):
I really like the new structure, thanks!

backend/src/exercise/http-handler/api/exercise.ts
Quixelation: I was confused too - but this endpoint seems to only have been used for checking whether an exercise exists or not
image

image

Quixelation: (screenshots from before drizzle migration)

ClFeSc: This is by design, correct. We never needed more from this route. Joining an exercise was always done via the websocket. We had the problem of #181 that the websocket would only tell you that an exercise does not exist after entering your name. And this route, introduced in #238, was meant to fill that gap without e.g. transporting unnecessary data like the exercise state.

backend/src/database/repositories/action-repository.ts
Quixelation: The problem initially was, that the ActionWrappers could get created during a state, where the exercise has not yet been saved in the Database (ExerciseFactory.fromFile) and therefore doesn't have an id yet.

We could circumvent this probem by using actionWrapper.exercise.exerciseId since, exerciseService.loadExercise() always makes sure the ExerciseWrapper has been saved to the DB and therefore has an id.

I'll look into changing that.

backend/src/exercise/active-exercise.ts
Quixelation: no

backend/src/database/repositories/action-repository.ts
Quixelation: we need to prevent that actions get saved in the database twice.
since actions dont have a uuid, their "uniqueness" is only given by the id, assigned by the database.
therefore we need to set the action id to the one returned by the database, until the action is removed from `temporaryActionHistory` (the array that is periodically saved to the database), to do an update instead of an insert, if the action happens to be *still* in `temporaryActionHistory` (here: "the array") and the array gets saved *again*.

This is probably a niche use-case, but an example of it could be a database transaction for saving the actions (here: `saveTick`) taking so long, that the next `saveTick` is called, before the first `saveTick` could properly remove the actions from the array.


now the fun part:
I did that to make sure that the order of `RETURNING` is still the same as in `VALUES`, to correctly assign the ids we got from the database to the respective actions. 
more: [https://stackoverflow.com/questions/5439293/is-insert-returning-guaranteed-to-return-things-in-the-right-order](https://stackoverflow.com/questions/5439293/is-insert-returning-guaranteed-to-return-things-in-the-right-order)

Quixelation: This would be a (quickly botched together) example of how a batched `INSERT` could look like:

```ts
public async saveActions(actions: ActionWrapper[]) {
    if (actions.length === 0) return;

    const dbInsertResult = await this.databaseConnection
        .insert(actionTable)
        .values(actions
            .map(actionWrapper => ({ actionWrapper, action: actionWrapper.getAction() }))
            .map(({ actionWrapper, action }) => ({
                actionString: action.actionString,
                emitterId: action.emitterId,
                exerciseId: actionWrapper.exercise.exerciseId,
                index: action.index,
                id: action.id,
            }))
        )
        .onConflictDoUpdate({
            target: actionTable.id,
            set: upsertHelper(actionTable),
        })
        .returning();

    for (let i = 0; i < dbInsertResult.length; i++) {
        const element = dbInsertResult[i];
        if (element?.id === undefined) {
            throw new Error('Could not upsert action');
        }

        const matchingAction = actions[i]

        if (!matchingAction) {
            throw new Error('Could not find matching action after upsert');
        };

        if (matchingAction) {
            matchingAction.getAction().id = element.id;
        }
        console.log(matchingAction.getAction());
    }
}
```

backend/src/database/schema.ts
```diff
@@ -0,0 +1,63 @@
...
+});
+export type ExerciseEntry = InferSelectModel<typeof exerciseTable>;
+
+export const actionTable = pgTable(
+    'action_entity',
+    {
+        ...baseTable,
+        emitterId: uuid(),
+        // You can use { mode: "bigint" } if numbers are exceeding js number limitations
+        index: bigint({ mode: 'number' }).notNull(),
```
fekoch, line 38: I am unsure of what this comment is trying to tell me

Quixelation: This comment was left automatically as part of drizzle pulling the first schema directly from the database.
doc-ref: https://orm.drizzle.team/docs/column-types/pg#bigint

```ts
// will be inferred as `number`
bigint: bigint({ mode: 'number' })
// will be inferred as `bigint`
bigint: bigint({ mode: 'bigint' })
```

i thought it would be useful in case we (or future students) run into issues here - but we can also just remove it :shrug:

fekoch (approved):
lgtm

hansegucker (approved):

PR #1182, opened 2025-12-03
Reduce  Usage of `StrictObject`

lukasrad02:
### PR Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,6 @@ Look at the [benchmark readme](./benchmark/README.md) for more information.
 ## Styleguide
 
 - names are never unique, ids are
-- Use [StrictObject](shared\src\utils\strict-object.ts) instead of `Object` wherever possible
 - A leading underscore should only be used
     - for private properties that may be used with getters/setters
     - to resolve certain naming conflicts (e.g. `.some(_item => ...)`)
--- a/benchmark/src/steps.ts
+++ b/benchmark/src/steps.ts
@@ -181,7 +181,7 @@ export const steps: Step<StepState>[] = [
             print(
                 // In the object are only entries we explicitly set -> no need to check for undefined
                 (
-                    TypeAssertedObject.entries(sortedTotalTimePerAction) as [
+                    Object.entries(sortedTotalTimePerAction) as [
                         ExerciseAction['type'],
                         number,
                     ][]
@@ -245,9 +245,10 @@ export const steps: Step<StepState>[] = [
             print(
                 // In the object are only entries we explicitly set -> no need to check for undefined
                 (
-                    TypeAssertedObject.entries(
-                        sortedNumberOfActionsPerType
-                    ) as [ExerciseAction['type'], number][]
+                    Object.entries(sortedNumberOfActionsPerType) as [
+                        ExerciseAction['type'],
+                        number,
+                    ][]
                 )
                     .map(([type, amount]) => `${type}: ${amount}`)
                     .join(', ')
--- a/benchmark/src/steps.ts
+++ b/benchmark/src/steps.ts
@@ -9,7 +9,7 @@ import {
     migrateStateExport,
     reduceExerciseState,
     sortObject,
-    StrictObject,
+    TypeAssertedObject,
     validateExerciseExport,
 } from 'fuesim-digital-shared';
 import { produce, freeze } from 'immer';
@@ -181,7 +181,7 @@ export const steps: Step<StepState>[] = [
             print(
                 // In the object are only entries we explicitly set -> no need to check for undefined
                 (
-                    StrictObject.entries(sortedTotalTimePerAction) as [
+                    TypeAssertedObject.entries(sortedTotalTimePerAction) as [
                         ExerciseAction['type'],
                         number,
                     ][]
@@ -197,14 +197,14 @@ export const steps: Step<StepState>[] = [
     new CalculationStep(
         'mostExpensiveAction',
         ({ benchmarkActions, newImmerDraft }) => {
-            const mostExpensiveAction = StrictObject.entries(
+            const mostExpensiveAction = TypeAssertedObject.entries(
                 benchmarkActions!
             )[0];
             if (!mostExpensiveAction) {
                 return `No actions`;
             }
             const summedUpExerciseTime =
-                StrictObject.values(benchmarkActions!).reduce(
+                TypeAssertedObject.values(benchmarkActions!).reduce(
                     // In the object are only entries we explicitly set
                     (totalTime, timePerAction) => totalTime! + timePerAction!,
                     0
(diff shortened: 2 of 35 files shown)


ClFeSc:
I like the idea! However, you might need to allow non-`string` keys (i.e. `number | string | symbol`) in `StrictObject` to allow usage of its method in more places. Currently, you only allow `string` which seems to be lost, however, at some point, leading to errors like `keyof T` cannot be assigned to `string`, although we have the constraint `T extends { [key: string]: any }` (e.g. in `shared/src/utils/sort-object.ts`).

fekoch:
After looking at `StrictObject`, I am unsure if it is desirable to use StrictObject.* as much as possible. It seems to me, as if it only serves as a (hidden) `as` directive?

For example:
```typescript
import {StrictObject} from "./strict-object";

interface Foo {
    a: number;
}

interface Bar extends Foo {
    b: number;
}

const c: Bar = {
    a: 1,
    b: 2,
};

function doSomthingToFoo(bar: Bar): Foo {
    return bar;
}

const keys = StrictObject.keys(doSomthingToFoo(c)) satisfies (keyof Foo)[]; // <- incorrect type

console.log(keys) // [ 'a', 'b' ]
```

This compiles without issues but results in unexpected behaviour.

ClFeSc:
Yes. As described in the doc comment of the `StrictObject` namespace, TypeScript's type system is not capable of and not designed to ensure *only* the typed properties are present. If no-one uses `as any` or similar to abuse, it will guarantee that a property that the typings say should be there is in fact there, the other direction, however, cannot hold as TS only checks whether the expected interface is present. `StrictObject` is a helper introduced to simplify some code where we know exactly which properties are present and want to transfer this knowledge into the type system.

ClFeSc:
An idea might be to introduce also a `WeakObject` or whatever that is the same as `Object` but makes you think about what you actually need as currently I could imagine some usages of `Object` being happily compatible with `StrictObject` but are not using it because it was either copied or the programmer simply did not think about it.

lukasrad02:
> I am unsure if it is desirable to use `StrictObject` as much as possible

At least this is the current recommendation of the styleguide: <https://github.com/hpi-sam/digital-fuesim-manv/blob/f77dae94c7711fc3908de66bebf7df83d3b32660/README.md?plain=1#L154>

My initial intention with this PR was not to introduce new rules but only enforce the existing styleguide with a linter rule, since I noticed many cases where `Object` was used, although `StrictObject` could also be used. As Clemens said:

> I could imagine some usages of Object being happily compatible with StrictObject but are not using it because it was either copied or the programmer simply did not think about it

I would *guess* that in most cases, it makes sense to use `StrictObject`, since we mostly use `keys()`/`values()`/`entries()` to iterate through `UUIDMap`s, `ResourceDescription`s, or similarly structured objects where we don't have any inheritance or other cases of re-typing.

If `StrictObject` is actually "wrong" (i.e., there are possible keys/values that don't match the type definition), one can still disable the rule for that line and hopefully use that as a reminder to (a) re-think the type definitions for that piece of code (why is `StrictObject` providing wrong typings`) or (b) provide an explanation of the object's structure in a comment.

I think that overall, the benefit of stricter typings outweighs the potential confusion due to type definitions that are to narrow. Without `StrictObject`, people might just use assertions or type casts in their code since they "know" that the type is right, which can hide type errors if the input object changes its structure during a refactoring.

lukasrad02:
> However, you might need to allow non-`string` keys (i.e. `number | string | symbol`) in `StrictObject` to allow usage of its method in more places. Currently, you only allow `string` which seems to be lost, however, at some point, leading to errors like `keyof T` cannot be assigned to `string`, although we have the constraint `T extends { [key: string]: any }` (e.g. in `shared/src/utils/sort-object.ts`).

I'll look into this

fekoch:
> StrictObject is a helper introduced to simplify some code where we know exactly which properties are present and want to transfer this knowledge into the type system

Do you have an example where this is actually necessary? The examples I tested all infer the types without issues while just using `Object` (or in some cases, there are even cleaner ways to do it, without even needing `Object.` at all)

lukasrad02:
> > StrictObject is a helper introduced to simplify some code where we know exactly which properties are present and want to transfer this knowledge into the type system
> 
> Do you have an example where this is actually necessary? The examples I tested all infer the types without issues while just using `Object`

`Object.values()` typically works well, but `Object.keys()` (or `.entries()`) suffers from the fact that TypeScript by default does not assume that no other keys than specified by the type definition are present (see [Clemens' comment](https://github.com/hpi-sam/digital-fuesim-manv/pull/1182#issuecomment-3607933684)). So searching for `StrictObject.keys` should reveal some examples where we actually benefit from the more strict typings.

The first example I found would be <https://github.com/hpi-sam/digital-fuesim-manv/blob/f77dae94c7711fc3908de66bebf7df83d3b32660/shared/src/simulation/activities/count-patients.ts#L65-L69>

If we were using `Object.keys` here, `patientStatus` would be typed as `string` and thus, indexing `patientCount` with `patientStatus` would not work (*Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '[…]'.*).

In general, `StrictObject` is helpful as soon as we deal with literal union types.

> or in some cases, there are even cleaner ways to do it, without even needing `Object.` at all

Obviously, there could be a pre-defined array of all possible `PatientStatus` values here so that we don't have to collect the keys from `patientStatusAllowedValues`. But with TypeScript's type system, we can't ensure the array contains all possible values, while we can ensure that an object has all possible values as keys.

lukasrad02:
*(Summary of or in-person discussion from today)*

- `StrictObject` is misleading, as it can result in wrong type definitions (see [Felix' comment](https://github.com/hpi-sam/digital-fuesim-manv/pull/1182#issuecomment-3607812482)). We want to rename it so that it reflect its actual behavior (something like "AssertPropertiesObject", a good name is still to be found)
- We don't want to add the linter rules proposed in this PR. Instead, the recommendation to always use `StrictObject` if possible should be dropped
- We want to reduce the overall number of `(Strict)Object.{keys, values, entries}()`-calls in favor of approaches that use less meta-programming. Removing some hardcoded types of domain objects (e.g. #1124) is already in progress, an example refactoring for the code discussed in [this comment](https://github.com/hpi-sam/digital-fuesim-manv/pull/1182#issuecomment-3608969424) is shown below. It is yet to be discussed whether this should only be a guideline for future PRs or whether existing code should also be refactored (maybe even with a linter warning for any use of `(Strict)Object.{keys, values, entries}()`), so that legacy code does not serve as a bad example for newly implemented features

---

As a refactoring of <https://github.com/hpi-sam/digital-fuesim-manv/blob/f77dae94c7711fc3908de66bebf7df83d3b32660/shared/src/simulation/activities/count-patients.ts#L65-L69>, one could (as suggested by Felix)

1. Zero-initialize `patientCount` (TypeScript will catch missing status colors, so this is type-safe)
2. Loop over all patients and increment the respective entry `patientCount`

This approach neither requires `Object.keys()` metaprogramming, nor is the final `as ResourceDescription<PatientStatus>` cast required

lukasrad02:
I've now implemented the following changes:

- Remove recommendation to use `StrictObject` from styleguide
- Rename `StrictObject` to `TypeAssertedObject` and highlight it's limitations in the JSdoc comment
- Use normal `Object` wherever possible (no compiler errors or linter warnings afterwards)
- Create #1388 for the remaining work (check how to get rid of `TypeAssertedObject` completely and reduce `Object` usage)

fekoch (approved):
Thank you! Looks good👍

PR #1202, opened 2026-01-14
fix error with state-export

### LINKED ISSUE #1200: Export of current simulation fails
When exporting current simulation, the following error occurs:

Image

Image

Quixelation:
This PR fixes #1200 so that exercise history exports no longer fail.

- it also adds an error-toast-message to signal to the user that a history-export has failed
- and 2 new simple test cases for history exports

### PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 
 ## [Unreleased]
 
-### Fixed
-
-- Exporting the history of an exercise no longer fails silently.
-
 ## [0.10.0] - 2025-12-08
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 
 ## [Unreleased]
 
+### Fixed
+
+- Exporting the history of an exercise no longer fails silently.
+
 ## [0.10.0] - 2025-12-08
 
 ### Added
--- a/backend/src/database/services/exercise-service.ts
+++ b/backend/src/database/services/exercise-service.ts
@@ -179,7 +179,7 @@ export class ExerciseService {
             throw new UnknownExerciseError(exerciseKey);
         const completeHistory: ExerciseTimeline['actionsWrappers'] = [
             ...(
-                await this.actionRepository.getActionsForExerciseId(exerciseKey)
+                await this.actionRepository.getActionsForExerciseId(activeExercise.exerciseId)
             ).map((action) => ({
                 action: action.actionString,
                 emitterId: action.emitterId,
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 
 ## [Unreleased]
 
+### Fixed
+
+- Exporting the history of an exercise no longer fails silently.
```
lukasrad02: I'm pretty sure the bug was caused by #1173, i.e., it was added after the last release. In this case, you don't have to add this to the changelog

lukasrad02:
Maybe its worth to do a small refactoring or add doc comments (in a separate PR so we can already merge this fix on `dev`) to prevent such issues in the future?

Currently, there is no strict separation between an `exerciseKey` (numerical, user-facing) and an `exerciseId` (UUID, used in the DB(?)). In the existing code, we always used "id" for what's called "key" now. Even the caller of `getTimeline(exerciseKey: string)` calls it with a variable named `exerciseId` as parameter for the key…  
One suggestion that came into my mind would be to use "pk" (primary key) for the UUIDs used in the database (from how it's typically done in Django, I don't know whether Drizzle might have its own conventions for this)

lukasrad02 (approved):
Fix looks good and I can confirm that the issue does not occur any longer.

Quixelation:
> Maybe its worth to do a small refactoring or add doc comments (in a separate PR so we can already merge this fix on `dev`) to prevent such issues in the future?
> 
> Currently, there is no strict separation between an `exerciseKey` (numerical, user-facing) and an `exerciseId` (UUID, used in the DB(?)). In the existing code, we always used "id" for what's called "key" now. [...] One suggestion that came into my mind would be to use "pk" (primary key) for the UUIDs used in the database (from how it's typically done in Django, I don't know whether Drizzle might have its own conventions for this)

yeah - the current seperation is a bit... meh (my pain-point is especially no differentiation/warning via typescript).
the backend should actually use "id" everywhere where we need the actual database id; and "key" whenever we are referring to a public facing exercise-identifier. 

the problem here should've been discovered by me earlier, since I was passing a `key` into a `get...For...Id()` method;
but i probably migrated this code, before we decided to differentiate `id`, `participantId` and `trainerId` with the above mentioned `id` vs `key` --> so i didn't notice it

---

i am unsure whether renaming the `id` attribute to `pk` (or similar) would help - 

the erroring method was `getActionsForExerciseId(exerciseId: string)`.
my mistake was inputting the `exerciseKey` instead of the `exerciseId` - in my opinion, the wording is clear here.

renaming to `pk` instead of `id` would result in (i am guessing that this is what you meant?):
`getActionsForExercisePk(exercisePk: string)` and could also lead to confusion regarding `PK=ParticipantKey | PrimaryKey | ...Key`?

---

it is important to note though, that the migration from `trainerId` to `trainerKey` has only been applied to the backend yet

---

>  Even the caller of `getTimeline(exerciseKey: string)` calls it with a variable named `exerciseId` as parameter for the key…

the http code doesnt seem to be migrated to the new terminology. i would need to update that.

PR #1217, opened 2026-01-26
Add OIDC Auth

Quixelation:
This PR adds the (non-optional) ability to add an OIDC-Provider for SSO into DFM.

Includes:
- Sessions
- OIDC-Flow
- Remove old sessions
- a working pipeline

Missing:
- tests

### PR Checklist


backend/src/database/schema.ts
```diff
@@ -51,6 +53,35 @@ export const actionTable = pgTable(
...
+    /**
+     * This should always be the sub claim from the OIDC provider
+     */
+    id: varchar().primaryKey().notNull(),
+    username: varchar().notNull(),
+    display_name: varchar().notNull(),
+    updatedAt: timestamp({ mode: 'date', precision: 3 })
+        .notNull()
+        .$defaultFn(() => new Date())
+        .$onUpdateFn(() => new Date()),
```
hansegucker: Isn't there a Drizzle built-in for this? `defaultNow` and `onUpdateNow`

backend/src/database/schema.ts
```diff
@@ -51,6 +53,35 @@ export const actionTable = pgTable(
 );
 export type ActionEntry = InferSelectModel<typeof actionTable>;
 
+export const userTable = pgTable('users', {
+    /**
+     * This should always be the sub claim from the OIDC provider
+     */
+    id: varchar().primaryKey().notNull(),
+    username: varchar().notNull(),
+    display_name: varchar().notNull(),
```
hansegucker:
```suggestion
    displayName: varchar().notNull(),
```

backend/src/database/schema.ts
```diff
@@ -51,6 +53,35 @@ export const actionTable = pgTable(
...
+    id: varchar().primaryKey().notNull(),
+    userId: varchar()
+        .notNull()
+        .references(() => userTable.id, {
+            onDelete: 'cascade',
+            onUpdate: 'cascade',
+        }),
+    createdAt: timestamp({ mode: 'date', precision: 3 })
+        .notNull()
+        .$defaultFn(() => new Date()),
```
hansegucker: same as above

backend/src/exercise/http-server.ts
```diff
@@ -22,10 +25,21 @@ export class ExerciseHttpServer {
     public constructor(
         app: Express,
         databaseService: DatabaseService,
-        exerciseService: ExerciseService
+        exerciseService: ExerciseService,
+        authService: AuthService
     ) {
         // TODO: Temporary allow all
-        app.use(cors());
+        // @Quixelation --> We need to restrict this in the future, especially when we have auth in place
```
hansegucker: Relevant?

frontend/src/app/core/auth.service.ts
```diff
@@ -0,0 +1,135 @@
...
+                    .catch((err) => {
+                        this.messageService.postMessage({
+                            title: 'Sitzung abgelaufen',
+                            color: 'warning',
+                            body: 'Ihre Sitzung konnte nicht verlängert werden. Bitte melden Sie sich erneut an.',
+                        });
+                    });
+            },
+            15 * 60 * 1000
+        ); // every 15 minutes
```
hansegucker: Maybe move this to a class constant.

shared/src/http-interfaces.ts
```diff
@@ -7,3 +7,23 @@ export interface ExerciseAccessIds {
...
+              readonly id: string;
+              readonly display_name: string;
+              readonly username: string;
+          }
+        /* not logged in */
+        | null
+        /* no data yet */
+        | undefined;
+    expired?: boolean;
+}
```
hansegucker: Maybe zod?

hansegucker (changes requested):

.env.example
```diff
@@ -50,3 +57,9 @@ DFM_DB_HOST=db
 DFM_DB_PORT=5432
 # Whether or not to log database queries to stdout
 DFM_DB_LOG=false
+
+# Authelia
```
lukasrad02: This should be more generic

```suggestion
# OIDC
```

.github/workflows/pipeline.yml
```diff
@@ -144,6 +144,14 @@ jobs:
                     --health-timeout 5s
                     --health-retries 5
                     --tmpfs /var/lib/postgresql/data
+            authelia:
+                image: docker.io/authelia/authelia:latest
+                restart: unless-stopped
+                ports:
+                    - 127.0.0.1:9091:9091
```
lukasrad02: If possible, use dynamically assigned ports (see the `ports` section of the PostgreSQL container). The assigned port number can be accessed using `job.services.authelia.ports['9091']` (see the `Create .env` step of this job)

In most cases, this does not matter, but it can be helpful for parallelization on selfhosted runners (last semester, we ran into issues, as only one job at a time was able to spawn the PostgreSQL container…)

lukasrad02 (commented):
Definitely not a full review. I just looked into this out of curiosity and noticed some things…

.github/workflows/pipeline.yml
Quixelation: there is currently the problem that the authelia container wont startup without the config file - but the config file is only present after actions/checkout has run.

i solved this by manually starting up a docker container and grepping the port-number - it works but is far from beautiful.
do you know any better methods?

shared/src/http-interfaces.ts
Quixelation: do you want to parse/validate the http body in the frontend?
i dont see a use-case for zod here - unless we want to do body validation?

.github/workflows/pipeline.yml
lukasrad02: I don't know what you're referring to.

You can use

```yaml
ports:
  - 9091
```

in the `service` spec and port 9091 will be forwarded to a randomly chosen available port on the host machine. This does not need any kind of config file for Authelia and shouldn't affect start-up, since the software in the container does not know anything about the outside port management.

You only have to retrieve the randomly chosen port to write it to the config file of DFM, so it can connect to Authelia

Quixelation: oh yes - that worked fine - thank you.

but there is a config file in `/docker/authelia-config/configuration.yml`.
authelia shuts down again if this file is not present:
image
this causes the pipeline to fail initializing the container, since the repo files are not present in the runner yet - it seems.

this is why i chose to manually start the docker container, so that the repo files - including the configuration file - could be pulled, so that authelia can actually start up.

lukasrad02: My only idea would be to use environment variables instead of the config file: <https://www.authelia.com/configuration/methods/environment/>

But I don't know whether it is possible to get rid of the warning if the config file does not exist at all. Maybe you can fix this with a custom entrypoint, like

```yaml
options: |
  --entrypoint "/bin/sh -c '<some command to create a minimal config file> && <original entrypoint of the image>'"
```

backend/src/database/schema.ts
Quixelation: i couldn't find any mention of onUpdtaeNow() ? for postgres atleast

backend/src/exercise/http-server.ts
Quixelation: es wäre ggf. hilfreich, dass man auf die APIs nicht von anderen Webseiten außer unserer eigenen zugreifen könnte. (?)

hansegucker: Yeah, we should configure this that way.

.github/workflows/pipeline.yml
```diff
@@ -147,6 +147,11 @@ jobs:
 
         steps:
             - uses: actions/checkout@v3
+            - name: Start Authelia OIDC Provider
+              run: |
+                  export AUTHELIA_CONTAINER_ID=$(docker run -p 9091/tcp -v ${{ github.action_path }}/docker/authelia-config:/config:ro -v authelia-secrets:/secrets -d docker.io/authelia/authelia:latest)
+                  export AUTHELIA_CONTAINER_PORT=$(docker port $AUTHELIA_CONTAINER_ID | grep -e ":[0-9]*" -o -m1 | cut -c 2-)
```
hansegucker:
```suggestion
                  export AUTHELIA_CONTAINER_PORT=$(docker port $AUTHELIA_CONTAINER_ID 9091/tcp | grep 0.0.0.0 | cut -c 9-
```

frontend/src/app/shared/components/user-account-navbar-item/user-account-navbar-item.component.html
```diff
@@ -0,0 +1,17 @@
...
+        <span ngbDropdownToggle style="cursor: pointer">{{
+            userData.displayName
+        }}</span>
+        <div ngbDropdownMenu>
+            <a ngbDropdownItem [href]="logoutUrl">Ausloggen</a>
+        </div>
+    </div>
+} @else if (userData === null) {
+    <div class="d-flex gap-2">
+        <a class="btn btn-primary" [href]="loginUrl"> Anmelden </a>
```
hansegucker:
```suggestion
        <a class="btn btn-primary" [href]="loginUrl">Anmelden</a>
```

docker/authelia-config/users.yml
```diff
@@ -0,0 +1,18 @@
+# yaml-language-server: $schema=https://www.authelia.com/schemas/latest/json-schema/user-database.json
+# for creating passwords: docker run --rm -it authelia/authelia:latest authelia crypto hash generate argon2
+users:
+    demo1:
+        disabled: false
+        displayname: 'John Doe'
+        password: '$argon2id$v=19$m=65536,t=3,p=4$RrAj9JJ+i2YCeX9d0o7DZA$S1ZRmsjwTWa1zMz3bcpHjQ1ae31tJf0Mt1vu0UKQcFY'
```
fekoch, line 8: Do we maybe want to version the development password somewhere in plaintext?

Quixelation: yes. it's "demo"

hansegucker (approved):

backend/src/database/services/database-service.ts
```diff
@@ -122,6 +122,16 @@ export class DatabaseService {
     }
 }
 
+export function onlySingle<T>(array: T[]): T | null {
```
fekoch: What do you think about moving the function into a utils file? (or maybe into the base repository)

backend/src/utils/frontend-http-helper.ts
```diff
@@ -0,0 +1,16 @@
...
+    if (data?.logoutstatus) {
+        url.searchParams.append('logoutstatus', data.logoutstatus);
+    }
+    if (data?.loginfailure) {
+        url.searchParams.append('loginfailure', data.loginfailure);
+    }
+    if (data?.loginsuccess) {
+        url.searchParams.append('loginsuccess', 'true');
+    }
+    return url.toString();
```
fekoch, line 15: Can we rename `logoutstatus`, `loginfailure` and `loginsuccess` to camelCase`?

Both here and in `AuthQueryParams`

backend/src/auth/oidc-service.ts
```diff
@@ -0,0 +1,243 @@
...
+                this._config = await oidc.discovery(
+                    new URL(Config.authUrl),
+                    Config.authClientId,
+                    Config.authClientSecret,
+                    undefined,
+                    {
+                        execute:
+                            process.env['NODE_ENV'] === 'development'
+                                ? [
+                                      /**
```
fekoch:
```suggestion
                                      /*
```

backend/src/auth/auth-service.ts
```diff
@@ -0,0 +1,97 @@
...
+                username: data.user.username,
+            });
+
+            const sessionRepoTrans =
+                this.sessionRepository.withConnection(userRepoTrans);
+            const sessionToken = await sessionRepoTrans.createSession({
+                validityDurationSeconds: this.SESSION_DURATION_S,
+                accessToken: data.accessToken,
+                userId: data.user.id,
+            });
```
fekoch:
```suggestion
        return this.userRepository.transaction(async (userRepoTransaction) => {
            await userRepoTransaction.upsertUser({
                id: data.user.id,
                displayName: data.user.displayName,
                username: data.user.username,
            });

            const sessionRepoTransaction =
                this.sessionRepository.withConnection(userRepoTransaction);
            const sessionToken = await sessionRepoTrans.createSession({
                validityDurationSeconds: this.SESSION_DURATION_S,
                accessToken: data.accessToken,
                userId: data.user.id,
            });
```

backend/src/auth/oidc-service.ts
```diff
@@ -0,0 +1,243 @@
...
+    }): Promise<{
+        url: URL;
+        cookies: { name: string; value: string }[];
+    }> {
+        // see https://github.com/panva/openid-client/blob/01f5fe37cbe78c8c2624f0e568fc3a7d4d2386eb/README.md
+        const codeVerifier = oidc.randomPKCECodeVerifier();
+        const codeChallenge =
+            await oidc.calculatePKCECodeChallenge(codeVerifier);
+        const state =
+            oidc.randomState().slice(0, 43).padEnd(43, '=') +
```
fekoch: I am confused by the `43` here. What is it's meaning? It is also used in `handleCallback()`.
Can you move it into a named constant?

backend/src/database/services/database-service.ts
Quixelation: into base-repository sounds good.
as a method or just a standalone function?
i think as a method would be preferred as to avoid confusion, to not use it in any other use case?

backend/src/auth/oidc-service.ts
```diff
@@ -0,0 +1,243 @@
...
+                path: '/api/auth/oidc-callback',
+                sameSite: 'lax',
+            });
+        }
+
+        res.redirect(authUrl.url.toString());
+    }
+
+    public async handleCallback(req: ExpressRequest, res: ExpressResponse) {
+        const state = req.cookies['state'];
```
fekoch:
```suggestion
        const state = req.cookies['state'] as string | undefined;
```

backend/src/database/services/database-service.ts
fekoch: Yes, I think if we put it into the base repository, that would be best :+1:

fekoch (approved):

PR #1228, opened 2026-02-02
Modernize pipeline

### LINKED ISSUE #1220: Improve pipeline caching
We currently cache all `node_modules` folders in the CI. This is [explicitly not recommended](https://github.com/actions/cache/blob/main/examples.md#node---npm). `actions/setup-node` by itself is capable of caching the system-wide package cache of npm, so that future installs don't need network access and can just copy the packages from that cache, which should be similarly fast.

For the caches that contain build artifacts (e.g. `shared/dist`), we [should use artifacts](https://docs.github.com/en/actions/concepts/workflows-and-actions/workflow-artifacts#artifacts-versus-dependency-caching) instead. (Or, if we want to keep using caches, we should fix the cache key. Currently, an update of one of the dependencies would not trigger a cache update, since only `shared/src` is cached…)

### LINKED ISSUE #1224: Build Docker Image in Pull Request Pipelines
Currently, the docker image is only built for commits on `dev` and `main`.

We should also build the Docker image in the pull request pipelines to catch errors in the build process. While most of the time, the docker build should complete without issues if the bare metal build was successful, this is not always the case. For example, when updating the Node version, the Dockerfile also has to use the correct base image.

lukasrad02:
This PR contains several changes to the GitHub pipeline:

- Update all used actions to the latest version
- Change caching & artifact upload behavior, fixes #1220
- Build docker container on every pipeline run and require successful build, fixes #1224
- Publish docker images to GHCR, see #1236 (for a smooth transition, images are still published to dockerhub, so the issue should remain opened)

### PR Checklist


.github/workflows/pipeline.yml
```diff
@@ -363,24 +374,36 @@ jobs:
...
               with:
                   context: .
                   file: docker/Dockerfile
                   push: true
                   tags: >
                       ${{ secrets.DOCKER_HUB_USERNAME }}/dfm:latest ,
-                      ${{ secrets.DOCKER_HUB_USERNAME }}/dfm:${{ env.VERSION_NAME }}
+                      ${{ secrets.DOCKER_HUB_USERNAME }}/dfm:${{ env.VERSION_NAME }} ,
+                      ghcr.io/${{ github.repository }}:latest ,
+                      ghcr.io/${{ github.repository }}:${{ env.VERSION_NAME }}
```
hansegucker, line 406: Let's use the the `ghcr.io` in the `docker-compose.yml`.

hansegucker (changes requested):
Thanks, I've only found a small bit.

.github/workflows/pipeline.yml
lukasrad02: Changed in d4d8869b07ee98c39df145770640acd5f8d967f6

hansegucker (approved):

PR #1268, opened 2026-02-10
Rename software to fuesim-digital

hansegucker:
### PR Checklist


fekoch (approved):

docker/authelia-config/configuration.yml
```diff
@@ -82,7 +82,7 @@ identity_providers:
                 - 'http://localhost:4200'
         clients:
             - client_id: 'dfm-backend'
```
christianschaeffer: ` - client_id: 'dfm-backend' `

while we are at it: The shorthand dfm refers to "digital FüSim MANV". It is only used occasionally, but might be changed as well to avoid confusion in the future.

frontend/src/app/shared/components/header/header.component.html
```diff
@@ -1,6 +1,6 @@
 <nav class="navbar bg-body-tertiary mb-4">
     <div class="container container-fluid">
-        <a routerLink="/" class="navbar-brand">Digitale FüSim MANV</a>
+        <a routerLink="/" class="navbar-brand">FüSim Digital</a>
```
christianschaeffer, line 3: Not really a technical proposal: While I understand you refuse to use the point as part of the technical project name, you might want to consider "FüSim.digital" for the title line and text mentions. It might be easier to read and help people memorize the URL better. However, I can live with either and don't have a strong opinion.

docker/authelia-config/configuration.yml
hansegucker: We decided to keep the prefix `DFM_` for environment variables to ease the migration process, but this occurrence is now changed.

.env.example
```diff
@@ -19,7 +19,7 @@ DFM_UPLOAD_LIMIT=200
 
 # persistent data storage path inside the container, e.g. used for SSL
 # if you change this from the default, keep in mind that the path needs to exist inside the container
-DFM_PERSISTENT_DATA_PATH="/dfm-data"
+DFM_PERSISTENT_DATA_PATH="/data"
```
christianschaeffer, line 22: Isn't it useful to have a kind of domain name prefix?

hansegucker: No, as this completely lives within the container.

test-scenarios
lukasrad02, line 1: I guess this change was by accident?

hansegucker: Yes, let's keep this thread open as a hint that I still have to change the strings in the test-scenarios repo.

PR #1280, opened 2026-02-15
Prevent reducing of actions that are not allowed for templates 

### LINKED ISSUE #1277: Template exercise can be started from simulation settings modal
When I'm in an exercise template, the button to start the exercise is not visible in the bottom toolbar. 

However, the simulated regions modal still contains the "start" button and the server still accepts the respective action, so that an exercise template can be started. Since the timer is also hidden, there is no way to notice that the exercise is already running and an exercise that has been started by accident may be running indefinitely (as long as at least one client joined the exercise).

This leads to patient health, transfers and simulated regions already progressing and exercises created from this template will start at whatever time the template has run instead of at 0:00, which may be confusing during exercise evaluation.

hansegucker:
### PR Checklist

### DIFF
--- a/frontend/src/app/pages/exercises/exercise/shared/simulation/trainer-modal/simulated-regions-modal/simulated-regions-modal.component.html
+++ b/frontend/src/app/pages/exercises/exercise/shared/simulation/trainer-modal/simulated-regions-modal/simulated-regions-modal.component.html
@@ -1,6 +1,8 @@
 <div class="modal-header">
     <h4 class="modal-title">Simulationsübersicht</h4>
-    <app-start-pause-button />
+    @if ((isTemplate$ | async) === false) {
+        <app-start-pause-button />
+    }
     <button
         type="button"
         class="btn-close"
--- a/backend/src/database/services/exercise-manager-service.ts
+++ b/backend/src/database/services/exercise-manager-service.ts
@@ -39,7 +39,7 @@ export class ExerciseManagerService {
         if (!exerciseTemplate) {
             throw new ApiError();
         }
-        const newExercise = ExerciseFactory.fromBlank();
+        const newExercise = ExerciseFactory.fromBlankTemplate();
         await exerciseService.createTemplate(newExercise, exerciseTemplate);
         return {
             ...exerciseTemplate,
--- a/backend/src/exercise/exercise-factory.ts
+++ b/backend/src/exercise/exercise-factory.ts
@@ -54,6 +54,17 @@ export class ExerciseFactory {
         );
     }
 
+    public static fromBlankTemplate() {
+        const activeExercise = this.fromBlank();
+        const adaptedStateString = {
+            ...activeExercise.getExercise().initialStateString,
+            isTemplate: true,
+        };
+        activeExercise.getExercise().initialStateString = adaptedStateString;
+        activeExercise.getExercise().currentStateString = adaptedStateString;
+        return activeExercise;
+    }
+
     /**
      * @param file A **valid** import file
      */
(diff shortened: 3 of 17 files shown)


hansegucker:
@lukasrad02 @fekoch @Quixelation I think this is a better approach and ensures that forbidden actions don't reduce.

@lukasrad02 With your knowledge of the whole software, it would be great if you could check the `forbidInTemplates` marks and maybe add them for the `simulation.ts` file if necessary.

frontend/src/app/pages/exercises/exercise/shared/simulation/trainer-modal/simulated-regions-modal/simulated-regions-modal.component.html
```diff
@@ -1,6 +1,8 @@
 <div class="modal-header">
     <h4 class="modal-title">Simulationsübersicht</h4>
-    <app-start-pause-button />
+    @if ((isTemplate$ | async) === false) {
+        <app-start-pause-button />
+    }
```
lukasrad02, line 5: Just an idea: You could also do the check inside the component and maybe instead of hiding the button only disable it and show a tooltip like "Sie bearbeiten eine Vorlage. Um die Übung zu starten, erstellen Sie zuerst eine Übung mit dieser Vorlage" *\[Edit: I just noticed that the toolbar already shows such a message, so it could easily be moved into the button component for central handling\]* or convert the start button into a button that automatically creates a new instance from the template (which would allow for quick testing of the template).

I would recommend to move the check inside the component so that users of the component don't have to handle template mode (the start action is not allowed anyways, but having a button that doesn't work might also be confusing).  
Converting the button into a "quick test" button can also be done in a separate PR (if you consider it generally useful)

lukasrad02 (commented):
In my opinion, it is a bit arbitrary which actions are allowed in a template and which aren't.

For example, why is it not allowed to already send an initial alarm group so that all exercises started from this template have some first vehicles on their way?

On the other hand, given the rather restrictive handling of actions, what is the use case of allowing (un-)loading a vehicle with personnel and even patients? If you want to support building scenarios where participants have to handle an already ongoing situation (e.g. to train higher-level command staff that needs some time to arrive), wouldn't it make sense to allow any action (including starting the exercise, maybe with a big warning that this cannot be undone and will be applied to all exercises started from this template)?

lukasrad02:
> wouldn't it make sense to allow any action (including starting the exercise, maybe with a big warning that this cannot be undone and will be applied to all exercises started from this template)?

One reason that would actually support not allowing to start templates is that started exercises with simulation features are not guaranteed to be migratable and thus, templates could suddenly become unusable if they have been stated once. But this case could also be handled by informing the user about the consequences of starting an exercise

christianschaeffer:
> For example, why is it not allowed to already send an initial alarm group so that all exercises started from this template have some first vehicles on their way?

That would indeed be very useful, especially for parallel exercises.

In addition, there should ideally be an extended "transfer overview" feature in advance to fine-tune arrivals or remove vehicles from transfers. Currently, it is not possible to remove vehicles during transfer, so users might accidentally ruin their exercise template if they add too many then-unremovable transferring vehicles somewhere.



> what is the use case of allowing (un-)loading a vehicle with personnel and even patients? [...] support building scenarios where participants have to handle an already ongoing situation (e.g. to train higher-level command staff that needs some time to arrive), 

Scenarios that start with an ongoing situation are not nice to have, but actually very important! Not only for the case of higher-level command staff arriving later mentioned here, but also for training of specific functions, such as transport organization, which is currently one of the main use cases.

However, I guess it would suffice to only add some initial alarm groups to the current list of permitted actions. That should allow building everything that is currently unsupported.



> wouldn't it make sense to allow any action (including starting the exercise, maybe with a big warning that this cannot be undone and will be applied to all exercises started from this template)?

Well, the exercise timer should start at the actual exercise start, but it can be implemented in many ways.

christianschaeffer:
Wild idea: Instead of "Start", add a "simulate for 1 minute" (*) button so trainers can progress the initial state without the hassle of actually starting the exercise and managing an ongoing timer. This would update patients, vehicle transfers (and technical challenges in the future) while ignoring simulated regions. 

This could also be used to properly evaluate patients' behaviour (which I think has never been done).

(*) Could be extended to allow for multiple step sizes.

PR #1285, opened 2026-02-24
Configure dependabot

hansegucker:
### PR Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,40 @@
+version: 2
+updates:
+    - package-ecosystem: 'npm'
+      directory: '/'
+      schedule:
+          interval: 'daily'
+
+    - package-ecosystem: 'npm'
+      directory: '/backend'
+      schedule:
+          interval: 'daily'
+
+    - package-ecosystem: 'npm'
+      directory: '/benchmark'
+      schedule:
+          interval: 'daily'
+
+    - package-ecosystem: 'npm'
+      directory: '/frontend'
+      schedule:
+          interval: 'daily'
+      groups:
+          angular:
+              applies-to: version-updates
+              patterns:
+                  - '@angular*'
+                  - '@ng-bootstrap*'
+                  - '@ngrx*'
+                  - 'ng-qrcode'
+                  - 'angular-eslint'
+
+    - package-ecosystem: 'npm'
+      directory: '/shared'
+      schedule:
+          interval: 'daily'
+
+    - package-ecosystem: 'github-actions'
+      directory: '/'
+      schedule:
+          interval: 'weekly'


.github/dependabot.yml
```diff
@@ -0,0 +1,40 @@
...
+          interval: 'daily'
+      groups:
+          angular:
+              applies-to: version-updates
+              patterns:
+                  - '@angular*'
+                  - '@ng-bootstrap*'
+                  - '@ngrx*'
+                  - 'ng-qrcode'
+                  - 'angular-eslint'
```
lukasrad02, line 30: I'm not sure whether this is sufficient. Through angular itself or related packages (e.g. `angular-eslint`), we also have peer dependencies on

- `eslint` (which might also affect `eslint-plugin-*` and `@eslint/*` packages)
- `typescript`
- `bootstrap` and `@popperjs/core` (although it seems that they're not enforced)

and maybe some more packages I've missed.

I don't know how exactly Dependabot performs the updates and whether it will manage to resolve those peer dependencies correctly. Maybe, we should keep it as-is and see whether it works well enough, instead of making the group to large.

---

Additionally, we should use the same version of TypeScript and ESLint across all subprojects to avoid inconsistencies. This should theoretically be possible by specifying `directories` instead of a single `directory`, but again, I'm not sure whether it's worth the effort…

.github/dependabot.yml
lukasrad02, line 1: I'm not sure what the default behavior is, but I think we should specify `versioning-strategy: increase` (see [docs](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#versioning-strategy--)).

Assuming the "Default behavior" from the docs refers to the default of the respective package manager, `npm update` does by default not update the version specifiers in the `package.json`, which is not what we want.

.github/dependabot.yml
```diff
@@ -0,0 +1,40 @@
...
+      directory: '/benchmark'
+      schedule:
+          interval: 'daily'
+
+    - package-ecosystem: 'npm'
+      directory: '/frontend'
+      schedule:
+          interval: 'daily'
+      groups:
+          angular:
```
lukasrad02, line 23: For angular, we might want to [ignore](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#ignore--) major upgrades, since they typically require migrations by `ng update`

lukasrad02 (commented):
Considering all my comments, it could make sense to re-structure the file as follows (pseudocode, not valid):

```txt
- npm in [/, /shared, /backend, /benchmark, /frontend]:
  ignore: angular and related, eslint, typescript
- npm in [/, /shared, /backend, /benchmark, /frontend]:
  ignore: major-upgrades
  group: angular and related, eslint, typescript
```

However, I don't know whether this will actually work:

- Does Dependabot correctly resolve peer dependencies (in contrast to `@angular/*`, where all packages should receive a new version at the same time, a new TypeScript version might be released independent from a new angular version)
- Does Dependabot group upgrades across directories (e.g., eslint in all subprojects)
- Do the overlapping rules work correctly (e.g., a minor upgrade of `foo` in `/shared` would be matched by both rules)

fekoch:
Out of scope

Issue #874, opened 2023-04-18
Change UUID generation to preserve exports


Nils1729:
When we change an action reducer in a way that alters the number of pseudorandom values taken from the state, old exports will fail because UUIDs  generated during the application of actions do not align with the original values.
This problem also exists if we adapt behaviors and activities in a way that influences the number of created activities and events.

PR #1329, opened 2026-03-20
Migrate simulated regions to Zod

hansegucker:
### Summary

This migrates the whole simulated region subtree including all behaviors, activities and events to Zod. I did some testing on the simulated regions and everything seems to work as intended – but as I didn't develop the simulation features, some additional testing could be a good idea.

### PR Checklist


shared/src/models/radiogram/status/radiogram-status.ts
```diff
@@ -1,3 +1,5 @@
-export class RadiogramStatus {
-    readonly type!: `${string}RadiogramStatus`;
-}
+import { z } from 'zod';
+
+export const radiogramStatusSchema = z.strictObject({
+    type: z.literal('radiogramStatus'),
+});
```
lukasrad02, line 5: I think you can completely remove this schema. With class-validator, we used it solely to derive the specialized classes from and we had the guarantee that every specialized radiogram statuses type has to conform to `readonly type!: `${string}RadiogramStatus`;`.

However, with

```ts
export const radiogramDoneStatus = z.strictObject({
    ...radiogramStatusSchema.shape,
    type: z.literal('doneRadiogramStatus'),
    publishTime: z.int().nonnegative(),
});
```

you _overwrite_ the `type` property from the base schema and there is no mechanism that enforces the structure of the type string. You can easily create a `fooBar`-typed radiogram status.

---

I guess we'll see this pattern for some more models within this PR. In my opinion, it would also be fine to keep those base schemas to make it easier to add new properties, since the shared base schema already exists. However, I would propose to remove the `type` property from the base schema, since it doesn't serve a purpose and is overridden in every specialized schema.

Whatever you decide for (keeping or removing the base schema), please do it consistently across all schemas in this PR.

shared/src/models/radiogram/utils/transfer-progress-scope.ts
```diff
@@ -1,14 +1,18 @@
...
-export type Scope = 'singleRegion' | 'transportManagement';
 
-export const scopeAllowedValues: AllowedValues<Scope> = {
-    singleRegion: true,
-    transportManagement: true,
-};
+export const transferProgressScopeAllowedValues = [
+    'singleRegion',
+    'transportManagement',
+] as const;
```
lukasrad02: This is not used outside of this file, since all depending models can now refer to the schema instead.

I would propose inlining the array into the definition of the schema below.

shared/src/models/radiogram/utils/transfer-progress-scope.ts
```diff
@@ -1,14 +1,18 @@
-import type { AllowedValues } from '../../../utils/validators/index.js';
+import { z } from 'zod';
 
 /**
  * Defines the scope of the radiogram on the transfer progress.
  * * `singleRegion`: The information is only about the simulated region that sent the radiogram
  * * `transportManagement`: The information is about all simulated regions that are managed
  *   by the transport management behavior of the simulated region that sent the radiogram
  */
```
lukasrad02, line 8: I think this comment should be moved to the type definition

shared/src/models/radiogram/exercise-radiogram.ts
```diff
@@ -1,73 +1,33 @@
...
-            ),
-        },
-    },
-];
+export const exerciseRadiogramSchema = z.discriminatedUnion('type', [
+    materialCountRadiogramSchema,
+    missingTransferConnectionRadiogramSchema,
+    patientCountRadiogramSchema,
+    personnelCountRadiogramSchema,
+    personnelCountRadiogramSchema,
```
lukasrad02: Duplicate schema

shared/src/models/radiogram/radiogram.ts
```diff
@@ -1,11 +1,12 @@
...
-    readonly id!: UUID;
-    readonly type!: `${string}Radiogram`;
-    readonly simulatedRegionId!: UUID;
-    readonly status!: ExerciseRadiogramStatus;
-    readonly informationAvailable!: boolean;
-    readonly informationRequestKey!: string | null;
-}
+export const radiogramSchema = z.strictObject({
+    id: uuidSchema,
+    type: z.literal('radiogram'),
```
lukasrad02: Similar to my comment for the `radiogramStatusSchema`: The `type` is overwritten in all specific radiogram types.

It might still make sense to keep it as some kind of documentation, but then, I would prefer

- a comment instead of actual code  
and
- using the correct specification (i.e., `z.templateLiteral([ z.string(), "Radiogram" ]);`) of the expected format

shared/src/models/radiogram/radiogram-helpers.ts
```diff
@@ -6,7 +6,7 @@ import {
     isUnpublishedRadiogramStatus,
     participantIdOfRadiogramStatus,
     publishTimeOfRadiogramStatus,
-} from './status/radiogram-status-helpers.js';
+} from './status/index.js';
```
lukasrad02: Is there a specific reason why you changed some of the imports to use a more generic path?

In the past, we had issues with recursive imports caused by overly generic imports. This causes hard-to-debug issues that are often not apparent during compile-time.

If there is no specific reason for this change, I'd prefer to keep using the imports from specific files instead of directories/index.js.

P.S.: Back then, there was no reliable tooling to detect recursive imports, maybe this got better meanwhile?

shared/src/models/radiogram/new-patient-data-requested-radiogram.ts
```diff
@@ -1,55 +1,27 @@
...
+    simulatedRegionId: UUID,
+    status: ExerciseRadiogramStatus
+): NewPatientDataRequestedRadiogram {
+    return {
+        id,
+        type: 'newPatientDataRequestedRadiogram',
+        simulatedRegionId,
+        status,
+        informationRequestKey: null,
+        informationAvailable: false,
```
lukasrad02: This does not align with the default value in the class-based model

shared/src/models/radiogram/missing-transfer-connection-radiogram.ts
```diff
@@ -1,60 +1,31 @@
...
+    status: ExerciseRadiogramStatus,
+    targetTransferPointId: UUID
+): MissingTransferConnectionRadiogram {
+    return {
+        id,
+        type: 'missingTransferConnectionRadiogram',
+        simulatedRegionId,
+        status,
+        targetTransferPointId,
+        informationAvailable: false,
```
lukasrad02: This does not align with the default value in the class-based model

lukasrad02 (changes requested):
Some first comments for you to work on, but I'm still far from done with the review

hansegucker:
Just as a reminder for me drop: Drop usages of `extend`

lukasrad02:
Along your IDE config changes, you could also update the `patterns` of `no-restricted-imports` as follows:

```ts
{
    group: ['**/index.js'],
    message: "Don't import directories from shared, only import single files to reduce the likelihood of recursive imports.",
},
```

However, since this currently causes 657 problems (at least on `dev`), it might make more sense to to this in a separate PR.

shared/src/models/simulated-region.ts
```diff
@@ -1,88 +1,48 @@
...
-    public readonly id: UUID = uuid();
-
-    @IsValue('simulatedRegion' as const)
-    public readonly type = 'simulatedRegion';
-
-    /**
-     * top-left position
-     *
-     * @deprecated Do not access directly, use helper methods from models/utils/position/position-helpers(-mutable) instead.
-     */
```
lukasrad02, line 38: This comment (and thus the deprecation warning) is missing in the Zod schema.

You can simply add jsdoc comments in the schema and the documentation will also be visible on the resulting, inferred type.

shared/src/simulation/activities/provide-personnel-from-vehicles.ts
```diff
@@ -1,68 +1,51 @@
...
-    implements SimulationActivityState
-{
-    @IsValue('providePersonnelFromVehicleActivity' as const)
-    public readonly type = 'providePersonnelFromVehicleActivity';
-
-    @IsUUID(4, uuidValidationOptions)
-    public readonly id: UUID;
+export const providePersonnelFromVehiclesActivitySchema = z.strictObject({
+    ...simulationActivityStateSchema.shape,
+    type: z.literal('providePersonnelFromVehiclesActivity'),
```
lukasrad02: There is a typo in the old literal, it's written without "s". If you change it here, keep in mind to write a migration

shared/src/simulation/behaviors/reportable-information.ts
lukasrad02, line 1: Is there a reason why you moved this from `utils.ts` into its own file?

I would prefer to not have any non-behavior-model code in either a shared `utils` file or as separate files in a `utils` directory, since this is how it has been handled so far.

shared/src/simulation/behaviors/manage-patient-transport-to-hospital.ts
```diff
@@ -1,192 +1,154 @@
...
-
-    static readonly create = getCreate(this);
+const vehiclesForPatientsSchema = z.strictObject({
+    type: z.literal('vehiclesForPatients'),
+    red: z.array(z.string()),
+    redIndex: z.number().int().nonnegative(),
+    yellow: z.array(z.string()),
+    yellowIndex: z.number().int().nonnegative(),
+    green: z.array(z.string()),
+    greenIndex: z.number().int().nonnegative(),
```
lukasrad02: Shouldn't `z.int()` (without `.number()`) be sufficient?

shared/src/simulation/behaviors/manage-patient-transport-to-hospital.ts
```diff
@@ -1,192 +1,154 @@
...
+         * Stores the amount of patients expected in some regions.
+         * Regions are NOT removed from this, if they are not managed anymore,
+         * as the behavior should still have its knowledge about how many patients
+         * are in this region. Therefore, to find the amount of patients in managed
+         * regions, one must filter this object to only contain keys that are also
+         * in {@link simulatedRegionsToManage}
+         */
+        patientsExpectedInRegions: z.record(
+            uuidSchema,
+            z.record(patientStatusSchema, z.number())
```
lukasrad02: I think this can be a `z.int().nonnegative()`

shared/src/simulation/behaviors/manage-patient-transport-to-hospital.ts
```diff
@@ -1,192 +1,154 @@
...
+        ),
+        patientsExpectedToStillBeTransportedByRegion: z.array(
+            z.lazy(() => patientsTransportPromiseSchema)
```
lukasrad02: This has to be an array of `patientsTransportPromiseSchema`

shared/src/simulation/behaviors/manage-patient-transport-to-hospital.ts
```diff
@@ -1,192 +1,154 @@
...
+        ),
+        transferredPatientCounts: z.record(patientStatusSchema, z.number()),
```
lukasrad02: This can be `.int().nonnegative()`

shared/src/simulation/behaviors/manage-patient-transport-to-hospital.ts
```diff
@@ -1,192 +1,154 @@
...
+        vehiclesForPatients: vehiclesForPatientsSchema,
+        /**
+         * @deprecated Use {@link updateRequestVehiclesDelay} instead
+         */
+        requestVehiclesDelay: z.number().int().nonnegative(),
+        /**
+         * @deprecated Use {@link updateRequestPatientCountsDelay} instead
+         */
+        requestPatientCountsDelay: z.number().int().nonnegative(),
+        promiseInvalidationInterval: z.number().int().nonnegative(),
```
lukasrad02: `.number()` is not required when using `.int()`

shared/src/simulation/behaviors/request.ts
```diff
@@ -1,85 +1,73 @@
...
      */
-    @IsInt()
-    @Min(0)
-    public readonly requestInterval: number = 1000 * 60 * 5;
-
-    @IsInt()
-    @Min(0)
-    public readonly invalidatePromiseInterval: number = 1000 * 60 * 30;
+    requestInterval: z.number().int().nonnegative(),
+    invalidatePromiseInterval: z.number().int().nonnegative(),
```
lukasrad02: `.number()` is not needed

shared/src/simulation/events/patients-counted.ts
```diff
@@ -1,27 +1,23 @@
...
+import type { ResourceDescription } from '../../models/utils/resource-description.js';
+import { simulationEventSchema } from './simulation-event.js';
 
-export class PatientsCountedEvent implements SimulationEvent {
-    @IsValue('patientsCountedEvent')
-    readonly type = 'patientsCountedEvent';
+export const patientsCountedEventSchema = z.strictObject({
+    ...simulationEventSchema.shape,
+    type: z.literal('patientsCountedEvent'),
+    patientCount: z.record(patientStatusSchema, z.number()),
```
lukasrad02:
```suggestion
    patientCount: z.record(patientStatusSchema, z.int().nonnegative()),
```

shared/src/simulation/utils/patients-transported-promise.ts
```diff
@@ -1,36 +1,26 @@
...
+import { z } from 'zod';
+import { type UUID, uuidSchema } from '../../utils/uuid.js';
 
-export class PatientsTransportPromise {
-    @IsValue('patientsTransportPromise')
-    readonly type = 'patientsTransportPromise';
+export const patientsTransportPromiseSchema = z.strictObject({
+    type: z.literal('patientsTransportPromise'),
+    promisedTime: z.number().int().nonnegative(),
+    patientCount: z.number().int().nonnegative(),
```
lukasrad02:
```suggestion
    promisedTime: z.int().nonnegative(),
    patientCount: z.int().nonnegative(),
```

shared/src/simulation/utils/resource-promise.ts
```diff
@@ -1,28 +1,24 @@
...
+    type VehicleResource,
+    vehicleResourceSchema,
+} from '../../models/utils/rescue-resource.js';
 
-export class ResourcePromise {
-    @IsValue('resourcePromise')
-    readonly type = 'resourcePromise';
+export const resourcePromiseSchema = z.strictObject({
+    type: z.literal('resourcePromise'),
+    promisedTime: z.number().int().nonnegative(),
```
lukasrad02:
```suggestion
    promisedTime: z.int().nonnegative(),
```

lukasrad02 (changes requested):

lukasrad02 (commented):
Changes look good and I didn't spot any issues during testing either.

Please fix the failing migrations so we can merge

lukasrad02:
> Please fix the failing migrations so we can merge

From looking at the names of the failing scenarios, I consider it likely that the scenarios are broken in the sense of #874, so there is nothing you can fix in this PR. So you may simply remove them.

lukasrad02 (approved):
Seems like there is an issue with OIDC in the Cypress tests, but apart from that, this PR looks good

PR #1342, opened 2026-03-25
Prefer function declarations

hansegucker:
### Summary

We really prefer them.

### PR Checklist

### DIFF
--- a/.eslint/typescript.eslintrc.mjs
+++ b/.eslint/typescript.eslintrc.mjs
@@ -1,3 +1,5 @@
+import esLintAntfuPlugin from 'eslint-plugin-antfu';
+
 export default {
     files: ['**/*.ts', '**/*.js'],
 
@@ -6,6 +8,10 @@ export default {
         '**/migrations/**/*.js',
         'jest.config.js',
     ],
+
+    plugins: {
+        antfu: esLintAntfuPlugin,
+    },
 
     rules: {
         /**
@@ -410,5 +416,6 @@ export default {
         '@typescript-eslint/no-misused-new': 'warn',
         '@typescript-eslint/no-non-null-assertion': 'off',
         '@typescript-eslint/no-unnecessary-type-assertion': 'warn',
+        'antfu/top-level-function': 'error',
     },
 };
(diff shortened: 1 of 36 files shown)


fekoch:
> We really prefer them.

@Quixelation and I conferred -- Actually, we really don't prefer them...

We really think the expression syntax should be used inside other functions/methods.

If function declarations are forced, then that should be only the case at the top level

Two possible solutions for that:

 - Using `'func-style': ['warn', 'expression', { overrides: { namedExports: 'declaration' } },],` (this would not apply for non-exported top level functions)
 - Using [eslint-plugin-antfu](https://www.npmjs.com/package/eslint-plugin-antfu), which provides `'antfu/top-level-function'`. (external eslint plugin, also used by [nextcloud](https://github.com/nextcloud-libraries/eslint-config/issues/1003))

(sorry, that you already changed 40 files)

shared/src/utils/uuid.ts
```diff
@@ -12,13 +12,9 @@ export function uuid(): UUID {
...
 
-const uuidValidationFailedWithId = (id: string) => `Got malformed id: '${id}'.`;
-const uuidValidationFailedMessage: (
-    validationArguments: ValidationArguments
-) => string = (validationArguments: ValidationArguments) =>
-    uuidValidationFailedWithId(String(validationArguments.value));
 export const uuidValidationOptions: ValidationOptions = {
-    message: uuidValidationFailedMessage,
+    message: (validationArguments: ValidationArguments) =>
+        `Got malformed id: '${validationArguments.value}`,
```
fekoch:
```suggestion
        `Got malformed id: '${validationArguments.value}.`,
```

ClFeSc: Just saw this while skimming my mails: In addition the full stop, maybe also add a closing single quote `'`? There is one before the value and then none afterwards.

Btw: The commit with the full stop (2dee1890fae82bef631a6121fb5e0f51499e648b) appears to be intended for squashing? I only know those `fixup!` commits from using `git commit --fixup` and then squashing with `git rebase --autosquash`.

fekoch: Good catch, I reverted the changes there anyways.

Yeah, they are intended for squashing. As I was not in direct contact with @hansegucker, I did not want to rebase his branch :sweat_smile:

Jogius (approved):
eslint plugin is a little sus but seems widely used so it should be fine.

PR #1343, opened 2026-03-26
Add Marketplace Backend

### LINKED ISSUE #1240: creating new Collections
As a user, I want to be able to create sets to group and manage exercise-objects.

**Acceptance criteria**
- [ ] There is a button that takes the user to a form for creating a set.
- [ ] The user can specify a name and description for the set.

### LINKED ISSUE #1243: Viewing Collections
As a user, I want to be able to see an overview of the sets I have created, as well as the sets belonging to my organization, so that I can manage and use them.

**Acceptance criteria**
- [ ] There is a list that displays all of the user's sets and the *public* sets belonging to their organization.
- [ ] Each set in the list displays the name, visibility (public/private) as a symbol, and the number of exercise objects it contains.

### LINKED ISSUE #1245: Creating Collections-Elements
As a user, I want to be able to create exercise objects in sets so that I can use them later in exercises.

**Acceptance criteria**
- [ ] There is a button that takes the user to a form for creating an exercise object.
- [ ] The user can select the type of exercise object they want to create (via a drop-down menu next to the “Create” button).
- [ ] The user can specify the properties of the exercise object in the form.
- [ ] The created exercise object is saved in the selected set.

### LINKED ISSUE #1248: Viewing Collection-Elements
As a user, I want to be able to see the exercise objects in a set so that I can manage and use them.

**Acceptance criteria**
- [ ] The set-detail view displays a list of all exercise objects contained in the set.
- [ ] Each exercise object in the list displays the type and key properties of the exercise object.
- [ ] The user can click on an exercise object to view or edit its details.

### LINKED ISSUE #1252: Editing Collection-Elements
As a user, I would like to edit my saved exercise objects so that I can adjust them over time.

**Acceptance criteria**

- [ ] There is a button on exercise objects that takes the user to an editing form.
- [ ] Editing exercise objects results in the creation of a new version of the containing set, which contains this new version of the exercise object.

```mermaid
stateDiagram
  direction LR
  state s3 <<fork>>
  state s6 <<join>>
  [*] --> Still
  Still --> s3
  s3 --> s1
  s3 --> s4
  s4 --> s5
  s2 --> s6
  s5 --> s6
  s6 --> s7
  s1 --> s8
  s8 --> s2
  s7 --> [*]
  Still:ExerciseElement Bearbeitung beendet
  s1:Kopie des ursprünglichen ExerciseElement erstellen
  s4:Kopie des enthaltenden Sets erstellen
  s5:Versionsnummer der Set-Kopie hochstufen
  s2:Versionsnummer der Element-Kopie hochstufen
  s7:Element-Kopie in der Set-Kopie überschreiben
  s8:Änderungen auf Element-Kopie anwenden
```

Quixelation:
**THIS UPDATES A ZOD VERSION IN SHARED**

### Summary

This PR:
 - includes the backend of the marketplace
 - the ExerciseState change from specific template attributes to a singular templates attribute

### PR Checklist

- [ ] If this PR adds or changes features or fixes bugs, this has been added to the changelog


backend/src/sse.ts
```diff
@@ -0,0 +1,47 @@
+import type {
+    Request as ExpressRequest,
+    Response as ExpressResponse,
+} from 'express';
+import { interval, Subject, takeUntil } from 'rxjs';
+
+export class SSE {
```
Jogius: Is there a technical reason to use SSE here instead of the existing websocket connection?
I'm not very well-versed in SSE, but from my perspective the main advantage over websockets is not having to configure the protocol-change, which we are already doing anyway.

backend/src/collections/collection-event-sender.ts
```diff
@@ -0,0 +1,176 @@
...
+        private readonly userId: string
+    ) {
+        this.sse = new SSE(req, res);
+        this.init();
+    }
+
+    private abort(error: string) {
+        console.error(error);
+        this.sse.sendEvent('error', { message: error });
+        this.sse.close();
```
Jogius: Is abort always a hard failure (i.e. the application crashes), so we want to close the SSE connection?

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1077 @@
...
+
+    public async getOrCreateJoinCode(collectionEntitiyId: CollectionEntityId) {
+        const existingCode = await this.getJoinCode(collectionEntitiyId);
+        if (existingCode && existingCode.expiresAt > new Date()) {
+            return existingCode;
+        }
+        const newCode = crypto
+            .randomBytes(8)
+            .toString('base64url')
+            .replaceAll(' ', '-');
```
Jogius: Do we have a common abstraction to create join codes (e.g. organizations, collections etc.?). If not, I think we should aim to create one.

Jogius: Also, I think `replaceAll(' ', ...)` is a no-op here, since base64url does not include spaces, right?

shared/src/marketplace/models/collection-elements.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const collectionElementsDtoSchema = z.strictObject({
+    /**
+     * Elements directly included in the collection
+     */
+    direct: z.array(elementDtoSchema),
+
+    /**
+     * Elements included in the collection via imports.
+     * This only inlcudes import-levels visible to the user
```
Jogius: `includes`

shared/src/marketplace/models/collection-elements.ts
```diff
@@ -0,0 +1,55 @@
...
+    imported: z.array(collectionElementsSingleSchema),
+
+    /**
+     * Elements being used by collection elements,
+     * but not directly visible to the user
+     * (e.g. elements being used in elements of collections in collections)
+     */
+    references: z.array(collectionElementsSingleSchema),
+});
+
```
Jogius: (Not directly related to the code, but came up here)
Do you have some drawing/concept of the current model? That would help a lot in reviewing this in-depth.

shared/src/marketplace/models/collection-elements.ts
```diff
@@ -0,0 +1,55 @@
...
+        allElements(): ElementDto[] {
+            return [
+                ...this.allVisibleElements(),
+                ...elements.references.flatMap(
+                    (reference) => reference.elements
+                ),
+            ];
+        },
+    };
+}
```
Jogius: Why are you returning functions here instead of directly returning the objects? Do we need the new instance of the arrays every time or is this just code style? I personally don't like this much, just legibility-wise.

shared/src/marketplace/models/collection-relationship.ts
```diff
@@ -0,0 +1,54 @@
...
+    role: collectionRelationshipTypeSchema,
+});
+export type CollectionRelationshipDto = z.infer<
+    typeof collectionRelationshipDtoSchema
+>;
+
+export const collectionRelationshipTypesDisplayNames: {
+    [key in (typeof collectionRelationshipTypeAllowedValues)[number]]: string;
+} = {
+    admin: 'Besitzer',
```
Jogius: Is this the owner or an admin (semantically, I would say there can only be a single owner and multiple admins, but there can only be a single Besitzer).

shared/src/marketplace/models/versioned-element-names.ts
```diff
@@ -0,0 +1,9 @@
+export function getVersionedElementTypeDisplayName(
+    // We cannot check if all VersionedElements are included, bc circular dependencies are hellish
+    versionedElementType: string
+): { singular: string; plural: string } | undefined {
+    return {
+        vehicleTemplate: { singular: 'Fahrzeug', plural: 'Fahrzeuge' },
+        alarmGroup: { singular: 'Alarmgruppe', plural: 'Alarmgruppen' },
+    }[versionedElementType];
+}
```
Jogius: I think we could get better typing by just directly exporting this object/map and indexing it directly when needed

shared/src/marketplace/models/versioned-element-names.ts
```diff
@@ -0,0 +1,9 @@
+export function getVersionedElementTypeDisplayName(
+    // We cannot check if all VersionedElements are included, bc circular dependencies are hellish
```
Jogius: Can you elaborate?

shared/src/state-migrations/22-remove-illegal-vehicle-movement-actions.ts
```diff
@@ -19,7 +19,7 @@ export const removeIllegalVehicleMovementActions22: Migration = {
                     intermediaryState,
                     'vehicle',
                     vehicleId
-                );
+                ) as WritableDraft<any>;
```
Jogius: Why can we just modify an old migration here, and why is this necessary?

shared/src/store/action-reducers/alarm-group.ts
```diff
@@ -167,6 +168,12 @@ export namespace AlarmGroupActionReducers {
                     'alarmGroup',
                     alarmGroupId
                 );
+
+                if (!isElementVersionId(alarmGroupId)) {
+                    throw new ReducerError(
+                        `Invalid alarmGroupVehicleId ${alarmGroupVehicleId} for AlarmGroup with id ${alarmGroup.id}`
```
Jogius: Why is the `alarmGroupVehicleId` invalid here when `alarmGroupId` is not an `elementVersionId`?

shared/src/store/action-reducers/vehicle-templates.ts
```diff
@@ -56,6 +60,12 @@ export class EditVehicleTemplateAction implements Action {
     public readonly personnelTemplateIds!: readonly UUID[];
 }
 
+/**
+ * @deprectated Use `[VehicleTemplate] Delete vehicleTemplate by entityId`
```
Jogius: `deprecated`

shared/src/store/action-reducers/vehicle-templates.ts
```diff
@@ -110,6 +129,12 @@ export namespace VehicleTemplateActionReducers {
             rights: 'trainer',
         };
 
+    /**
+     * @deprectated Use `[VehicleTemplate] Delete vehicleTemplate by entityId`
```
Jogius: `deprecated`

shared/src/store/action-reducers/vehicle-templates.ts
```diff
@@ -130,6 +155,35 @@ export namespace VehicleTemplateActionReducers {
...
+                )) {
+                    alarmGroup.alarmGroupVehicles = Object.fromEntries(
+                        Object.entries(alarmGroup.alarmGroupVehicles).filter(
+                            ([_, vehicle]) =>
+                                vehicle.vehicleTemplateId !== vehicleTemplate.id
+                        )
+                    );
+                }
+                return draftState;
+            },
```
Jogius: This is duplicate code, we should either call the old reducer function or extract the code into a common util function.

Jogius (commented):
Partial review of `shared` and `backend`

shared/src/marketplace/models/collection-elements.ts
Quixelation: i kinda like the aspect that it (hopefully?) doesnt need to compute the others if they are not needed.
and i am personally not a big fan of having multiple properties with duplicate data.

shared/src/marketplace/models/collection-relationship.ts
Quixelation: it represents an admin - since "admin" is the highest permission level.
it was requested in a UI Feedback Round that "Admin" be changed to "Besitzer" for better user undertstanding.

we could instead use "Verwalter"?

backend/src/database/repositories/collection-repository.ts
Quixelation: > Do we have a common abstraction to create join codes (e.g. organizations, collections etc.?). If not, I think we should aim to create one.

i totally agree - i want to wait for organisations in this case, to refactor my implementation based on that

backend/src/database/default-data/collection-default-data.ts
```diff
@@ -0,0 +1,59 @@
...
+    title: 'FüSim MANV 2025 Übungselemente',
+    description: '',
+    archived: false,
+    createdAt: new Date(2022, 6, 27), // 27.07.2022 (date where, notsan was added e.g.)
+    editedAt: new Date(2023, 6, 10), // 10.07.2023 (date where, nef-vehicle was edited e.g.)
+    draftState: false,
+    elementCount: 0,
+    version: 1,
+    visibility: 'embedded',
+    elements: Object.values(defaultVehicleTemplatesById).map((template) => ({
```
fekoch, line 44: Should `defaultVehicleTemplatesById` be moved also into this file here in the future?

backend/src/database/schema.ts
```diff
@@ -22,19 +28,24 @@ import {
...
-function typedUUID<T>() {
+function typedUUID<T = string>() {
     return uuid().$type<T>();
 }
+function defaultUUID<T = string>() {
+    return typedUUID<T>().default(sql`uuid_generate_v4()`);
+}
+function defaultPrefixedUUID(prefix: string) {
+    return varchar().$defaultFn(() => `${prefix}_${crypto.randomUUID()}`);
+}
```
fekoch, line 55: It seems surprising to me, that we arbitrarily use the database built-in or node's function for generating uuids. What are your thoughts on using the same mechanism for both?

backend/src/database/schema.ts
```diff
@@ -151,6 +162,140 @@ export const actionTable = pgTable(
...
+    };
+}
+
+export const collectionTable = pgTable(
+    'collections',
+    {
+        ...stateVersionedEntity<CollectionEntityId, CollectionVersionId>('set'),
+        title: varchar().notNull(),
+        description: varchar().notNull(),
+        visibility: varchar().notNull().default('private'),
```
fekoch: This seems like an enum and not a varchar to me?

backend/src/database/schema.ts
```diff
@@ -151,6 +162,140 @@ export const actionTable = pgTable(
...
+        draftState: boolean().notNull(),
+        archived: boolean().notNull().default(false),
+        // This is just easier for everybody
+        elementCount: integer().notNull().default(0),
```
fekoch: I tend to disagree, this seems like a good way to forget updating the count at some point.

I would prefer just counting the corresponding entries, when we select a collection or, if we really want to add this into the data model, update the values using triggers...

Quixelation: I totally agree. I plan to abolish that attribute.

Quixelation: the problem was that we have quite the calculation, but I think postgres should be quite fine joining 2-3 tables.

backend/src/database/schema.ts
```diff
@@ -151,6 +162,140 @@ export const actionTable = pgTable(
...
+    },
+    (table) => [
+        unique('unique_set_version').on(table.entityId, table.version),
+        unique('unique_set_id').on(table.entityId, table.versionId),
```
fekoch: I feel like you need to explain to me (and probably document somewhere) the different versions and ids.

Some questions I have:
- Is `entityId` + `version` the natural primary key of a collection? Why did you decide on `versionId` for the pk?
- If `versionId` is the primary key on `collections`, why do we need an additional UNIQUE constraint using it?

backend/src/database/schema.ts
lukasrad02: If you decide to let the JS code generate the UUID, I'd suggest to use the `uuid()` function from shared for consistency (also used in shared/frontend):

<https://github.com/hpi-sam/fuesim-digital/blob/049d2909fdb7b5a45228b77334f9a6dcb9ca951b/shared/src/utils/uuid.ts#L8-L11>

backend/src/database/schema.ts
Quixelation: > Is entityId + version the natural primary key of a collection? Why did you decide on versionId for the pk?

because handling a singular pk is much easier - especially when debugging.
i can just filter by versionId for example (scan with my eyes) instead of having to first filter by entityId and then having to search for my version. i am generally personally not a fan of primary keys larger than 1 column for non-mapping-tables...

> If versionId is the primary key on collections, why do we need an additional UNIQUE constraint using it?

i will remove that

shared/src/marketplace/models/versioned-elements.ts
```diff
@@ -0,0 +1,21 @@
+import * as z from 'zod';
+import type { Immutable } from 'immer';
```
fekoch, line 2: unused import

shared/src/marketplace/models/versioned-element-model.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const versionedElementModelSchema = z.strictObject({
+    entity: z.object({
+        entityId: elementEntityIdSchema,
+        versionId: elementVersionIdSchema,
+        type: collectionElementTypeSchema,
+    }),
+});
+
+export type VersionedElementModel = z.infer<typeof versionedElementModelSchema>;
```
fekoch, line 22: Can you explain this type to me?

backend/src/database/schema.ts
```diff
@@ -151,6 +163,180 @@ export const actionTable = pgTable(
...
+
+export const collectionUserMappingTable = pgTable(
+    'collection_user_mapping',
+    {
+        id: defaultUUID().primaryKey().notNull(),
+        collection: varchar().notNull().$type<CollectionEntityId>(),
+        userId: varchar()
+            .notNull()
+            .references(() => userTable.id, { onDelete: 'cascade' }),
+        role: varchar().notNull().$type<CollectionRelationshipType>(),
```
fekoch: enum?

backend/src/database/schema.ts
```diff
@@ -151,6 +163,180 @@ export const actionTable = pgTable(
...
+        unique('unique_template_version').on(table.entityId, table.version),
+        unique('unique_template_id').on(table.entityId, table.versionId),
+    ]
+);
+
+export const collectionUserMappingTable = pgTable(
+    'collection_user_mapping',
+    {
+        id: defaultUUID().primaryKey().notNull(),
+        collection: varchar().notNull().$type<CollectionEntityId>(),
```
fekoch, line 417: foreign key?

backend/src/database/schema.ts
```diff
@@ -151,6 +163,180 @@ export const actionTable = pgTable(
...
+        userId: varchar()
+            .notNull()
+            .references(() => userTable.id, { onDelete: 'cascade' }),
+        role: varchar().notNull().$type<CollectionRelationshipType>(),
+    },
+    (table) => [
+        unique('unique_collection_user').on(table.userId, table.collection),
+    ]
+);
```
fekoch, line 427: Here, I would really argue for using the natural primary key collection(id) + userId

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1407 @@
...
+            await this.databaseConnection
+                .select()
+                .from(collectionJoinCodesTable)
+                .where(
+                    eq(collectionJoinCodesTable.collection, collectionEntitiyId)
+                )
+        );
+    }
+
+    public async getOrCreateJoinCode(collectionEntitiyId: CollectionEntityId) {
```
fekoch:
```suggestion
    public async getOrCreateJoinCode(collectionEntityId: CollectionEntityId) {
```

backend/src/database/schema.ts
Quixelation: not possible since it's not a unique column

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1407 @@
...
+                        and(
+                            eq(collectionJoinCodesTable.code, code),
+                            gt(collectionJoinCodesTable.expiresAt, new Date())
+                        )
+                    )
+            )?.collection ?? null
+        );
+    }
+
+    public async getUserRoleInCollection(
```
fekoch: I don't like that this method is this complicated for such a simple operation. At least one part should be simple `SELECT role FROM collectionUserMappings WHERE userId = ${userId} AND collection = ${collectionEntityId};`-query.

For the `other`-role: It seems, like it should be also just `SELECT visibility FROM latest_collections WHERE entityId = ${collectionEntityId};`. (Or `SELECT visibility FROM collections WHERE entityId = ${collectionEntityId} ORDER BY version DESCENDING LIMIT 1`, if we do not want to introduce a view for `latest_collections`)

Maybe my point is mostly, that the `latestCollectionVersionNumbers` seems like a very cumbersome abstraction for an operation (just querying the latest collectionVersions) that should be straightforward.

shared/src/marketplace/models/versioned-element-model.ts
Quixelation: its supposed to serve a the base for elements (in the exercise) which have references to the marketplace (having a TemplateVersion with versionId and entityId). those elements can then have the attribute `entity` (optional, since we want to support exercise elements that are not tied to the marketplace - e.g. a vehicle from an old exercise import)

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1407 @@
...
+                .leftJoin(
+                    elementTable,
+                    eq(
+                        elementCollectionMappingTable.elementVersionId,
+                        elementTable.versionId
+                    )
+                )
+                .groupBy(collectionTable.versionId)
+        );
+    }
```
fekoch, line 350: You do not need joins for this query. The query can directly run on `elementCollectionMappingTable`

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1407 @@
...
+
+    private latestCollections(opts: { allowDraftState?: boolean }) {
+        const latestCollectionVersionNumbers =
+            this.latestCollectionVersionNumbers({
+                allowDraftState: opts.allowDraftState,
+            });
+
+        return this.databaseConnection.$with('latestCollectionsView').as(
```
fekoch: We should not name non-views View, as this could lead to confusion. 

Also, we should probably have a view like this as an actual database view.

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1407 @@
...
+                            eq(
+                                elementTable.version,
+                                latestElementVersionNumbers.latestversion
+                            )
+                        )
+                    )
+            );
+    }
+
+    public async getOrCreateDraftState(
```
fekoch: Maybe rename to `getOrCreateDraftCollectionVersion`? `draftState` is a very generic name and has a high overlap with multiple other parts of the codebase

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1407 @@
...
+    ): Promise<[CollectionVersion, boolean]> {
+        const result = this.onlySingle(
+            await this.databaseConnection
+                .select()
+                .from(collectionTable)
+                .where(
+                    and(
+                        eq(collectionTable.entityId, collectionEntityId),
+                        eq(collectionTable.draftState, true)
+                    )
```
fekoch, line 416: You assume that only ever one collectionVersion of a collectionEntity is in `draftState=true`. Do you think we should try to enforce this constriant (using triggers) in the database?

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1407 @@
...
+                )
+        );
+
+        if (result === null) {
+            const latestVersion = this.strict(
+                await this.getLatestCollectionByEntityId(collectionEntityId)
```
fekoch, line 422: Can't we assume that, if there is a collectionVersion with `draftState=true`, that version must be that "latest" one? If so, we could directly query for the latest version and check if it has `draftState` set or not

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1407 @@
...
+            await this.copyReferencesBetweenCollections(
+                latestVersion.versionId,
+                newCollection.versionId
+            );
+
+            await this.copyDependenciesBetweenCollections({
+                sourceVersion: latestVersion.versionId,
+                targetVersion: newCollection.versionId,
+            });
+
```
fekoch, line 447: This seems like an atomic operation to me. Are there cases, when one would need the `copyReferencesBetweenCollections()` and `copyDependenciesBetweenCollections()` functions separately? If not, maybe a single function, (maybe even in the database), `createNewCollectionVersionOf()` would be the cleaner call

fekoch (dismissed):
[wip] I am still working through the collection-repository. Two general poitns:

- there are some leftovers from the data type renamings
- I was thinking: maybe it would be helpful for the readability of this (and overall every) repository, if we always annotate the repository function with types from the business logic (or variation of those). The auto-inferred types get pretty large and confusing, and named types communicate the intent more clearly

backend/src/database/repositories/collection-repository.ts
Quixelation: Views in drizzle are still very awful DX since they can generate non-working migrations (statements in the wrong order) - thats why i switched away from views.

backend/src/database/repositories/collection-repository.ts
Quixelation: i disagree since they serve different purposes.
copyDependencies[...] copies the entries in the dependencies-table while copyReferences[...] copies the mappings in the element_collection_mappings table with isBaseReference=false iirc.

copyReferences basically only copies over references to the new collection version, while copyElements also creates a copy of the elements in the element table.

we have to main operations:
- Creating a new Version of a Collection:
    - copyReferences
    - copyDependencies
- Creating a whole duplicate of a Collection:
    - copyElements
    - copyDependencies

fekoch: Great, convinced :+1:

backend/src/database/default-data/collection-default-data.ts
```diff
@@ -0,0 +1,58 @@
...
+
+function newElementEntityId(uuid: string) {
+    return elementEntityIdSchema.parse(`element_entity_${uuid}`);
+}
+
+// WARNING:: DO NOT CHANGE THESE VALUES UNLESS ABSOLUTELY NECESSARY,
+// OTHERWISE YOU MIGHT BREAK EXISTING EXERCISES THAT RELY ON THESE DEFAULT ELEMENTS.
+// IF YOU WANT TO CHANGE THESE ELEMENTS,
+// PLEASE CREATE A NEW SET OF DEFAULT ELEMENTS (e.g.  FüSim Digital {YEAR} Übungselemente)
+const fuesimManvDefaultCollectionData: DefaultCollection = {
```
Jogius: If the recommended way to add new default items is to add a new collection, the variable name should mirror the collection title imo (i.e. contain the year 2025 in some manner).

backend/src/database/repositories/collection-repository.ts
```diff
@@ -0,0 +1,1406 @@
...
+                )
+            );
+        });
+
+        console.log(
+            '[CollectionRepository]: Default collection data has been set'
+        );
+    }
+
+    public async getJoinCode(collectionEntitiyId: CollectionEntityId) {
```
Jogius:
```suggestion
    public async getJoinCode(collectionEntityId: CollectionEntityId) {
```

shared/package-lock.json
```diff
@@ -9051,9 +9051,10 @@
             }
         },
         "node_modules/zod": {
-            "version": "4.1.13",
-            "resolved": "https://registry.npmjs.org/zod/-/zod-4.1.13.tgz",
-            "integrity": "sha512-AvvthqfqrAhNH9dnfmrfKzX5upOdjUVJYFqNSlkmGf64gRaTzlPwz99IHYnVs28qYAybvAlBV+H7pn0saFY4Ig==",
+            "version": "4.3.6",
+            "resolved": "https://registry.npmjs.org/zod/-/zod-4.3.6.tgz",
+            "integrity": "sha512-rftlrkhHZOcjDwkGlnUtZZkvaPHCsDATp4pGpuOOMDaTdDDXF91wuVDJoWoPsKX/3YPQ5fHuF3STjcYyKr+Qhg==",
+            "license": "MIT",
```
Jogius, line 9245: Potentially unrelated?

shared/src/state.ts
```diff
@@ -256,6 +234,10 @@ export class ExerciseState {
      * **Important**
      *
      * This number MUST be increased every time a change to any object (that is part of the state or the state itself) is made in a way that there may be states valid before that are no longer valid.
+     *
+     * WARNING: Before incresing this number, make sure to check:
```
Jogius:
```suggestion
     * WARNING: Before increasing this number, make sure to check:
```

shared/src/http-interfaces.ts
```diff
@@ -210,3 +223,408 @@ export const updateParallelExerciseInstancesSchema = z.object({
...
+    public readonly Response!: TResponse extends z.ZodType
+        ? ImmutableInfer<TResponse>
+        : never;
+}
+
+/* eslint-disable @typescript-eslint/naming-convention */
+
+export namespace Marketplace {
+    // eslint-disable-next-line @typescript-eslint/no-shadow
+    export namespace Element {
```
Jogius, line 28: This concept of nested namespaces and type definitions for http interfaces seems to be new, and only used here for now -- is it in scope to migrate other http routes to this system too?

shared/src/http-interfaces.ts
```diff
@@ -210,3 +223,408 @@ export const updateParallelExerciseInstancesSchema = z.object({
...
+    public readonly responseSchema: TResponse;
+    public readonly Request!: TRequest extends z.ZodType
+        ? ImmutableInfer<TRequest>
+        : never;
+    public readonly Response!: TResponse extends z.ZodType
+        ? ImmutableInfer<TResponse>
+        : never;
+}
+
+/* eslint-disable @typescript-eslint/naming-convention */
```
Jogius, line 25: I _personally_ like this style, and also the naming, but it feels weird to go against our linters conventions so knowingly. I think this should be discussed.

shared/src/utils/infer.ts
```diff
@@ -0,0 +1,4 @@
+import type { Immutable } from 'immer';
+import type { z } from 'zod';
+
+export type ImmutableInfer<T> = Immutable<z.infer<T>>;
```
Jogius: If we introduce this, we should migrate other instances of `Immutable<z.infer<T>>` to use `ImmutableInfer<T>` too (e.g. `RestrictedZone`, `Vehicle`)

shared/src/utils/hybrid-id.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from 'zod';
+import { elementVersionIdSchema } from '../marketplace/models/versioned-id-schema.js';
+import { uuidSchema } from './uuid.js';
+
+export const hybridIdSchema = z.union([uuidSchema, elementVersionIdSchema]);
+
+export type HybridId = z.infer<typeof hybridIdSchema>;
```
Jogius: Can we come up with a more semantic name for this? `HybridId` could mean anything

shared/src/models/template-helpers.ts
```diff
@@ -0,0 +1,15 @@
+import type { ExerciseState } from '../state.js';
+import type { Template } from './template.js';
+
+export function getTemplates<T extends Template['type']>(
+    draftState: ExerciseState,
+    templateId: T
```
Jogius: I think this isn't the `templateId` but the `templateType`/`type`

shared/src/store/action-reducers/exercise.ts
```diff
@@ -204,10 +205,16 @@ export namespace ExerciseActionReducers {
...
             if (mutablePartialExport.mapImageTemplates !== undefined) {
                 if (mode !== 'append') {
-                    draftState.mapImageTemplates = {};
+                    // Remove all mapImageTemplates
+                    draftState.templates = Object.fromEntries(
+                        Object.entries(draftState.templates).filter(
+                            ([_, template]) =>
+                                template.type !== 'mapImageTemplate'
+                        )
+                    );
```
Jogius: I've seen this `Object.fromEntries(Object.entries(...))` pattern enough times to propose we build a helper method for this (maybe even explicitly with `filter`).

shared/src/state-migrations/migration-functions.ts
```diff
@@ -74,6 +72,7 @@ type MigrateStateFunction = (state: object) => void;
 export interface Migration {
     action: MigrateActionFunction | null;
     state: MigrateStateFunction | null;
+    unmigratableActions?: true;
```
Jogius: Should this not be `false` by default?

Jogius (commented):
Another partial review (bottom up this time)

shared/package-lock.json
Quixelation: no, i think i made that change because our zod versions between shared/frontend/backend were out of sync causing issues with typings

backend/src/database/default-data/collection-default-data.ts
Quixelation: i chose that name because it still had Füsim **MANV**. To represent, that these were the old elements from the MANV Version of the software. 
it not really that up to date anymore since basically everyone knows about the rename now and associates the elements also with the new Füsim Digital name.

So would be open to just renaming the default collection to "Füsim Digital 2025/26" or smth

shared/src/http-interfaces.ts
Quixelation: while i am a fan of the nested namespaces *especially* because of the IDE autocompletions.
**BUT** I dont think we need to strictly apply this pattern to the rest of the software and migrate the entire http routes to this system (especially not in this PR). imo (maybe thats also a team discussion again), it should be acceptable to just have this new portion with namespace while we keep the rest as-is.

shared/src/store/action-reducers/exercise.ts
Quixelation: i would totally agree. but not in-scrope for this PR

shared/src/state-migrations/migration-functions.ts
Quixelation: this is not the default value, just the type: it can be either undefined or true.
my thought process behind that was that we dont need to specify that we can successfully migrate our actions in every migration and as such only make this field valid if we specify that we *can not* migrate.
this would make migrations with unmigratable actions more obvious.

but i also see how this can be an anti-pattern, so if you're of the opinion that we should just make it `boolean` i would also be on board

backend/src/database/default-data/collection-default-data.ts
Quixelation: we could but i dont technically see it in scope for this PR since we probably still have _some_ tangled code hanging onto this.

backend/src/database/repositories/collection-repository.ts
Quixelation: will be removed

shared/src/marketplace/models/collection-elements.ts
Quixelation: refactor to single exported functions

shared/src/marketplace/models/collection-relationship.ts
Quixelation: Verwalternde Orga vs Betrachtende Orgs

shared/src/http-interfaces.ts
Quixelation: new Route() goes into `utils`
- AGENDA: nested namespaces + code style

shared/src/utils/infer.ts
Quixelation: we remove this type

shared/src/utils/hybrid-id.ts
Quixelation: remove if possible

Quixelation: `elementOrTemplateId`

backend/src/database/schema.ts
Quixelation: outdated since replaced with organsiation owner system

backend/src/database/repositories/collection-repository.ts
Quixelation: a common abstraction is not really helpful here since we have different requirements - it was changed to be be the same DX

hansegucker (approved):

PR #1353, opened 2026-04-01
fix: guard non-array vehicle priorities in preview

### LINKED ISSUE #1300: Uncaught TypeError in SimulatedRegionPreview Component
Adding vehicles to a simulated region and (un-)loading their contents results in a TypeError:

```
Uncaught TypeError: ids.includes is not a function
    ngOnInit simulated-region-preview.component.ts:158
```

This may be related to some flaky behaviour (disappearing vehicles / personnel) I experienced inside this view.

saschabuehrle:
Fixes #1300

`vehicleTemplatePriorities` can come through in an unexpected shape while loading/unloading data in the simulated region view. The component was calling `.includes()` directly and crashed when that value was not an array.

This normalizes the priorities before using them so non-array payloads fall back to an empty list instead of throwing.

Greetings, saschabuehrle


hansegucker:
@saschabuehrle Is this AI generated?

PR #1384, opened 2026-04-20
Implement Exercise Lazy-Loading

### LINKED ISSUE #1234: Do not load all exercises on server start up
Only load exercises on first use and deload the exercise after the last participant left the exercise.

Jogius:
### Summary

Closes #1234 by not loading all exercises from the database on startup. Instead, they are only migrated on startup and then lazily loaded when needed. They are purged from memory when no users are left in the exercise, directly after the save operation (which happens every 10 seconds).

### PR Checklist

### DIFF
--- a/backend/src/database/services/exercise-service.spec.ts
+++ b/backend/src/database/services/exercise-service.spec.ts
@@ -20,9 +20,10 @@ describe('Exercise-Service', () => {
     // In a naive implementation it can happen that such actions get removed from memory without being saved to the database.
     it('does not throw away actions while saving', async () => {
         const exerciseKeys = await createExercise(environment);
-        const exercise = environment.services.exerciseService.getExerciseByKey(
-            exerciseKeys.trainerKey
-        );
+        const exercise =
+            await environment.services.exerciseService.getExerciseByKey(
+                exerciseKeys.trainerKey
+            );
         const markAsAboutToBeSaved =
             exercise.markAsAboutToBeSaved.bind(exercise);
 
@@ -66,9 +67,10 @@ describe('Exercise-Service', () => {
     });
     it('does correctly update lastUsedAt', async () => {
         const exerciseKeys = await createExercise(environment);
-        const exercise = environment.services.exerciseService.getExerciseByKey(
-            exerciseKeys.trainerKey
-        );
+        const exercise =
+            await environment.services.exerciseService.getExerciseByKey(
+                exerciseKeys.trainerKey
+            );
         const beforeAction = Date.now();
         exercise.applyAction(
             {
--- a/backend/src/database/services/exercise-service.spec.ts
+++ b/backend/src/database/services/exercise-service.spec.ts
@@ -51,17 +51,16 @@ describe('Exercise-Service', () => {
             null
         );
 
-        const saveTick: () => Promise<void> = (
-            environment.server as any
-        ).saveTick.bind(environment.server);
-        await saveTick();
+        const saveUnsavedExercises = async () =>
+            environment.services.exerciseService.saveUnsavedExercises();
+        await saveUnsavedExercises();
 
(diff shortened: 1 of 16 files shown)


backend/src/fuesim-server.ts
```diff
@@ -8,24 +8,25 @@ export class FuesimServer {
     private readonly _httpServer: ApiHttpServer;
     private readonly _websocketServer: ExerciseWebsocketServer;
 
-    private readonly saveTickInterval = 10_000;
+    private readonly saveAndUnloadTickInterval = 10_000;
 
-    private readonly saveHandler: PeriodicEventHandler;
+    private readonly savAndUnloadHandler: PeriodicEventHandler;
```
hansegucker:
```suggestion
    private readonly saveAndUnloadHandler: PeriodicEventHandler;
```

backend/src/fuesim-server.ts
```diff
@@ -8,24 +8,25 @@ export class FuesimServer {
     private readonly _httpServer: ApiHttpServer;
     private readonly _websocketServer: ExerciseWebsocketServer;
 
-    private readonly saveTickInterval = 10_000;
+    private readonly saveAndUnloadTickInterval = 10_000;
```
hansegucker: Maybe we can find a better name for this tick handler? `housekeepingHandler`?

hansegucker (changes requested):
Looks great!

backend/src/fuesim-server.ts
Jogius: Maybe `exerciseHousekeeping` or `exerciseMaintenance` to be a little less vague? Just `housekeeping` works for me too, though. 

Edit: `exerciseUpkeep`?

fekoch:
Can you maybe test (using a large exercise), if we need some kind of grace period before purging exercises?

But, if the (re-)loading of even larger exercises happens fast enough, I do not think we should needlessly complicate the mechanism.

fekoch (changes requested):
As discussed @Jogius :

We should not load exercises into `ActiveExercises`, if we simply want to read or update metadata.

A plausible structure would be:

When reading or updating information of an exercise, a method (in `ExerciseService` or `ExerciseManagerService`) must first check, if there is a corresponding `ActiveExercise` currently in memory.

- If not, the operation should directly run on top of the database.
- If yes, the operation can simply be done on the in-memory object.

This should also prevent race conditions

backend/src/database/services/exercise-service.spec.ts
```diff
@@ -50,25 +51,25 @@ describe('Exercise-Service', () => {
...
+            environment.services.exerciseService.saveUnsavedExercises();
+        await saveUnsavedExercises();
 
         expect(markAsAboutToBeSavedMock).toHaveBeenCalledTimes(1);
         // one action still unsaved
         expect(exercise.temporaryActionHistory.length).toBe(1);
 
         markAsAboutToBeSaved();
-        await saveTick();
+        await saveUnsavedExercises();
```
hansegucker, line 63:
```suggestion
        await environment.services.exerciseService.saveUnsavedExercises();
```

backend/src/database/services/exercise-service.spec.ts
```diff
@@ -50,25 +51,25 @@ describe('Exercise-Service', () => {
             null
         );
 
-        const saveTick: () => Promise<void> = (
-            environment.server as any
-        ).saveTick.bind(environment.server);
-        await saveTick();
+        const saveUnsavedExercises = async () =>
+            environment.services.exerciseService.saveUnsavedExercises();
+        await saveUnsavedExercises();
```
hansegucker, line 56:
```suggestion
```

hansegucker:
We should pay attention that we have to update the cached exercise instance on changes.

Issue #1135, opened 2025-11-03
Add setting to enable "load vehicle" button for participants


lukasrad02:
The button to load all personnel and materials into a vehicle is only shown for trainers, not for participants. While this resembles challenges in real exercises (finding all these people and things might take some time, especially on larger sites), it might be undesired in some scenarios (e.g., when using discussing strategy in a larger group of people and using the software as a digital replacement of a physical "Planspielplatte").

Trainer should therefore be able to enable the load button for participants. It might make sense to have this setting also toggle the highlighting of personnel and material of a vehicle (#882), which is currently only available to trainers.

christianschaeffer:
De facto it is possible to "load everything", namely when you load a patient into a patient-carrying vehicle.

So there is no reason to hide this button.

lukasrad02:
Alternatively, we could also change the patient load behavior (and make it dependent on the proposed setting) so that either the button is not available and loading a patient does not load personnel and material, or the button is available and loading a patient also loads personnel and material.

What do you think about this?

lukasrad02:
@christianschaeffer reminder to answer

christianschaeffer:
> Alternatively, we could also change the patient load behavior (and make it dependent on the proposed setting) so that either the button is not available and loading a patient does not load personnel and material, or the button is available and loading a patient also loads personnel and material.

I like this configurable approach. It'S not a big difference, but makes all kinds of scenarios explicit. 

However, please consider our remaining capacities. If it's too much effort, simply adding the button is already an improvement.

PR #1423, opened 2026-04-28
Add configuration options for vehicle loading/highlighting for vehicle loading

lukasrad02:
### Summary

- Enable the button to completely load a vehicle for participants. Contributes to #1135
- Highlight related elements (e.g. vehicle + personnel) for nobody, trainers, or everybody

### PR Checklist


lukasrad02:
Test scenarios at <https://github.com/hpi-sam/fuesim-digital-public-test-scenarios/pull/37>

lukasrad02:
Regarding #1135, I've decided to initially only implement a setting that toggles the visibility of the load all button for participants, but not change the behavior of loading all personnel and materials when loading a patient. By default, the button is now available to participants, resolving inconsistency that loading a patient implicitly loads all personnel and materials.

Disabling this implicit behavior if the button is disabled cannot be done properly: When reducing an action, we don't know whether this action was proposed by a trainer or participant, so the only option would be to pass this as an additional property of the `LoadVehicleAction`, that can then be evaluated by the reducer. We could use the permission function of the action to check that the value matches the actual role of the client who sent the action, but this feels like a really hacky workaround.

What do you think, @hpi-sam/bp2025hg1?

christianschaeffer:
@lukasrad02 I cannot speak for the BP Team, but for me, optionally hiding the button but changing nothing else feels like a reasonable compromise.

lukasrad02:
Just for the record: Changing the behavior when loading a patient is still possible even after this PR has been merged. So when in doubt, we can just merge this PR, but keep #1135 open for further discussion

lukasrad02:
> Regarding #1135, I've decided to initially only implement a setting that toggles the visibility of the load all button for participants, but not change the behavior of loading all personnel and materials when loading a patient. By default, the button is now available to participants, resolving inconsistency that loading a patient implicitly loads all personnel and materials.
> 
> Disabling this implicit behavior if the button is disabled cannot be done properly: When reducing an action, we don't know whether this action was proposed by a trainer or participant, so the only option would be to pass this as an additional property of the `LoadVehicleAction`, that can then be evaluated by the reducer. We could use the permission function of the action to check that the value matches the actual role of the client who sent the action, but this feels like a really hacky workaround.
>
> Changing the behavior when loading a patient is still possible even after this PR has been merged. So when in doubt, we can just merge this PR, but keep https://github.com/hpi-sam/fuesim-digital/issues/1135 open for further discussion

@hansegucker if I remember correctly, you agreed to this approach, right?

hansegucker:
> @hansegucker if I remember correctly, you agreed to this approach, right?

Yes.

docs/src/2_exercises/general_settings1.png
hansegucker, line 1: I don't like that there is an auto-saving option and options that have to be manually saved with the button in one card. In my opinion, we rather introduce more cards than combine this in one card.

frontend/src/app/pages/exercises/exercise/shared/exercise-settings/exercise-settings-modal/exercise-settings-modal.component.html
```diff
@@ -42,6 +42,33 @@ <h4 class="modal-title">Einstellungen</h4>
             >
                 Karteneinstellungen speichern
             </button>
+
+            <hr />
+
+            <div class="form-group mb-3">
+                <label class="form-label mb-1"
+                    >Markiere zusammengehörende Elemente auf der Karte</label
+                >
```
hansegucker:
```suggestion
                <label class="form-label mb-1">
                    Markiere zusammengehörende Elemente auf der Karte
                </label>
```

hansegucker (changes requested):

docs/src/2_exercises/general_settings1.png
lukasrad02: Do you have a suggestion how to name this new card? Currently, this feature only affects vehicles, so one could place it into the vehicles card. But I intentionally named the feature in a way that it can be extended to other elements, so it basically is a "map" feature. Having two "map" cards is also weird…

hansegucker: As I don't have a better idea right now, let's keep it your way for the moment.

lukasrad02:
@hansegucker please keep in mind that we have to merge <https://github.com/hpi-sam/fuesim-digital-public-test-scenarios/pull/37> first

hansegucker (approved):

PR #1474, opened 2026-05-19
Prevent Start of Templates From Simulation Overview

### LINKED ISSUE #1277: Template exercise can be started from simulation settings modal

fekoch:
### Summary
Fixes #1277. Additionally, it implements @lukasrad02 [proposal](https://github.com/hpi-sam/fuesim-digital/pull/1280#discussion_r2817350281) to instead let trainers easily create an exercise to test it.

### PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,17 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 - Add documentation for most features and integrate it with the software.
 - Zooming the map via touch input can be disabled by both trainers and participants per device. This helps preventing accidental zooming if multiple participants are working on a single device simultaneously.
 
+### Changed
+
+- Exercises can be directly created when editing exercise templates.
+- Templates can be directly deleted when editing them.
+
 ### Fixed
 
 - Creating a parallel exercise with many (>10) exercise instances no longer slows down the server significantly.
 - Massively reduce the server start up time with many exercises.
 - Old exercise exports don't fail the validation anymore
+- Exercises can no longer be accidentally started from the simulation overview.
 
 ## [0.12.1] - 2026-05-21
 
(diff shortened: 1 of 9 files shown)


hansegucker:
I think this should be targeted to `dev`.

CHANGELOG.md
lukasrad02, line 1: Please move the new entries to the `[Unreleased]` section

backend/src/database/services/exercise-service.ts
lukasrad02, line 1: What is the context of these changes? On a first glance, they don't seem to be related to the goal of this PR

lukasrad02 (dismissed):
@hansegucker do you still plan to merge #1280 (updated to which actions should be allowed, e.g., it meanwhile is possible to send/schedule alarm groups in a template)?

Otherwise, it might make sense to at least block the start action on the server side in this PR.

backend/src/database/services/exercise-service.ts
fekoch: Without that, the exercise type was not correctly set in the exercise state.

fekoch:
> @hansegucker do you still plan to merge #1280 (updated to which actions should be allowed, e.g., it meanwhile is possible to send/schedule alarm groups in a template)?
> 
> Otherwise, it might make sense to at least block the start action on the server side in this PR.

I think, we should not start prematurely checking the exercise type in actions, before we implement some kind of clean api for that (like #1280). For now, I would leave that as is.

hansegucker (approved):

PR #1475, opened 2026-05-19
Add touch zoom lock option

lukasrad02:
### Summary

Adds a button to both trainer and participant map that disables zoom via touch input. This helps preventing accidental zooming if multiple participants are working on a single device simultaneously.

### PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project does **not** adhere to [Semantic Versioning](https://semver.org
 ### Added
 
 - Add documentation for most features and integrate it with the software.
+- Zooming the map via touch input can be disabled by both trainers and participants per device. This helps preventing accidental zooming if multiple participants are working on a single device simultaneously.
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


hansegucker (approved):
There is a merge conflict.

fekoch (changes requested):
I just tried it using an ipad and there the title text of the lock button has confusing behaviour.

As the lock (and all the other) buttons stay "pressed" until a user taps somewhere else, the title text stays shown. For most of the other buttons, this is not really noticable, but for the lock-button, it is confusing. It always shows the previous title text:

- "Zoom per Touchscreen aktiv. [...]" after tapping to lock and
- "Zoom per Touchscreen deaktiviert. [...]" after tapping to unlock.

Further, if you toggle it directly, it does not clear the previous title.

As the title text does not provide any information to a touch user before pressing and, when shown, is confusing, let's remove the title text altogether and show a informational toast when the lock state changes.

lukasrad02:
> As the title text does not provide any information to a touch user before pressing and, when shown, is confusing, let's remove the title text altogether and show a informational toast when the lock state changes.

I agree that the tooltips are not helpful for touch-based input, but I'd suggest to still keep them for users with a mouse who may wonder what the lock button can be used for. To address the issue of the stale text, I'd use static text ("Zoom per Touch an-/ausschalten").
I'd add an additional toast for touch users and check whether I can find a way to prevent the tooltip from staying open on touch devices.

What do you think about this?

fekoch (approved):
great! thank you!

PR #1479, opened 2026-05-21
Handle saving of exercise states without crashing the server

### LINKED ISSUE #1478: Deleting an exercise template that still has active exercises crashes the backend
It is possible to delete an exercise template even while exercises that are based on the template exist, there is no warning regarding possible side effects.

When interacting with an exercise based on the deleted template (e.g. adding elements from the editor or visiting `/exercises` listing this exercise), the backend crashes, dumping the exercise state.

hansegucker:
### Summary

Closes #1478 

The issue was that the save handler in the backend tried to save the exercise with `baseTemplateId` still set to the cached template id and not to null (on deleting an exercise template, the database sets this to null in all related exercises).

### PR Checklist

### DIFF
--- a/backend/src/database/repositories/exercise-repository.ts
+++ b/backend/src/database/repositories/exercise-repository.ts
@@ -136,20 +136,29 @@ export class ExerciseRepository extends BaseRepository {
             );
     }
 
-    public async saveExerciseState(exercise: ExerciseInsert) {
+    public async saveExerciseState(
+        exercise: ExerciseInsert & { id: ExerciseId }
+    ) {
+        const {
+            currentStateString,
+            initialStateString,
+            stateVersion,
+            id,
+            tickCounter,
+        } = exercise;
         const exercisePatch = {
-            ...exercise,
+            currentStateString,
+            initialStateString,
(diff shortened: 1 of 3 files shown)


backend/src/database/repositories/exercise-repository.ts
```diff
@@ -136,20 +136,23 @@ export class ExerciseRepository extends BaseRepository {
...
+    public async saveExerciseState(
+        exercise: ExerciseInsert & { id: ExerciseId }
+    ) {
+        const { currentStateString, initialStateString, stateVersion, id } =
+            exercise;
         const exercisePatch = {
-            ...exercise,
+            currentStateString,
+            initialStateString,
+            stateVersion,
```
Jogius, line 152: I would argue that we do still want to save the entire exercise to the database, including fields like `tickCounter`, since they might change. Maybe it would be better to save all fields _except_ the `baseTemplateId`? Or even better, to set the `baseTemplateId` to null on all exercises in memory when deleting a template?

Jogius (dismissed):

fekoch (approved):

PR #581, opened 2022-12-17
Feature/state benchmark

Dassderdie:
* The first commit moves the state migrations from the `backend` to `shared`
* The benchmark itself is in an extra package that imports `shared`. I tried to include it directly in `shared` (problems with `node` typings and different code endpoints) and the root (Not a big advantage over the new package and potential problems with root, as the parent package, depending on a child folder)
* Because we might have exercise states for which we want to run the benchmarks but which are also confidential (https://github.com/hpi-sam/digital-fuesim-manv_test-scenarios), there is a `data` folder. The benchmark will be run for all exercises in the folder. Files with `.permanent.json` will be included by git; all others will be ignored.
* There are a lot more benchmarks that could be useful. The ones that are currently implemented should be enough to spot significant performance regressions. To benchmark the impact of optimizations one would probably need to add more.

### FYI: Possible performance improvements
 
I'm thinking about refactoring the patient-health and the tick action. Instead of changing the health of each patient every tick, each patient gets an `internalHealth` property and a `currentHealthChangePerSecond` and `currentHealthChangeStartTime`. The `health` of the patient can then be calculated via `health = internalHealth + (exerciseState.currentTime - currentHealthChangeStartTime) * currentHealthChangePerSecond`. The tick action would then only include patients, whose `currentHealthChangePerSecond` changed. This would drastically reduce the size of the export, the time needed for the validation, and the time needed for the ExerciseTick.
This could be implemented together with #490

<details>
<summary>Current exemplary output</summary>

```console

C:\Users\julia\Documents\Studium\Bachelorproject\digital-fuesim-manv>npm run benchmark

> digital-fuesim-manv@0.0.0 benchmark
> cd benchmark && npm run benchmark


> digital-fuesim-manv-benchmark@0.0.0 benchmark
> (pwd && npm run benchmark:linux-macos) || npm run benchmark:windows

'pwd' is not recognized as an internal or external command,
operable program or batch file.

> digital-fuesim-manv-benchmark@0.0.0 benchmark:windows
> set NODE_ENV=production&& node --experimental-specifier-resolution=node --loader ts-node/esm src/app.ts

(node:7000) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time(Use `node --trace-warnings ...` to show where the warning was created)
(node:7000) ExperimentalWarning: The Node.js specifier resolution flag is experimental. It could change or be removed at any time.

System Information:
  OS:      Windows 10 Pro 10.0.22621
  #CPUs:   16
[...]

PR #1486, opened 2026-05-22
Remove redundant operations to reduce server startup time

hansegucker:
### Summary

While reading the whole code around saving and restoring exercises, I found out the following things:

1. When applying an action, the `currentStateString` is updated and the action is added to the `temporaryActionHistory`.
2. On the next run of `saveTick`, both the current state and the newly added actions are stored in the database in one transaction (so either the updated state and the action is saved or none of them is saved).

On the next server restart, the following things happen:

1. All exercises are fetched from the database and, if outdated, are migrated to the current state version.
2. After this, only up-to-date exercise states are in the database or the server doesn't start.
3. Now, in `restore`, all actions are fetched again, are applied to the initial state, the current state gets set again (to the same current state as before, as the current state was either already up-to-date due to the above-mentioned save logic or due to the migration logic), and the actions are trashed as they are in the database already (of course).

So step 3 seems redundant to me and also causes to take the server start up more than 10 minutes on production.

But: I only discussed this with @Jogius and I am not sure if I missed something. Therefore, I would we clad if @ClFeSc and @lukasrad02 could take a look at this.

### PR Checklist

### DIFF
--- a/backend/src/exercise/active-exercise.ts
+++ b/backend/src/exercise/active-exercise.ts
@@ -5,10 +5,9 @@ import type {
     ExerciseKey,
     ParticipantKey,
     TrainerKey,
+    ExerciseState,
 } from 'fuesim-digital-shared';
 import {
-    ExerciseState,
-    validateExerciseState,
     applyAction,
     cloneDeepMutable,
     reduceExerciseState,
@@ -286,30 +285,11 @@ export class ActiveExercise {
         }
     }
 
-    public restore(keepActions: boolean): void {
-        // Check State Version
-        if (this.exercise.stateVersion !== ExerciseState.currentStateVersion) {
-            throw new RestoreError(
-                `The exercise was created with an incompatible version of the state (got version ${this.exercise.stateVersion}, required version ${ExerciseState.currentStateVersion})`,
-                this.exercise.id
-            );
-        }
-
-        // Validate initial state
-        const errors = validateExerciseState(this.exercise.initialStateString);
-        if (errors.length > 0) {
-            throw new ValidationErrorWrapper(errors);
-        }
-
-        this.restoreState(keepActions);
-    }
-
     /**
-     * @param keepActions This indicates whether to keep the actions that were applied while restoring in the array (when `true`) or to remove them (when `false` and when the database gets used)
      * Recreates the {@link currentState} by applying all actions from {@link temporaryActionHistory} to the {@link initialState}
      * as well as adding actions to the end to gracefully mark the end of the previous exercise session.
      */
-    private restoreState(keepActions: boolean) {
+    public restoreState() {
         let currentState = cloneDeepMutable(this.exercise.initialStateString);
 
         this.temporaryActionHistory.forEach((actionWrapper) => {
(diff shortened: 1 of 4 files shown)


ClFeSc:
@hansegucker I've taken a quick glimpse and feel like that this should be a correct change. I'll look into it in more detail hopefully tomorrow.

backend/src/exercise/active-exercise.ts
lukasrad02, line 296: *(Feel free to move this into a separate issue. I noticed it during the review, but it was already broken on `dev`)*

I'm surprised about this validation. If I did not miss anything, `restoreState()` is only called at <https://github.com/hpi-sam/fuesim-digital/blob/511034f5f174c3ab1d34c2fef3fa87f81b802f6d/backend/src/database/services/exercise-service.ts#L184> and there a validation is already performed before: <https://github.com/hpi-sam/fuesim-digital/blob/511034f5f174c3ab1d34c2fef3fa87f81b802f6d/backend/src/database/services/exercise-service.ts#L137-L138>

However, no matter whether I keep this additional validation or remove it, the server crashes (HTTP 500) when uploading an invalid exercise export. I think there are two issues here:

- There is no exception handling in the call chain from `router.post('/exercise', () => {/*...*/})`, so any `ValidationErrorWrapper` will not be caught
- The error on the server side is `TypeError: Cannot read properties of undefined (reading 'reducer')`. It seems like neither `this.validateAction()` nor `validateExerciseExport` handles invalid (For testing, I've edited the `type` of a `history.actionHistory[]` action entry, so the error message makes sense in this context)

ClFeSc (approved):
The code looks good, I think. I tested it a bit but I don't have my setup for this project as good under control as to test it properly with having an old state in the database and then migrating in the database. If you prefer I can put in some more work, though.

Also, I did a little digging through the history of how this has come to light:

1. The concept of state migration is younger than the concept of loading exercises from database. The first commit I could find that introduces state migrations (f86e816f1ef8051b665e60598e3e1cd81cc99f41) has the first commit introducing restoring from database (4b4746515df649454799cfb37baac1f107bdffb7) as its parent. Thus, from the very beginning of the design process, the divide into migrating the state and restoring from database existed.
1. Additionally, I think, in the beginning, migrating did not require applying the actions. The first commit to introduce an actually usable migration (afc7fa943e67accd68fc83f8193c1aec2eb34f73) shows this quite nicely: We migrate the state and then all actions. A migration function could have chosen to do this, but it wasn't baked into the migration itself like it is now (maybe introduced by #581 or #722, but I haven't verified this).

Thus, it might be time to finally clean up this mess.

hansegucker:
> The code looks good, I think. I tested it a bit but I don't have my setup for this project as good under control as to test it properly with having an old state in the database and then migrating in the database. If you prefer I can put in some more work, though.
> 
> Also, I did a little digging through the history of how this has come to light:
> 
>     1. The concept of state migration is younger than the concept of loading exercises from database. The first commit I could find that introduces state migrations ([f86e816](https://github.com/hpi-sam/fuesim-digital/commit/f86e816f1ef8051b665e60598e3e1cd81cc99f41)) has the first commit introducing restoring from database ([4b47465](https://github.com/hpi-sam/fuesim-digital/commit/4b4746515df649454799cfb37baac1f107bdffb7)) as its parent. Thus, from the very beginning of the design process, the divide into migrating the state and restoring from database existed.
> 
>     2. Additionally, I think, in the beginning, migrating did not require applying the actions. The first commit to introduce an actually usable migration ([afc7fa9](https://github.com/hpi-sam/fuesim-digital/commit/afc7fa943e67accd68fc83f8193c1aec2eb34f73)) shows this quite nicely: We migrate the state and then all actions. A migration function could have chosen to do this, but it wasn't baked into the migration itself like it is now (maybe introduced by #581 or #722, but I haven't verified this).
> 
> 
> Thus, it might be time to finally clean up this mess.

Thanks for verifying our thoughts and the tests!

PR #1512, opened 2026-06-03
Add auto-incrementing counter in vehicle names

### LINKED ISSUE #373: Default vehicle names
* the vehicle names should by default be a number that increments by 1 for each new vehicle

lukasrad02:
### Summary

Closes #373

### PR Checklist


shared/src/store/action-reducers/vehicle.ts
```diff
@@ -55,6 +56,23 @@ import { logVehicleAdded, logVehicleRemoved } from './utils/log.js';
...
+ * Increments {@link draftState.vehicleCounter} and replaces the first `#` in the vehicle name by the current counter
+ * @param draftState The state to operate on
+ * @param vehicle The vehicle to apply the number on
+ * @returns The mutated vehicle
+ */
+function applyVehicleNumber(
+    draftState: WritableDraft<ExerciseState>,
+    vehicle: WritableDraft<Vehicle>
+) {
+    const vehicleNumber = ++draftState.vehicleCounter;
```
ClFeSc: The original issue mentioned no sharing of numbers across vehicle templates (see https://github.com/hpi-sam/fuesim-digital/issues/373#issuecomment-1187437240).

hansegucker: I also the think the counter should be per vehicle template.

lukasrad02: Implemented in 7b471ab1cf2152231e9b7e2179d7b30b9480940e

shared/src/store/action-reducers/vehicle.ts
```diff
@@ -55,6 +56,26 @@ import { logVehicleAdded, logVehicleRemoved } from './utils/log.js';
...
+function applyVehicleNumber(
+    draftState: WritableDraft<ExerciseState>,
+    vehicle: WritableDraft<Vehicle>
+) {
+    if (!Object.hasOwn(draftState.vehicleCounters, vehicle.templateId))
+        draftState.vehicleCounters[vehicle.templateId] = 0;
+
+    const vehicleNumber = ++draftState.vehicleCounters[vehicle.templateId]!;
+
+    vehicle.name = vehicle.name.replace('#', String(vehicleNumber));
```
ClFeSc, line 113: You are currently increasing the counter every time, even when no `#` is actually present in the name. Does this make sense? It means that there would be gaps in the counter when one changes the default vehicle name of a template to one without a `#` and then back to one that contains a `#`. I'm not sure, though, I feel like this could also be a desired behaviour.

ClFeSc, line 113: You are currently only replacing one `#`. Wouldn't it make more sense to replace all `#` using `replaceAll`?

ClFeSc (commented):
Just two things I want to bring up to a discussion.

shared/src/store/action-reducers/vehicle.ts
lukasrad02: I was also thinking about this, but I did not find a fully satisfying solution. `replaceAll` would turn vehicle number `1` with template name `RTW ##` into `RTW 11`, which could be confused with vehicle number `11`. With the current implementation, we would end up with `RTW 1#`, which would at least be obviously misconfigured. And I could imagine people actually trying multiple `#` to get zero-padding.

Of course, we could also implement padding, escaping (for literal `#` in the name) and so on, but this all feels like over-engineering to me…

lukasrad02: I intended it to be this way, but this is not a strong opinion.

The numbering will appear like containing gaps anyways, since a vehicle from an alarm group receives its number the moment the alarm group is *sent*. So any manually vehicle added manually while an alarm group has been sent, but not yet arrived, will already get a higher number.

I think this is okay, since I rather understand the goal of this feature as "we want vehicle names that are automatically unique (instead of all vehicles being `RTW ???`) to be able to identify a specific vehicle", not as "we want vehicle names that follow a strict naming convention that conveys meaning (time of arrival etc)"

ClFeSc: Fair, I didn't think of padding or something like this. Then leave it like it is, in my opinion. As you have said, it will very quickly show undesired behaviour this way when we leave it like that.

ClFeSc: Also fair. Like with the other comment I think that we have a lot of room for more complex behaviour that would be quite overengineered for now.

lukasrad02:
Test scenarios: <https://github.com/hpi-sam/fuesim-digital-public-test-scenarios/pull/40>

shared/src/store/action-reducers/vehicle.ts
```diff
@@ -95,6 +95,26 @@ export function isVehicleLoading(
...
+ * @param draftState The state to operate on
+ * @param vehicle The vehicle to apply the number on
+ * @returns The mutated vehicle
+ */
+function applyVehicleNumber(
+    draftState: WritableDraft<ExerciseState>,
+    vehicle: WritableDraft<Vehicle>
+) {
+    if (!Object.hasOwn(draftState.vehicleCounters, vehicle.templateId))
+        draftState.vehicleCounters[vehicle.templateId] = 0;
```
ClFeSc, line 109: Maybe consider deleting this entry again when a template gets deleted? There is not really a point in keeping it any longer as it will never be used again.

shared/src/state-migrations/56-vehicle-counters.ts
```diff
@@ -0,0 +1,15 @@
+import type { UUID } from '../utils/uuid.js';
+import type { Migration } from './migration-functions.js';
+
+export const vehicleCounters56: Migration = {
```
ClFeSc, line 16: This migration breaks old states where a template contains a `#` with the meaning of a literal number sign (hash). Vehicles created from such templates would have a literal `#` in their name in previous versions but that gets replaced by the number with this migration. Take [exercise-state-681282-incrementing-vehicle-names.json](https://github.com/user-attachments/files/28994326/exercise-state-681282-incrementing-vehicle-names.json). It was created in 7c3188b40691adedde600f46b4e8dadaff7d74ed (the commit at the root of this PR) and importing it in that version shows a mix of `???` and `#`, just as I have typed in. Importing it in the latest version of this PR (ded314f1664af6c6e949620893033c763eb3a196) leads to a replacement of the `#` from left to right (as I've created the vehicles). This does not feel like a useful behaviour.

One possible solution would be to add a flag to the add vehicle action that controls whether to replace `#`, that gets set to `false` in this migration and will be `true` for all future calls, though this seems inelegant.

ClFeSc (changes requested):
I've given it a short test and it felt very intuitive as it is implemented right now. However, I don't think the migration is correct as of yet as it is modifying the names of existing vehicles.

shared/src/state-migrations/56-vehicle-counters.ts
lukasrad02: I was also thinking about this, but did not find a satisfying solution. I also don't like adding a separate flag to be used for the migration. The alternative would be to do implement some way of escaping for literal `#` characters in vehicle names, which we decided against in <https://github.com/hpi-sam/fuesim-digital/pull/1512#discussion_r3386410113>.

So I think the question is what we want to achieve with the migrations. I feel like we don't have to ensure that old exercises do not change at all, as long as the overall behavior stays the same. Especially, if the behavior we want to preserve is something that is impossible to achieve in newly created exercises (currently, there is no escaping, so the first `#` will *always* be replaced) and the change will likely only affect a small number of exercises (in the exercises I have observed, I haven't seen any vehicles with `#` in their names). So, I'd prefer to go with one of the following approaches:

- Leave the implementation and migration as-is and accept that vehicle names can change (in likely only a small number of exercises)
- Implement escaping for literal `#` signs in vehicle names and adapt the migration to escape all `#` characters

---

Discussing the migration: What do you think about replacing the `???` with `#` when importing vehicle templates? This should not affect existing vehicles (since the add action already contains a copy of the vehicle template, not a reference/id to it), but would automatically add numbering to all newly added vehicles in imported exercises.

However, it does change the behavior of imported exercises and, in contrast to the issue discussed above, can easily be prevented by not just implementing this part of the migration

ClFeSc: > I was also thinking about this, but did not find a satisfying solution. I also don't like adding a separate flag to be used for the migration. The alternative would be to do implement some way of escaping for literal # characters in vehicle names, which we decided against in https://github.com/hpi-sam/fuesim-digital/pull/1512#discussion_r3386410113.

Yeah, I agree, no escaping. This would just be too much for such a small problem.

> So I think the question is what we want to achieve with the migrations. I feel like we don't have to ensure that old exercises do not change at all, as long as the overall behavior stays the same. Especially, if the behavior we want to preserve is something that is impossible to achieve in newly created exercises (currently, there is no escaping, so the first # will always be replaced) and the change will likely only affect a small number of exercises (in the exercises I have observed, I haven't seen any vehicles with # in their names).

I'd argue that changes in vehicle names on the map after a migration is generally not acceptable. If I want to name my RTWs `RTW 1 #blaulicht`, `RTW 2 #blaulicht`, and so on, I don't want that to be changed. However, I agree with your assessment that this is probably quite rare. I couldn't come up with a realistic example where a `#` might be used. @christianschaeffer do you know of examples where a genuine, literal `#` might be used as part of vehicle names? If not, I'd still not be happy to keep it like it is, but I'd be ok with it.

> So, I'd prefer to go with one of the following approaches:
>
>   * Leave the implementation and migration as-is and accept that vehicle names can change (in likely only a small number of exercises)
>   * Implement escaping for literal # signs in vehicle names and adapt the migration to escape all # characters

If @christianschaeffer does not raise objections, I'd go with the first option. If there is an objection, maybe the second option, or an exercise setting that enables/disables this feature. We could then disable this setting in the migration of old exercises but enable it in new ones (and maybe even enable it after the migrations are done, if possible).

> Discussing the migration: What do you think about replacing the ??? with # when importing vehicle templates? This should not affect existing vehicles (since the add action already contains a copy of the vehicle template, not a reference/id to it), but would automatically add numbering to all newly added vehicles in imported exercises.
> 
> However, it does change the behavior of imported exercises and, in contrast to the issue discussed above, can easily be prevented by not just implementing this part of the migration

I was also thinking about this. I think that this might even be a desired behaviour, as I can't imagine anyone using `???` as an intended literal in vehicle template's default names. And if anyone does not want to use the auto-incrementing numbers they can always rename them. In contrast to vehicles within the map we only have a few here and they are all in one place, so I'd say this is fine.

Additionally, as this is a new feature, people might otherwise not be aware and use their own templates with `???` for forever without realising they could use `#` instead.

christianschaeffer: I don't see any important domain-specific uses of `#` in vehicle names. And as long as we don't modify already placed vehicles, I have nothing against changing the behavior of template default names in some cases, i.e., in very old exercises.
So @lukasrad02, you can go ahead with the first of your proposed implementation schemes.

Also, I approve of replacing `???` with `#` in templates (again, not in instances).

Lastly, though I'm sure you already considered it, please make sure that `#` also works in the vehicle names of alarm groups.

ClFeSc: The current implementation would change all vehicle names of vehicles added to the map while the source template name had a `#`. However, as you say, this will be very rare (if it exists at all), so that should be fine, I guess.

I haven't tested alarm groups but I think they should work as they use the same method of creating a vehicle, I think. I'm not sure, though.

lukasrad02: > I don't see any important domain-specific uses of `#` in vehicle names. And as long as we don't modify already placed vehicles, I have nothing against changing the behavior of template default names in some cases, i.e., in very old exercises.

As Clemens said, the suggested (variant 1) approach may modify existing vehicles, if a vehicle template name or alarm group vehicle name contains `#`. (Editing vehicle templates is only possible since approx. 1 year, but setting the name of alarm group vehicles was always possible).

So, what I'd interpret from your reply would be to implement escaping to preserve all vehicle names in existing exercises. EDIT: Implementing escaping shouldn't take long, so even if it might not be strictly necessary, I think it does not harm either. So this might be the better approach than continuing discussing this

---

> Lastly, though I'm sure you already considered it, please make sure that # also works in the vehicle names of alarm groups.

Yes, this also works for alarm groups. The vehicle name that can be edited in the alarm groups modal is still only a template, so any `#` entered there (be it the default name from the vehicle template or some custom input) will be replaced with the counter value.

(Since the names are created when the alarm group is sent, the vehicles from an alarm group invocation will always have consecutive numbers\*, even if they don't arrive at the same time and vehicles have been edited from the editor in between. I think this is fine, since the primary goal of this feature is to provide unique names, not to carry any specific meaning, as already discussed in <https://github.com/hpi-sam/fuesim-digital/pull/1512#discussion_r3386445804>)  
\* within a single vehicle template

christianschaeffer: > As Clemens said, the suggested (variant 1) approach may modify existing vehicles if a vehicle template name or alarm group vehicle name contains #. (Editing vehicle templates is only possible since approx. 1 year, but setting the name of alarm group vehicles was always possible).
>
> So, what I'd interpret from your reply would be to implement escaping to preserve all vehicle names in existing exercises. EDIT: Implementing escaping shouldn't take long, so even if it might not be strictly necessary, I think it does not harm either. So this might be the better approach than continuing discussing this

To be honest, I don't really remember the details of the template-refactoring-case, the old logic, and how it leads to overwriting existing vehicles' names now. Under the new circumstances, where there is no trivial separation between instantiated vehicles and "templates" in the editor or alarm groups, **I take back the strict requirement to leave stuff on the map unchanged**.

If you didn't start with an escaping implementation yet, feel free to stop this and simply keep the behavior as is.

shared/src/store/action-reducers/vehicle.ts
lukasrad02: Good idea! Implemented in a3fa724a3de089036c3ad33e57af436059666aea

shared/src/state-migrations/56-vehicle-counters.ts
ClFeSc, line 16: From the missing replacement of `???` by `#` I assume you've decided against that?

ClFeSc (approved):
Looks good to me! Just one question regarding the omission of replacing `???` with `#` in templates.

shared/src/state-migrations/56-vehicle-counters.ts
lukasrad02: Oh, the thread just got to long and I forgot about this. Thank you for the reminder!

Implemented in 9cc7ce7276438e1cded8179a1bfdaf2070f1f32f

ClFeSc (approved):
Looks good, tested the import with an old export and everything worked as expected.

PR #1593, opened 2026-06-30
Add Zod schema for id maps

hansegucker:
### Summary

In order to harden the validation of the exercise state, I reintroduced the functionality for id maps.

### PR Checklist

### DIFF
--- a/shared/src/state.ts
+++ b/shared/src/state.ts
@@ -25,7 +25,7 @@ import {
 import { patientSchema } from './models/patient.js';
 import { hospitalPatientSchema } from './models/hospital-patient.js';
 import { patientCategorySchema } from './models/patient-category.js';
-import { uuid, uuidSchema } from './utils/uuid.js';
+import { getIdMapSchema, uuid, uuidSchema } from './utils/uuid.js';
 import { SpatialTree } from './models/utils/spatial-tree.js';
 import { logEntrySchema } from './models/log-entry.js';
 import {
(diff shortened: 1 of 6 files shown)


shared/src/state.ts
```diff
@@ -76,46 +76,46 @@ export const exerciseStateSchema = z.strictObject({
...
     hospitalPatients: z.record(uuidSchema, hospitalPatientSchema),
-    alarmGroups: z.record(uuidSchema, alarmGroupSchema),
+    alarmGroups: getIsIdMapSchema(alarmGroupSchema),
 
-    clients: z.record(uuidSchema, clientSchema),
+    clients: getIsIdMapSchema(clientSchema),
     /** All client names that are currently in the exercise or joined the exercise in the past */
     collectedClientNames: z.array(z.string()),
 
     radiograms: z.record(uuidSchema, exerciseRadiogramSchema),
```
lukasrad02: I think this should also use the id map schema.

Globally searching for `z.record(uuid` also gives some more results. Some of them are false positives where using the new schema is infeasible (e.g., `hospitalPatients` only have a `patientId` property, no `id` property), but most of the time, the new schema could be used (also in other files/models than the root state object)

(Edit: In some cases, including this one, you'll have to apply my suggested patch to `getIsIdMapSchema`, before you can use it)

shared/src/utils/uuid.ts
```diff
@@ -12,3 +12,23 @@ export function uuid(): UUID {
...
+        (val) => {
+            for (const [key, value] of Object.entries(val)) {
+                if (key !== value.id) {
+                    return false;
+                }
+            }
+            return true;
+        },
+        {
+            error: "There is an object that's saved with another id than the object's id.",
```
lukasrad02:
```suggestion
            error: "There is an object that's saved with another key than the object's id.",
```

Maybe this is easier to understand than using "id" in both cases

shared/src/utils/uuid.ts
```diff
@@ -12,3 +12,23 @@ export function uuid(): UUID {
 export type UUID = string;
 
 export const uuidSchema = z.uuidv4();
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const schemaWithId = z.object({
+    id: uuidSchema,
+});
+export function getIsIdMapSchema<T extends typeof schemaWithId>(schema: T) {
```
lukasrad02: I'd prefer dropping the `Is` from the name. For the old validators, starting with `Is` was a typical pattern also for the built-ins (`IsInt`, `IsString`, etc), but for the Zod schemas, it is not

lukasrad02: Constraining the schema to extend `z.object({id: uuidSchema})` does not work with composite schemas (e.g. the `exerciseRadiogramSchema` discriminated union), even if all possible schemas have an `id` property. Instead, use the following approach:


```suggestion
interface WithId {
    id: UUID;
}

export function getIsIdMapSchema<S extends z.core.$ZodType<WithId>>(schema: S) {
```

`z.core.$ZodType<T>` restricts the possible schemas to only such where `z.infer<schema>` extends `T` (see <https://zod.dev/library-authors#how-to-accept-user-defined-schemas>)

lukasrad02 (changes requested):

hansegucker:
@lukasrad02 Thanks for the review, I changed the mentioned pieces.

shared/src/state.ts
```diff
@@ -76,46 +76,46 @@ export const exerciseStateSchema = z.strictObject({
...
-    patients: z.record(uuidSchema, patientSchema),
-    materials: z.record(uuidSchema, materialSchema),
+    vehicles: getIdMapSchema(vehicleSchema),
+    personnel: getIdMapSchema(personnelSchema),
+    patients: getIdMapSchema(patientSchema),
+    materials: getIdMapSchema(materialSchema),
 
-    restrictedZones: z.record(uuidSchema, restrictedZoneSchema),
+    restrictedZones: getIdMapSchema(restrictedZoneSchema),
     measures: z.record(measureSchema.shape.id, measureSchema),
```
lukasrad02: I don't know whether there is a special reason why `.shape.id` is used instead of `uuidSchema`, but on the first glance, those two also just look like id maps

shared/src/state.ts
```diff
@@ -76,46 +76,46 @@ export const exerciseStateSchema = z.strictObject({
...
-    materialTemplates: z.record(uuidSchema, materialTemplateSchema),
-    personnelTemplates: z.record(uuidSchema, personnelTemplateSchema),
+    vehicleTemplates: getIdMapSchema(vehicleTemplateSchema),
+    materialTemplates: getIdMapSchema(materialTemplateSchema),
+    personnelTemplates: getIdMapSchema(personnelTemplateSchema),
     measureTemplates: z.record(z.string(), measureTemplateCategorySchema),
-    mapImageTemplates: z.record(uuidSchema, mapImageTemplateSchema),
+    mapImageTemplates: getIdMapSchema(mapImageTemplateSchema),
 
     scoutables: z.record(scoutableSchema.shape.id, scoutableSchema),
```
lukasrad02: Same thing as before, also looks like an id map

shared/src/state.ts
lukasrad02, line 141: I think it should be possible to chain `getIdMapSchema(…).optional()` to be able to use it here

lukasrad02 (changes requested):
Is there a reason why you left quite a few id maps unvalidated? I've pointed out some locations that were actually not matched by my previously suggested global search, but there are even ones that should have been found (e.g. `{Patient, PatientTemplate, HospitalPatient}.healthStates`)

shared/src/state.ts
hansegucker: Yes, but it's not an id map schema (resourceDescriptionSchema doesn't contain an id).

lukasrad02 (approved):
Since we know that we have at least one exercise with a UUID mismatch in the ID maps, you should be prepared that the server won't start after the next release with this PR.

Just to keep in mind, so we can reduce the downtime after said release.

PR #1608, opened 2026-07-19
Add more export functionality for exercises/exercise templates in the exercise manager

### LINKED ISSUE #1476: Export  all feature for exercise templates
There should be a button to export all exercise templates of a user as json files.

hansegucker:
### Summary

Close #1476  + some bonus functionality (also for exercises and for single exercise templates)

### PR Checklist

### DIFF
--- a/backend/src/database/services/exercise-manager-service.ts
+++ b/backend/src/database/services/exercise-manager-service.ts
@@ -7,9 +7,11 @@ import type {
     TrainerKey,
     OrganisationId,
 } from 'fuesim-digital-shared';
+import { ZipArchive } from 'archiver';
 import type { ExerciseRepository } from '../repositories/exercise-repository.js';
 import type { SessionInformation } from '../../auth/auth-service.js';
 import {
+    type ExerciseEntry,
     type ExerciseInsert,
     type ExerciseTemplateDetailsEntryWithUserRole,
     type ExerciseTemplateInsert,
@@ -298,5 +300,81 @@ export class ExerciseManagerService {
         return this.exerciseService.getExercisesViewportsById(
             exerciseTemplate.exercise.id
         );
+    }
+
+    public async buildExercisesArchive(
+        exercises: ExerciseEntry[],
+        session?: SessionInformation
+    ) {
+        const archive = new ZipArchive();
+        for (const exercise of exercises) {
+            // eslint-disable-next-line no-await-in-loop
+            const stateExport = await this.exerciseService.getExport(
+                exercise.trainerKey,
+                true,
+                session
+            );
+            const buffer = Buffer.from(JSON.stringify(stateExport));
+            archive.append(buffer, {
+                name: `exercise-state-${exercise.participantKey}.json`,
+            });
+        }
+        await archive.finalize();
+        return archive;
+    }
+
+    public async exportAllExercisesForOrganisation(
+        organisationId: OrganisationId,
+        session: SessionInformation
+    ) {
+        const organisation =
+            await this.organisationRepository.getOrganisationById(
+                organisationId
+            );
+        if (!organisation) {
+            throw new NotFoundError();
+        }
+        const isMember =
+            await this.organisationRepository.isMemberOfOrganisationById(
+                organisationId,
+                session.user.id
+            );
+        if (!isMember) {
+            throw new PermissionDeniedError();
+        }
+
+        const exercises =
+            await this.exerciseRepository.getAllExercisesForOrganisation(
+                organisationId
+            );
+        return this.buildExercisesArchive(exercises, session);
+    }
+
+    public async exportAllExerciseTemplatesForOrganisation(
+        organisationId: OrganisationId,
+        session: SessionInformation
+    ) {
+        const organisation =
+            await this.organisationRepository.getOrganisationById(
+                organisationId
+            );
+        if (!organisation) {
+            throw new NotFoundError();
+        }
+        const isMember =
+            await this.organisationRepository.isMemberOfOrganisationById(
+                organisationId,
+                session.user.id
+            );
+        if (!isMember) {
+            throw new PermissionDeniedError();
+        }
+
+        const exercises = (
+            await this.exerciseRepository.getAllExerciseTemplatesForOrganisation(
+                organisationId
+            )
+        ).map((exerciseTemplate) => exerciseTemplate.exercise);
+        return this.buildExercisesArchive(exercises, session);
     }
 }
--- a/backend/src/database/services/exercise-manager-service.ts
+++ b/backend/src/database/services/exercise-manager-service.ts
@@ -6,9 +6,11 @@ import type {
     TrainerKey,
     OrganisationId,
 } from 'fuesim-digital-shared';
+import { ZipArchive } from 'archiver';
 import type { ExerciseRepository } from '../repositories/exercise-repository.js';
 import type { SessionInformation } from '../../auth/auth-service.js';
 import {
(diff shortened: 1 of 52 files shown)


backend/src/database/services/exercise-manager-service.ts
```diff
@@ -296,4 +298,80 @@ export class ExerciseManagerService {
...
+            const stateExport = await this.exerciseService.getExport(
+                exercise.trainerKey,
+                true,
+                session
+            );
+            const buffer = Buffer.from(JSON.stringify(stateExport));
+            archive.append(buffer, {
+                name: `exercise-state-${exercise.participantKey}.json`,
+            });
+        }
```
Jogius: Why not
```suggestion
        const exports = await Promise.all(exercises.map((ex) => this.exerciseService.getExport(ex.trainerKey, true, session));
        
        for (const [i, export] of Object.entries(exports)) {
            const buffer = Buffer.from(JSON.stringify(export));
            archive.append(buffer, {
                name: `exercise-state-${exercises[i].participantKey}.json`,
            });
        }
```
or equivalent instead of disabling the linter?

backend/src/database/services/exercise-manager-service.ts
```diff
@@ -296,4 +298,80 @@ export class ExerciseManagerService {
...
+            );
+        if (!isMember) {
+            throw new PermissionDeniedError();
+        }
+
+        const exercises = (
+            await this.exerciseRepository.getAllExerciseTemplatesForOrganisation(
+                organisationId
+            )
+        ).map((exerciseTemplate) => exerciseTemplate.exercise);
```
Jogius, line 380: I would argue the full templates, including the template name and description, should be exported.

backend/src/database/services/exercise-service.ts
```diff
@@ -473,16 +468,53 @@ export class ExerciseService {
                 emitterId: actionWrapper.getAction().emitterId,
                 time: actionWrapper.getAction().index,
             })),
-        ]
-            // TODO: Is this necessary?
-            .sort((a, b) => a.time - b.time);
+        ].sort((a, b) => a.time - b.time);
```
Jogius, line 472: Have you decided that this is strictly necessary? otherwise, keep the todo

backend/src/routers/organisation-router.ts
```diff
@@ -110,6 +112,38 @@ export function createOrganisationRouter(
...
+        .get(async (req, res) => {
+            const id = organisationIdSchema.parse(req.params.id);
+
+            const archive =
+                await exerciseManagerService.exportAllExercisesForOrganisation(
+                    id,
+                    req.session!
+                );
+
+            res.attachment(`exercises-${id}.zip`);
```
Jogius: As a user, I want the filename to contain a stripped but human readable version of the organisation name instead of the non human readable id.

backend/src/routers/organisation-router.ts
```diff
@@ -110,6 +112,38 @@ export function createOrganisationRouter(
...
+        .get(async (req, res) => {
+            const id = organisationIdSchema.parse(req.params.id);
+
+            const archive =
+                await exerciseManagerService.exportAllExerciseTemplatesForOrganisation(
+                    id,
+                    req.session!
+                );
+
+            res.attachment(`exercise-templates-${id}.zip`);
```
Jogius: As a user, I want the filename to contain a stripped but human readable version of the organisation name instead of the non human readable id.

frontend/src/app/shared/components/exercise-card/exercise-card.component.html
```diff
@@ -51,15 +51,51 @@ <h2 class="card-title h4 d-flex">
...
                 <button
-                    class="btn btn-outline-danger"
-                    (click)="deleteExercise()"
+                    type="button"
+                    class="btn btn-outline-primary"
+                    ngbDropdownToggle
                 >
-                    <i class="bi bi-trash me-2"></i>
-                    Löschen
+                    <i class="bi bi-three-dots-vertical"></i>
```
Jogius: I like the idea of this dropdown 

Image

but I *personally* dont like the look of the button 

Image


(specifically the combination of the three vertical dots and the dropdown arrow). I think something like "Mehr" or "Aktionen" looks much better imo.

Image

Image

If you think it looks fine with the dots, you can just resolve this -- I might feel unreasonably strongly about this.

frontend/src/app/shared/components/exercise-template-card/exercise-template-card.component.html
```diff
@@ -106,16 +106,55 @@ <h2 class="card-title h4 d-flex mt-2">
...
                 <button
-                    class="btn btn-outline-danger"
-                    (click)="deleteExerciseTemplate()"
+                    type="button"
+                    class="btn btn-outline-primary"
+                    ngbDropdownToggle
                 >
-                    <i class="bi bi-trash me-2"></i>
-                    Löschen
+                    <i class="bi bi-three-dots-vertical"></i>
```
Jogius: Same here.

Jogius (changes requested):
Minor changes, overall looks very good. Please fix the pipeline though.

Question: is it in scope to enable the importing of such an exported zip file? I think this would greatly enhance the value created by these changes, but if it's complex, it might not be worth the time.

backend/src/database/services/exercise-manager-service.ts
hansegucker: But they wouldn't be importable then.

frontend/src/app/shared/components/exercise-card/exercise-card.component.html
hansegucker: I agree with the "Mehr" solution.

backend/src/routers/organisation-router.ts
hansegucker: Do you have suggestion how to the slugifying?

backend/src/database/services/exercise-manager-service.ts
Jogius: Fair point. Since this would be a change to export formats / we would need to introduce another one, that's probably not worth the time.

backend/src/routers/organisation-router.ts
Jogius: There are a couple of common ways to do this -- if we don't want to introduce a new dependency (which I'm against), we could e.g. use this function

```js
const slugify = (text) => {
  return text
    .toString()                     // Cast to string (optional)
    .normalize('NFKD')              // The normalize() using NFKD method returns the Unicode Normalization Form of a given string.
    .replace( /[\u0300-\u036f]/g, '' )   // Remove diacritics
    .toLowerCase()                  // Convert the string to lowercase letters
    .trim()                         // Remove whitespace from both sides of a string (optional)
    .replace(/\s+/g, '-')           // Replace spaces with -
    .replace(/[^\w\-]+/g, '')       // Remove all non-word chars
    .replace(/\_/g,'-')             // Replace _ with -
    .replace(/\-\-+/g, '-')         // Replace multiple - with single -
    .replace(/^-+|-+$/g, '');       // Remove leading and trailing -
}
```
as found in [this gist](https://gist.github.com/codeguy/6684588?permalink_comment_id=4325476#gistcomment-4325476)

Jogius (approved):
lgtm

Jogius:
I think you have the same problem as I did on another PR, where the tests now take longer than the timeout (4m) to run. We should probably increase the time here before merging, so that we don't have a failing pipeline on dev-next.

PR #1649, opened 2026-07-30
Address Angular warnings

lukasrad02:
### Summary

- Remove unused imports from components
- Use alternative for deprecated SASS function

### PR Checklist

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -111,7 +111,7 @@ jobs:
               with:
                   install-command: (cd shared && npm ci --no-audit) && (cd frontend && npm ci --no-audit)
             - name: Build frontend
-              run: cd frontend && npm run build
+              run: cd frontend && npm run build:deployment
 
     build-backend:
         timeout-minutes: 5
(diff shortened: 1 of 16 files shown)


hansegucker:
Can we try to convince our linting job to fail when such an unused import occurs?

lukasrad02:
> Can we try to convince our linting job to fail when such an unused import occurs?

Yes, this seems to be possible: <https://angular.dev/extended-diagnostics#configuration>

We just have to decide

- whether we want to treat all warnings as errors in the pipeline (maybe excluding some expected warnings, such as the one about `emitDecoratorMetadata`) or only check for unused imports explicitly
- whether we want this to be part of the frontend build job (which might be confusing if the build fails due to something which is rather a linter warning) or of the lint job (which would result in building the frontend during the lint job, since as far as I know, there is no way to check for Angular warnings without a full build)

My preference would be to fail on all (except for explicitly excluded) warnings and check it in the build job.

hansegucker:
> My preference would be to fail on all (except for explicitly excluded) warnings and check it in the build job.

I agree.

lukasrad02:
I have now implemented checking for Angular diagnostics in the pipeline

hansegucker (approved):

PR #1706, opened 2026-09-18
Fix migration testing

hansegucker:
### Summary

This ensures that history and current state migrations are properly tested and also improves the speed of the tests by dropping end-to-end testing.

Related: https://github.com/hpi-sam/fuesim-digital-public-test-scenarios/pull/44

### PR Checklist

- [ ] If this PR adds or changes features or fixes bugs, this has been added to the changelog.
- [ ] If this PR adds or changes features, the related documentation has been updated.
- [ ] If this PR adds new actions or other ways to alter the state, [test scenarios](https://github.com/hpi-sam/fuesim-digital-public-test-scenarios) have been added.
- [ ] By signing off my commits (`git commit -s`), I certify that I have read and adhere to the terms of the [Developer Certificate of Origin 1.1](https://developercertificate.org/) and license the code in this Pull Request under this projects license ([LICENSE-README.md](https://github.com/hpi-sam/fuesim-digital/blob/dev/LICENSE-README.md)).
- [ ] If I have used third party code that requires attribution, I have mentioned it in the code and updated [inspired-by-or-copied-from-list.html](https://github.com/hpi-sam/fuesim-digital/blob/dev/inspired-by-or-copied-from-list.html).


backend/src/test/http-migration.spec.ts
```diff
@@ -89,13 +88,13 @@ describe('migration', () => {
...
-                                .send({
-                                    importObject: exercise,
+                            await environment.services.exerciseService.createExerciseFromFile(
+                                {
                                     organisationId: null,
-                                } satisfies PostExerciseRequestData)
-                                .expect(201);
+                                },
+                                exercise,
+                                false
```
lukasrad02, line 96: Instead of strictly passing `false` here, we could also consider dropping the duplication of `complete` (with action history) and `current` (state snapshot only) test scenarios and running the migration for every scenario twice:

- including the action history and forbidding discarding
- stripping the action history before upload and testing the migration of the state snapshot

With your current approach, you only ensure that the action migrations are always executed, but as long as no separate `current` scenario exists, the state migration may not be tested (or, to be more precise, will only be tested on the always identical `initialState`)

lukasrad02 (commented):
If you change the migration tests anyway, what do you think about moving them to `shared` and directly calling `createExerciseFromFile` or even only `migrateStateExport`?

Currently, the migration tests take quite some time to run and we might be able to save time if we do not include the HTTP request and database transactions for every scenario…

shared/src/state-migrations/migrations.spec.ts
lukasrad02, line 1: It might make sense to replace the `if (!cond) console.warn('');` checks with `expect(cond)` (if `expect` can be used outside `it`) or some other code that causes the test to fail.

Especially since many migration tests log to the console by themselves (all history-based migrations for pre-marketplace exports), the test output is pretty cluttered. Therefore, one might not notice if a scenario is not tested due to a malformed directory/file name…

(Edit: Proving my point, the test scenarios for state version 62 are currently not included in the tests, since the subdirectory (`one-action`, …) is missing and nobody noticed yet. Feel free to also fix this as part of this PR, so we don't get merge conflicts with the submodule changes)

lukasrad02 (commented):
Nice work regarding the migration test refactoring!

Sadly, even the refactored migration tests do not catch the bug in migration 63 if I revert the `materials` change there. I suppose this is caused by dropping the history for all pre-marketplace exports.  
For a sufficient coverage, I think it would make sense to recreate at least the most basic history-based test scenarios (e.g. `everything-dropped-on-the-map-once-complete` from state 23) or ideally even all test scenarios for the current state version.

hansegucker:
> For a sufficient coverage, I think it would make sense to recreate at least the most basic history-based test scenarios (e.g. `everything-dropped-on-the-map-once-complete` from state 23) or ideally even all test scenarios for the current state version.

Opened a follow-up issue for that: https://github.com/hpi-sam/fuesim-digital/issues/1708

lukasrad02 (approved):
Keep in mind to merge <https://github.com/hpi-sam/fuesim-digital-public-test-scenarios/pull/44> before and point the submodule to the merge commit on `main` instead of the commit on the separate branch

